PR #12, opened 2016-11-08
FEATURE: graphiql dev tool

unclecheese:
This change adds an implementation of the popular [graphiql](https://github.com/graphql/graphiql) plugin to your SilverStripe installation. By default, it's only accessible in dev mode.

## Possible Concerns:

* **Does this really belong in this module?** Could be separate, but given that the devtool is so useful and graphql is so new, I think the case can easily be made for its inclusion out of the box.

* **The route for the tool `/graphiql/`**, which is super cryptic and too similar to the primary `/graphql/` route. Open to other ideas.

* **Lots of boilerplate client side code**. The tool runs on React, so a `webpack.config.js`, `package.json`, etc., are all included. We may want that to not be so isolated, and perhaps work with the other webpack configs, react libraries, etc.


client/src/index.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import {render} from 'react-dom';
+import GraphiQL from 'graphiql';
+import fetch from 'isomorphic-fetch';
+import 'graphiql/graphiql.css';
+
+function graphQLFetcher(graphQLParams) {
+  return fetch(`${window.location.origin}/${GRAPHQL_ROUTE}/`, {
```
wilr: @unclecheese breaks for me using localhost/<project>/ perhaps use something like `document.querySelectorAll('base')[0].href`

src/GraphiQLController.php
```diff
@@ -0,0 +1,49 @@
...
+			}
+		}
+
+		if(!$route) {
+			throw new \RuntimeException("There are no routes set up for a GraphQL server. You will need to add one to the SilverStripe\Control\Director.rules config setting.");
+		}
+
+		Requirements::customScript(
+<<<JS
+var GRAPHQL_ROUTE = '{$route}';
```
chillu: Ultimately I want to support multiple graphql endpoints (e.g. a limited "public" and full "secure" one), so we'd have to make the route selectable. Good enough for now though :)

chillu:
Hey Aaron, sorry for getting back so late on this! Did you see that you can download this as an Electron app for OSX? That's what I've been using so far: https://github.com/skevy/graphiql-app

It's not universal, so Windows and Linux users will still have trouble with trying out GraphQL queries, so there's definitely merit in shipping this in the module - at least until standalone installations become more commonplace :)

If you can address Will's concern and my little niggle about the URL path, I'm happy to merge

src/GraphiQLController.php
unclecheese: Yeah, agreed. A more future-proofed API would be `/dev/graphiql?endpoint=/my-graphql`, with a default to the first one defined.

_config/routes.yml
```diff
@@ -4,3 +4,10 @@ Name: graphqlroutes
 SilverStripe\Control\Director:
   rules:
     'graphql': 'SilverStripe\GraphQL\Controller'
+---
+only:
+  environment: 'dev'
+---
+SilverStripe\Control\Director:
+  rules:
+    'graphiql': 'SilverStripe\GraphQL\GraphiQLController'
```
unclecheese: Have fixed this.

unclecheese:
Have addressed the above concerns and rebased against master.

chillu:
Sweet, just confirmed it works for me locally - thanks Aaron!

Issue #9, opened 2016-10-31
Scaffold types and queries from DataObject


chillu:
At the moment there's a lot of boilerplate involved in creating GraphQL types ([example](https://github.com/open-sausages/silverstripe-asset-admin/tree/experiments/graphql/code/GraphQL)). Reduce this by inferring data from DataObjects (where types/queries are based on them).
- Set fields through whitelist
- Infer field types from `DBField` (String, Text, Enum, etc)
- Field name mapping (see #8)
- Build read queries based on type
- Allow polymorphism based on interfaces (e.g. a `createPage` mutation which works with all subclasses of `Page`)
- Build create/update/delete queries
- Build input types for mutation queries
- Custom field resolvers
- Custom query resolvers
- Read queries optionally based on `SearchContext` definitions
- Register types and queries with a schema manager
- Allow extension of definitions (e.g. whitelist a new field added by `DataExtension`)

See https://github.com/chillu/silverstripe-graphql/issues/8 for a start on an API.

chillu:
Some links for @unclecheese:

 * https://github.com/larsbs/graysql-orm-loader
 * https://github.com/overblog/GraphQLBundle
 * https://github.com/goco-inc/graphql-activerecord
 * https://github.com/colepatrickturner/graphql-rails-resolver
 * https://github.com/4rthem/graphql-mapper
 * https://medium.com/the-graphqlhub/graphql-tour-interfaces-and-unions-7dd5be35de0d#.8imiz7f3j
 * http://stackoverflow.com/questions/34726666/real-world-example-of-graphqlinterfacetype-and-graphqluniontype
 * https://facebook.github.io/graphql/#sec-Inline-Fragments
 * https://github.com/charlieschwabacher/gestalt
 * https://github.com/facebook/graphql/issues/114

chillu:
I've discussed this a bit more with @unclecheese, and here's some really rough drafts on how this could look like. Overall, we feel that YAML could provide a good result for most use cases, as long as you can write custom resolver classes. These would be injectable, and implement `Extension` so they can be augmented by third-party code.

```yaml
GraphQL:
  Member:
    operations:
      - all: true # implies 'create', 'read', 'update', 'delete'
    fields:
      - Email
      - Fullname
      - Groups # Implicitly creates a "Group" GraphQL type
    resolver: MyGraphQLMemberResolver
  Group:
    operations:
      - read: true #only allow reads
    fields:
      Title
```

```php
[...]

PR #20, opened 2016-12-12
 FEATURE: DataObject scaffolding

unclecheese:
Per https://github.com/silverstripe/silverstripe-graphql/issues/9


phpunit.xml
```diff
@@ -8,6 +8,7 @@
     processIsolation="false"
     stopOnFailure="false"
     syntaxCheck="false"
+    colors="true">
```
chillu: That'll muck with the Travis output unless we specifically disable it there

src/Manager.php
```diff
@@ -41,6 +43,35 @@ public static function createFromConfig($config)
...
 
+        if(isset($config['scaffolding'])) {
+        	$scaffolder = GraphQLScaffolder::createFromConfig($config['scaffolding']);
+        } else {
+        	$scaffolder = new GraphQLScaffolder();
+        }
+
+        if(isset($config['scaffolding_providers'])) {
+        	foreach($config['scaffolding_providers'] as $provider) {
+        		if(!class_exists($provider)) {
```
chillu: Indentation

src/Scaffolding/Creators/DataObjectTypeCreator.php
```diff
@@ -0,0 +1,76 @@
...
+namespace SilverStripe\GraphQL\Scaffolding\Creators;
+
+use SilverStripe\GraphQL\TypeCreator;
+use SilverStripe\GraphQL\Manager;
+use SilverStripe\ORM\DataObjectInterface;
+use GraphQL\Type\Definition\ResolveInfo;
+
+/**
+ * Creates a GraphQL type for a DataObject
+ * @package SilverStripe\GraphQL\Scaffolding\Creators
```
chillu: I think we can leave `@package` out of our PHPDoc now, that's why we have namespaces

src/Scaffolding/Creators/DataObjectTypeCreator.php
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * @var array
+     */
+    protected $fieldsMap;
+
+
+    /**
+     * DataObjectTypeCreator constructor.
+     * @param Manager $manager
+     * @param $typeName
```
chillu: Type hints for all params please

src/Scaffolding/Creators/OperationCreatorTrait.php
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * OperationCreatorTrait constructor.
+     * @param Manager $manager
+     * @param $operationName
+     * @param $typeName
+     * @param null $resolver
+     * @param array $argsMap
+     */
+    public function __construct(Manager $manager, $operationName, $typeName, $resolver = null, $argsMap = [])
```
chillu: From the `getResolver()` implementation, it doesn't look like the `$resolver` argument should be optional?

src/Scaffolding/DataObjectTypeTrait.php
```diff
@@ -0,0 +1,88 @@
...
+        return $this->dataObjectName;
+    }
+
+    /**
+     * @return string
+     */
+    public function typeName()
+    {
+        /* Type names must be unique and predictable. Cannot allow customisation. */
+        //return $this->dataObjectTypeName ?: $this->getDataObjectInstance()->singular_name();
```
chillu: Remove commented out code

src/Scaffolding/Scaffolders/CreateOperationScaffolder.php
```diff
@@ -0,0 +1,93 @@
...
+        parent::__construct(
+            'create' . ucfirst($this->typeName()),
+            $this->typeName()
+        );
+
+        $args = [];
+        $instance = $this->getDataObjectInstance();
+
+        // Setup default input args.. Placeholder!
+        foreach ($instance->db() as $dbFieldName => $dbFieldType) {
```
chillu: Why would all database columns be accepted as input arguments here, when we define an input *type* later on in the class?

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
```diff
@@ -0,0 +1,338 @@
...
+    /**
+     * @var array
+     */
+    protected $fields;
+
+    /**
+     * DataObjectScaffold constructor.
+     * @param $dataObjectName
+     */
+    public function __construct($dataObjectName)
```
chillu: Arg shouldn't be called "name" if you expect a "class"?

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
```diff
@@ -0,0 +1,338 @@
...
+        $extraDataObjects = $this->getExtraDataObjects();
+        $fields = array_unique($this->fields->toArray());
+
+        if (empty($fields)) {
+        	$fields = Config::inst()->get(self::class, 'default_fields');        	
+        }
+
+        foreach ($fields as $fieldName) {
+            $result = $instance->obj($fieldName);
+            if ($result instanceof DBField) {
```
chillu: How do we deal with getters? Force $casting? Then we should throw an exception if the result is not an instance of DBField

src/Scaffolding/Scaffolders/DeleteOperationScaffolder.php
```diff
@@ -0,0 +1,50 @@
...
+            'ID' => 'ID!'
+        ];
+
+        parent::__construct(
+            'delete' . ucfirst($this->typeName()),
+            $this->typeName()
+        );
+
+        $this->setResolver(function ($object, array $args, $context, $info) {
+            if (singleton($this->dataObjectName)->canDelete()) {
```
chillu: Needs to be called on the actual object, not the singleton

src/Scaffolding/Scaffolders/DeleteOperationScaffolder.php
```diff
@@ -0,0 +1,50 @@
...
+    use DataObjectTypeTrait;
+
+    /**
+     * DeleteOperationScaffolder constructor.
+     * @param null $dataObjectName
+     */
+    public function __construct($dataObjectName)
+    {
+        $this->dataObjectName = $dataObjectName;
+        $this->args = [
```
chillu: Could we make this accept a List of IDs by default? It would help with the current case of asset admin batch deletes, and my gut feel is that it's going to be a common requirement in web apps

src/Scaffolding/Scaffolders/DeleteOperationScaffolder.php
```diff
@@ -0,0 +1,50 @@
...
+            'delete' . ucfirst($this->typeName()),
+            $this->typeName()
+        );
+
+        $this->setResolver(function ($object, array $args, $context, $info) {
+            if (singleton($this->dataObjectName)->canDelete()) {
+                $obj = DataList::create($this->dataObjectName)
+                    ->byID($args['ID']);
+
+                if ($obj) {
```
chillu: Should throw an exception if the object can't be found

src/Scaffolding/Scaffolders/GraphQLScaffolder.php
```diff
@@ -0,0 +1,160 @@
...
+                            $constStr = sprintf(
+                                '%s::%s',
+                                GraphQLScaffolder::class,
+                                strtoupper($op)
+                            );
+                            if (defined($constStr)) {
+                                $ops[] = constant($constStr);
+                            } else {
+                                throw new \Exception(
+                                    "Invalid operation: $op"
```
chillu: Mention the context here to make it debuggable (e.g. $dataObjectName)

README.md
```diff
@@ -243,6 +243,305 @@ This will create a new member with an email address,
 which you can pass in as query variables: `{"Email": "test@test.com"}`.
 It'll return the new `ID` property of the created member.
 
+## Scaffolding DataObjects into the Schema
+
+Making a DataObject accessible through the GraphQL API involves quite a bit of boilerplate. In the above example, we can see that creating endpoints for a query and a mutation requires creating three new classes, along with an update to the configuration, and we haven't even dealt with data relations yet. For applications that require a lot of business logic and specific functionality, an architecture like this affords the developer a lot of control, but for developers who just want to make a given model accessible through GraphQL with some basic Create, Read, Update, and Delete operations, scaffolding them can save a lot of time and reduce the clutter in your project.
```
unclecheese: I think I had just spent the previous eight hours defining ClassNames. ;-)

README.md
```diff
@@ -243,6 +243,305 @@ This will create a new member with an email address,
...
+In order to register the scaffolding provider with the manager, we'll need to make an update to the config.
+
+```yaml
+SilverStripe\GraphQL:
+  schema:
+    scaffolding_providers:
+      - My\Project\Post
+```
+
+### Scaffolding DataObjects through the Config layer
```
unclecheese: OK, will reverse those two paragraphs.

README.md
```diff
@@ -243,6 +243,305 @@ This will create a new member with an email address,
...
+
+```
+
+### Exposing a DataObject to GraphQL
+
+Let's now expose the `Post` type. We'll choose the fields we want to offer, along with a simple query and mutation.
+
+**Code**:
+```php
+class Post extends DataObject implements ScaffoldingProvider {
```
unclecheese: Was trying to keep things brief, as that was already shown, but yeah, probably better to clarify.

README.md
```diff
@@ -243,6 +243,305 @@ This will create a new member with an email address,
...
+	//...
+    public function provideGraphQLScaffolding(GraphQLScaffolder $scaffolder)
+    {
+    	$scaffolder->dataObject(Post::class)
+    		->addFields(['ID','Title','Content'])
+    		->query('readPosts', function() {
+    			return Post::get()->limit(10);
```
unclecheese: Agreed

README.md
```diff
@@ -243,6 +243,305 @@ This will create a new member with an email address,
...
+**YAML**:
+```yaml
+SilverStripe\GraphQL:
+  schema:
+    scaffolding:
+      My\Project\Post:
+        operations: [CREATE, READ, UPDATE, DELETE]
+```
+
+> As shorthand, the expression `[CREATE, READ, UPDATE, DELETE]` can also be expressed as `operations: all`.
```
unclecheese: Behind the scenes, these all map to class constants `GraphQLScaffolder::READ`, etc.. there is no `GraphQLScaffolder::ALL` constant, and I think capitalisation would confuse that mapping.

README.md
```diff
@@ -243,6 +243,305 @@ This will create a new member with an email address,
...
+SilverStripe\GraphQL:
+  schema:
+    scaffolding_providers:
+      - MyFamily\Model\Post
+    scaffolding:
+      My\Project\Post:
+        fields: [ID, Title, Content]
+        queries:
+          readPosts:
+            resolver: MyResolver
```
unclecheese: I have added a bit of lead-up to it now.

src/Manager.php
unclecheese: Not sure what I missed here. Can you elaborate?

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
unclecheese: Well, we could try to be smart about it, seeing as there are only four GraphQL field types.

Have updated to:
```
        foreach ($fields as $fieldName) {
            $result = $instance->obj($fieldName);
            if ($result instanceof DBField) {
                $typeName = $result->config()->graphql_type;
                $fieldMap[$fieldName] = (new TypeParser($typeName))->toArray();
            } else if(is_bool($result)) {
            	$fieldMap[$fieldName] = ['type' => Type::boolean()];
            } else if(is_int($result)) {
            	$fieldMap[$fieldName] = ['type' => Type::int()];
            } else if(is_float($result)) {
            	$fieldMap[$fieldName] = ['type' => Type::float()];
            } else {
            	$fieldMap[$fieldName] = ['type' => Type::string()];
            }
        }
```

examples/code/CommentsResolver.php
```diff
@@ -0,0 +1,20 @@
...
+    public function resolve($object, $args, $context, $info)
+    {
+		$comments = $object->Comments();
+		
+		if(isset($args['Today']) && $args['Today']) {
+			$comments = $comments->where('DATE(Created) = DATE(NOW())');
+		}
+
+		return $comments;
+    }
```
robbieaverill: Daaaaaamn! That's much easier to digest!

src/Manager.php
robbieaverill: @unclecheese PSR-2 is spaces instead of tabs. You could probably just run `phpcbf --standard=PSR2 ./` to sort everything out at once.

README.md
chillu: That's fine, but from memory it was inconsistent - some of the example classes had the `MyProject` namespace

README.md
```diff
@@ -438,6 +438,727 @@ This will create a new member with an email address, which you can pass in as
...
+* Via the config layer (declaratively)
+
+The example code will show demonstrate both methods for each section.
+
+### Our example
+
+For these examples, we'll imagine we have the following model:
+
+```php
+namespace MyProject\GraphQL;
```
chillu: ORM model classes shouldn't live in a `MyProject\GraphQL` namespace - in and of themselves, they have nothing to do with GraphQL, and it'll tend to confuse devs (by wrongly assuming there's something special about these models). I'd suggest putting them in a `MyProject` only (rather than a more verbose `MyProject\Model\`

README.md
```diff
@@ -438,6 +438,727 @@ This will create a new member with an email address, which you can pass in as
...
+
+### Exposing a DataObject to GraphQL
+
+Let's now expose the `Post` type. We'll choose the fields we want to offer, along with a simple query and mutation. To
+resolve queries and mutations, we'll need to specify the name of a resolver class. This class
+must implement the `SilverStripe\GraphQL\Scaffolding\ResolverInterface`. (More on this below).
+
+**Via YAML**:
+```
+SilverStripe\GraphQL:
```
chillu: Needs `yaml` markdown syntax marking

README.md
```diff
@@ -438,6 +438,727 @@ This will create a new member with an email address, which you can pass in as
...
+  schema:
+    scaffolding:
+      types:
+        MyProject\GraphQL\Post:
+          fields: [ID, Title, Content]
+          operations:
+            read: true
+            create: true
```
chillu: Doesn't have `update: true`, yet has an `updatePost()` query example below

README.md
```diff
@@ -438,6 +438,727 @@ This will create a new member with an email address, which you can pass in as
...
+	readPosts {
+		Title
+    	Content
+  }
+}
+```
+
+```
+mutation UpdatePost($ID: ID!, $Input: PostUpdateInputType!)
+	updatePost(ID:$ID, Input: { Title: "Some new Title" }) {
```
chillu: It's a bit confusing to write a query accepting variables here (`$ID` and `$Input`), and then only using one of the variables in the payload (`$ID`), but hardcoding the value of the other one (`Input: { Title: "Some new Title" }`). We've already used variables before when introducing mutations, so should stay consistent here.

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
```diff
@@ -0,0 +1,513 @@
...
+
+        $this->dataObjectClass = $dataObjectClass;
+    }
+
+    /**
+     * Adds visible fields.
+     *
+     * @param array $fields
+     */
+    public function addFields(array $fields)
```
chillu: I think `addFields()` should support adding a description as well, we want to encourage well documented APIs right? We could accept both numeric and assoc arrays (with some `array_filter(array_keys($fields), 'is_numeric')` concoction).

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
```diff
@@ -0,0 +1,513 @@
...
+
+        return $this;
+    }
+
+    /**
+     * @param $field
+     *
+     * @return mixed
+     */
+    public function addField($field)
```
chillu: Same here, `addField()` should have `$description` as an optional second argument

README.md
```diff
@@ -438,6 +438,727 @@ This will create a new member with an email address, which you can pass in as
...
+
+We can now access the posts via GraphQL.
+```
+query {
+	readPosts {
+		Title
+    	Content
+  }
+}
+```
```
chillu: At this point you probably should describe how the `readPosts` naming was generated (automatically inferred from singular_name and plural_name)

README.md
```diff
@@ -438,6 +438,727 @@ This will create a new member with an email address, which you can pass in as
...
+SilverStripe\GraphQL:
+  schema:
+    scaffolding:
+      types:
+        MyProject\GraphQL\Post:
+          fields: [ID, Title, Content]
+          operations:
+            read:
+              args:
+                StartingWith: String
```
chillu: `StartingWith` is a bit of a confusing argument example, particularly as a slight variation from the `StartsWith` ORM filter. Is there a reason you didn't just use `Title` as the arg name and perform a partial match?

README.md
```diff
@@ -438,6 +438,727 @@ This will create a new member with an email address, which you can pass in as
...
+```php
+	$scaffolder
+		->type(Post::class)
+    		->addFields(['ID','Title','Content'])
+    		->operation(GraphQLScaffolder::READ)
+    			->addArgs([
+    				'StartingWith' => 'String'
+    			])
+        		->setResolver(function($obj, $args) {
+        			$list = Post::get();
```
chillu: We probably shouldn't have any examples without `canView()` checks, I'd add at least a PHP comment like `// ... canView checks`. It's too easy for devs to assume that scaffolded resolvers will magically take care of this.

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
```diff
@@ -0,0 +1,513 @@
...
+            }
+
+            $this->addAllFieldsExcept($config['fieldsExcept']);
+        }
+
+        if (!isset($config['operations'])) {
+            return $this;
+        }
+
+        if ($config['operations'] === '*') {
```
chillu: The README mentions `all`, but the code only accepts `*`.

README.md
chillu: My point is: Why *isn't* there a `GraphQLScaffolder::ALL` constant as well? Constants are a way to communicate type strictness and avoid misspelling. Also, the README mentions `all`, but the code only accepts `*`. Can we just make this `GraphQLScaffolder::ALL`?

src/Scaffolding/Scaffolders/GraphQLScaffolder.php
```diff
@@ -0,0 +1,258 @@
...
+use SilverStripe\GraphQL\Scaffolding\Util\OperationList;
+use SilverStripe\GraphQL\Scaffolding\Util\ScaffoldingUtil;
+use SilverStripe\GraphQL\Scaffolding\Interfaces\ManagerMutatorInterface;
+use SilverStripe\ORM\ArrayLib;
+
+/**
+ * The entry point for a GraphQL scaffolding definition. Holds DataObject type definitions,
+ * and their nested Mutation/Query definitions.
+ */
+class GraphQLScaffolder implements ManagerMutatorInterface
```
unclecheese: Yeah, I thought about that, but there's also `DataObjectScaffolder`, `QueryScaffolder`, `OperationScaffolder`, etc. I'm not sure I'm comfortable with it being `Scaffolder` in that context, but I agree that `GraphQL` is redundant. 

`SchemaScaffolder`?

tests/ScaffoldingTest.php
```diff
@@ -0,0 +1,789 @@
+<?php
+
+namespace SilverStripe\GraphQL;
+
+use SilverStripe\Dev\SapphireTest;
+// use GraphQL\Type\Definition\Type;
```
unclecheese: These were written before a massive change in architecture and API. I'm working on the new test suite right now.

src/Scaffolding/Scaffolders/GraphQLScaffolder.php
chillu: Yeah `SchemaScaffolder` works for me.

src/Scaffolding/Util/TypeParser.php
```diff
@@ -0,0 +1,123 @@
...
+     */
+    protected $defaultValue = null;
+
+    /**
+     * TypeParser constructor.
+     * @param $rawArg
+     */
+    public function __construct($rawArg)
+    {
+        if (!preg_match('/^([A-Za-z]+)(!?)(\s*=\s*(.*))?/', $rawArg, $matches)) {
```
chillu: Protip: You can use a [non-capturing group](http://stackoverflow.com/questions/3512471/what-is-a-non-capturing-group-what-does-a-question-mark-followed-by-a-colon) in regexes to avoid leaving out `$matches[3]` here: `^([A-Za-z]+)(!?)(?:\s*=\s*(.*))?`

src/Scaffolding/Scaffolders/CRUD/Create.php
```diff
@@ -0,0 +1,103 @@
...
+        $this->dataObjectClass = $dataObjectClass;
+
+        parent::__construct(
+            'create'.ucfirst($this->typeName()),
+            $this->typeName()
+        );
+
+        // Todo: this is totally half baked
+        $this->setResolver(function ($object, array $args, $context, $info) {
+            if (singleton($this->dataObjectClass)->canCreate()) {
```
chillu: All `can*()` checks should pass in `$context['currentMember']` - an attempt to decrease the dependence of globals throughout the framework (Session dependency in `DataObject::can*()`). At the moment it's just future proofing, but it's pretty much always clearer to pass in context explicitly

chillu:
As a general note, all GraphQL fields should start lowercase, to distinguish them from types starting with uppercase. This clashes with our SilverStripe conventions (DataObject properties start with uppercase). I've written https://github.com/silverstripe/silverstripe-graphql/blob/master/src/Util/CaseInsensitiveFieldAccessor.php for this purpose. Doesn't necessarily need to be solved in this PR, but we need to address it soon (for query args and query/type fields)

README.md
```diff
@@ -438,6 +438,727 @@ This will create a new member with an email address, which you can pass in as
...
+
+```php
+namespace MyProject\GraphQL;
+use SilverStripe\GraphQL\Scaffolding\Interfaces\ResolverInterface;
+
+class MyResolver implements ResolverInterface
+{
+    public function resolve($object, $args, $context, $info)
+    {
+		$post = Post::get()->byID($args['ID']);
```
chillu: Needs `canEdit()` check

README.md
```diff
@@ -438,6 +438,727 @@ This will create a new member with an email address, which you can pass in as
...
+This resolver class may now be assigned as either an instance, or a string to the query or mutation definition.
+
+```php
+	$scaffolder
+		->type(Post::class)
+			->operation(GraphQLScaffolder::UPDATE)
+				->setResolver(MyResolver::class)
+				->end();
+```
+Or...
```
chillu: I'd just add the second variation as an inline comment on the first code example, for brevity's sake

README.md
```diff
@@ -438,6 +438,727 @@ This will create a new member with an email address, which you can pass in as
...
+query {
+	readPosts {
+		Title
+    	Content
+  }
+}
+```
+
+```
+mutation UpdatePost($ID: ID!, $Input: PostUpdateInputType!)
```
chillu: Mention somewhere around here that permission constraints are enforced by the built-in resolvers

src/Scaffolding/Util/TypeParser.php
unclecheese: Ah, I had been wondering if that was possible. Thanks.

README.md
```diff
@@ -438,6 +438,727 @@ This will create a new member with an email address, which you can pass in as
...
+        	->setUsePagination(false)
+        	->setResolver(function($obj, $args) {
+        		return Post::get()->sort('Date', 'DESC')->first();
+        	})
+        	->end()
+```
+
+**GraphQL**
+```
+mutation updatePostTitle(ID: 123, NewTitle: 'Foo') {
```
chillu: Missing `$` in args

README.md
```diff
@@ -438,6 +438,727 @@ This will create a new member with an email address, which you can pass in as
...
+query readPages {
+	Page
+}
+
+query readSiteTrees {
+	SiteTree
+}
+
+mutation createRedirectorPage {
+	RedirectorPageCreateInputType
```
chillu: Are you missing those input types in your `type` list above?

src/Scaffolding/Scaffolders/CRUD/Read.php
```diff
@@ -0,0 +1,48 @@
...
+
+        $typeName = $this->getDataObjectInstance()->plural_name();
+        $typeName = str_replace(' ', '', $typeName);
+        $typeName = ucfirst($typeName);
+        $operationName = 'read'.$typeName;
+
+        parent::__construct($operationName, $this->typeName());
+
+        $this->setResolver(function ($object, array $args, $context, $info) {
+            $list = DataList::create($this->dataObjectClass);
```
chillu: Shouldn't this filter by `canView()`?

src/Manager.php
```diff
@@ -6,16 +6,14 @@
 use GraphQL\Schema;
 use GraphQL\GraphQL;
 use SilverStripe\Core\Injector\Injector;
-use SilverStripe\Core\Injector\Injectable;
 use GraphQL\Type\Definition\ObjectType;
 use GraphQL\Error;
 use GraphQL\Type\Definition\Type;
-use SilverStripe\Security\Member;
```
robbieaverill: Hey Aaron - you'll need this back in again, it's used for the member methods at the bottom - I think this is why your tests are failing ATM

README.md
unclecheese: OK, have changed the namespace to `MyProject` for everything.

README.md
unclecheese: OK

README.md
unclecheese: Have changed the example to `CreatePost` instead

README.md
unclecheese: Have added an explanation

README.md
unclecheese: Have shown the input variables below the mutation

README.md
unclecheese: OK, have changed to `PartialMatch`.

README.md
unclecheese: Have added a simple canView() check.

README.md
unclecheese: OK

src/Scaffolding/Scaffolders/CRUD/Read.php
unclecheese: Yeah, but you have to use `filterByCallback` if you want to be thorough. Otherwise, a simple check against a singleton could work.

src/Manager.php
unclecheese: Have replaced this.

README.md
unclecheese: Yes. Have added them.

README.md
unclecheese: Fixed

src/Scaffolding/Scaffolders/CRUD/Read.php
robbieaverill: This callback would need to comply with the `OperationResolver` interface right? `$info` would need a `ResolveInfo` type hint

unclecheese: Mmm... If you passed it a named class, or an instance, it has to implement `OperationResolver`, but AFAIK, there's no way to enforce an argument signature on a closure.

README.md
unclecheese: All the CRUD operations now have permission checks.

tractorcow:
Some weird rebasing going on here. :P

src/Scaffolding/Scaffolders/CRUD/Create.php
```diff
@@ -0,0 +1,104 @@
...
+        );
+
+        // Todo: this is totally half baked
+        $this->setResolver(function ($object, array $args, $context, $info) {
+            if (singleton($this->dataObjectClass)->canCreate($context['currentMember'])) {
+                $newObject = Injector::inst()->createWithArgs(
+                    $this->dataObjectClass,
+                    $args
+                );
+                $newObject->write();
```
tractorcow: See below (or is it above?) on my comment regarding using a transaction. We should do it here too.

src/Scaffolding/Scaffolders/OperationScaffolder.php
```diff
@@ -0,0 +1,341 @@
...
+     */
+    protected $resolver;
+
+    /**
+     * @var array
+     */
+    protected $args = [];
+
+    /**
+     * @param $name
```
tractorcow: `@return` type please. :)

src/Scaffolding/Scaffolders/OperationScaffolder.php
```diff
@@ -0,0 +1,341 @@
...
+                return Delete::class;
+        }
+
+        return null;
+    }
+
+    /**
+     * OperationScaffolder constructor.
+     *
+     * @param null $operationName
```
tractorcow: Please use a non-null param type descriptor (and elsewhere).

src/Scaffolding/Scaffolders/OperationScaffolder.php
```diff
@@ -0,0 +1,341 @@
...
+        if (isset($config['args'])) {
+        	if(!is_array($config['args'])) {
+        		throw new Exception(sprintf(
+        			'args must be an array on %s',
+        			$this->operationName
+        		));
+        	}
+        	foreach($config['args'] as $argName => $argData) {
+        		if(is_array($argData)) {
+					if(!isset($argData['type'])) {
```
tractorcow: Indentation is a bit weird.

src/Pagination/PaginatedQueryCreator.php
```diff
@@ -22,11 +22,11 @@
     /**
      * @param Manager $manager
      */
-    public function __construct(Manager $manager)
+    public function __construct(Manager $manager, Connection $connection = null)
     {
         parent::__construct($manager);
 
-        $this->connection = $this->connection();
+        $this->connection = $connection ?: $this->connection();
```
tractorcow: I might need to catch up with @chillu on this one; In ReadFileQueryCreator in asset-admin, the connection() is scaffolded there. I might understand this design a bit better if I saw the proposed updated version of that class.

src/Scaffolding/Scaffolders/SchemaScaffolder.php
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    /**
+     * Find or make a query.
+     *
+     * @param string $name
+     * @param string $class
+     * @param null   $resolver
+     *
+     * @return bool|QueryScaffolder
```
tractorcow: It doesn't seem to return any bool does it?

src/Scaffolding/Util/ScaffoldingUtil.php
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * Given a DataObject subclass name, transform it into a sanitised (and implicitly unique) type
+     * name suitable for the GraphQL schema
+     *
+     * @param $class
+     * @return mixed
+     */
+    public static function typeNameForDataObject($class)
+    {
+        $typeName = Config::inst()->get($class, 'table_name', Config::UNINHERITED) ?:
```
tractorcow: Nice. :)

templates/GraphiQL.ss
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html lang="en">
```
tractorcow: As discussed earlier, please move to namespaced folder matching the controller FQN.

src/Pagination/PaginatedQueryCreator.php
tractorcow: I guess what confuses me is that sometimes:
 - $connection arg to constructor is ignored
 - sometimes it's mandatory
 - sometimes it's optional
 - sometimes it's returned from connection() method
 - sometimes it's stored and just ignored.
 
 Looking again at the code, this seems to be where these are constructed.
 
 ```php
 $queryCreator = Injector::inst()->create($queryCreatorClass, $manager);
 ```
 
 So my guess is we remove the `$connection` arg entirely from the constructor?
 
Probably just needs clarification. :P

templates/GraphiQL.ss
unclecheese: This shouldn't even be here. It's an artefact of the bad rebasing you were referring to.

examples/_config/config.yml
```diff
@@ -12,3 +12,46 @@ SilverStripe\GraphQL:
   authenticators:
     - class: SilverStripe\GraphQL\Auth\BasicAuthAuthenticator
       priority: 10
+    scaffolding_providers:
+     - My\Project\Post
```
unclecheese: This is in the `examples/` directory, so this is just an example of how you would set up your config with registered scaffolding providers.

src/FieldCreator.php
```diff
@@ -82,7 +82,7 @@ public function getAttributes()
             $attributes['type'] = $type;
         }
 
-        $resolver = $this->getResolver();
+        $resolver = $this->getResolver();
```
unclecheese: Have done.

src/GraphiQLController.php
```diff
@@ -0,0 +1,49 @@
...
+use SilverStripe\Control\Controller as BaseController;
+use SilverStripe\Control\Director;
+use SilverStripe\View\Requirements;
+ 
+class GraphiQLController extends BaseController 
+{
+	/**
+	 * @var string
+	 */
+	protected $template = 'GraphiQL';
```
unclecheese: Yeah, this feature was supposed to be deleted in this https://github.com/silverstripe/silverstripe-graphql/commit/4e3a3259620d548159dea594a940c3ce8747399c but at some point a rebase got stuffed.

src/Scaffolding/Interfaces/CRUDInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace SilverStripe\GraphQL\Scaffolding\Interfaces;
+
+interface CRUDInterface
```
unclecheese: OK

src/Pagination/PaginatedQueryCreator.php
```diff
@@ -22,11 +22,11 @@
     /**
      * @param Manager $manager
      */
-    public function __construct(Manager $manager)
+    public function __construct(Manager $manager, Connection $connection = null)
```
unclecheese: OK

src/Scaffolding/Interfaces/Configurable.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace SilverStripe\GraphQL\Scaffolding\Interfaces;
+
+interface Configurable
```
unclecheese: Have renamed to `ConfigurationApplier`

src/Scaffolding/Scaffolders/CRUD/Create.php
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * @return InputObjectType
+     */
+    protected function generateInputType()
+    {
+        $fields = [];
+        $instance = $this->getDataObjectInstance();
+
+        // Setup default input args.. Placeholder!
+        $db = (array) Config::inst()->get(
```
unclecheese: I think I tried that, but it omits the `fixed_fields`. If you know of a better way to capture those, please share, because this is ugly as.

src/Scaffolding/Scaffolders/CRUD/Create.php
```diff
@@ -0,0 +1,104 @@
...
+            'db',
+            Config::INHERITED
+        );
+
+        unset($db['ID']);
+
+        foreach ($db as $dbFieldName => $dbFieldType) {
+            $result = $instance->obj($dbFieldName);
+            $typeName = $result->config()->graphql_type;
+            $arr = [
```
unclecheese: Maybe. The config does set a default one for the base class `SilverStripe\ORM\FieldType\DBField:`, so you should always get something, unless the user nukes that setting.

src/Pagination/PaginatedQueryCreator.php
unclecheese: Yeah, good catch. There are a bit of acrobatics happening here. The issue is that in order to make scaffolding work, these creators need to be composable. We don't have the luxury of being able to define a subclass on the fly (unless we want to use PHP7 anonymous classes.. oh, please say yes).

The way the `PaginatedQueryCreator` was designed, it imposes on its descendants the creation of a `connection()` method. This solution was the best work around I could come up with without drastically changing the API surface and/or internals of `PaginatedQueryCreator`, but perhaps we should carve it up a bit per your suggestion. I'm quite partial to splitting it up into two methods.

src/Scaffolding/Scaffolders/CRUD/Create.php
tractorcow: Do you mean the inverse, fieldSpecs() includes fixed_fields, but `db` doesn't?

src/Scaffolding/Scaffolders/CRUD/Create.php
tractorcow: Ok, that's fair enough. :)

src/Pagination/PaginatedQueryCreator.php
chillu: So fundamentally the `PaginatedQueryCreator` API needs to communicate that it can't function without a `Connection`. At the moment it does this two ways: 

1. Through `abstract function connection()` (used for manual definitions), usually instanciated through a class name string in YAML config
2. Through a `$connection` constructor arg (used for scaffolding), instanciated through scaffolding in PHP

We could stick with a mandatory constructor arg for both cases, and do something with `Injector` to services to allow instanciation from a class name string with the correct `Connection` instance. That'll make the logic quite unintuitive since a separate `MyQueryConnection` class needs to be created, and injected as a `%$connection` service into `PaginatedQueryCreator`.

I'm in favour of the split Damian recommended, with two changes: `createConnection()` rather than `scaffoldConnection()`, since I think this method should be used for the baseline implementation. And making the `$connection` constructor arg mandatory for `PaginationScaffolder`. 

How does this look? https://gist.github.com/chillu/c0ece0978e4fbe77180acc8cc0067e19

tractorcow: +1 @chillu , although making that second argument mandatory could cause problems. The code I linked above won't know if that argument is necessary or not, since it takes an arbitrary class name.

`$queryCreator = Injector::inst()->create($queryCreatorClass, $manager);`

Maybe instead of making it mandatory, make it error if not passed AND if createConnection() returns null.

src/Scaffolding/Scaffolders/CRUD/Delete.php
```diff
@@ -0,0 +1,71 @@
...
+    public function __construct($dataObjectClass)
+    {
+        $this->dataObjectClass = $dataObjectClass;
+
+        parent::__construct(
+            'delete'.ucfirst($this->typeName()),
+            $this->typeName()
+        );
+
+        $this->setResolver(function ($object, array $args, $context, $info) {
```
chillu: I've done this now, see commit "Use DB transaction for multi delete"

src/Scaffolding/Scaffolders/CRUD/Create.php
chillu: This should be an atomic operation already, it's a single record (unlike `DELETE`) - unless you assume that a `write()` might trigger other writes?  We don't do that anywhere else in the CMS, so don't think we should start here.

src/Scaffolding/Scaffolders/CRUD/Update.php
```diff
@@ -0,0 +1,122 @@
...
+     *
+     * @return InputObjectType
+     */
+    protected function generateInputType()
+    {
+        $fields = [];
+        $instance = $this->getDataObjectInstance();
+
+        // Setup default input args.. Placeholder!
+        $db = (array) Config::inst()->get(
```
chillu: Done

src/Scaffolding/Scaffolders/CRUD/Create.php
chillu: I've amended the `testCreateOperationInputType` and `testUpdateOperationInputType` tests to check for the `Created` field after changing this to `DataObjectSchema`, works fine

src/Scaffolding/Scaffolders/OperationScaffolder.php
chillu: Done

examples/code/Post.php
```diff
@@ -0,0 +1,141 @@
...
+			->type('SilverStripe\CMS\Model\RedirectorPage')
+			->addFields(['ExternalURL', 'Content'])
+			->operation(SchemaScaffolder::READ)
+			->end()
+			->operation(SchemaScaffolder::CREATE)
+			->end()
+			->end()
+			->type('Page')
+			->addFields(['BackwardsTitle'])
+			->end();
```
chillu: Would be good to have this indented to retain some sanity ;)

sminnee:
Woohoo! Great to see this merged

PR #21, opened 2016-12-13
Pagination support

### LINKED ISSUE #7: GraphQL Paginated Asset List
As a user I want to read queries by page, so that I can open large folders.

## Acceptance criteria

 * Asset List View is migrated to use GraphQL for fetching data
 * Asset List view is paginated when data is pulled via GraphQL

## Tasks

 * [ ] Helper to create a connection from an ArrayList
 * [x] Node and edge types
 * [x] Default connection arguments for queries (start and limit)
 * [x] Unit tests for new helpers
 * [x] ~~Adapt asset pagination~~
 * [x] Update README with pagination example

Note: We could adapt the [graphql-php-relay Connection.php](https://github.com/ivome/graphql-relay-php/blob/master/src/Connection/Connection.php), but can't use it directly since it assumes a cursor-based pagination

## Research

Required for https://github.com/silverstripe/silverstripe-asset-admin/issues/159

There's two common approaches:
- Page number or offset pagination. More or less the standard.
- Cursor-based pagination. Great for fast changing data like social feeds, hence Relay mandates this approach. Can be implemented in the underlying PHP library via [graphql-relay-php](https://github.com/ivome/graphql-relay-php).

Read [graphql.org/learn/pagination](http://graphql.org/learn/pagination/) and [Understanding Pagination in GraphQL](https://dev-blog.apollodata.com/understanding-pagination-rest-graphql-and-relay-b10f835549e7#.mzuj6sxlp) for details.
- Relay only [supports](https://github.com/facebook/relay/issues/540) cursor based pagination, Apollo [supports](http://dev.apollodata.com/react/pagination.html) both. 
- Cursor-based pagination is probably overkill for our average CMS use case (data isn't changing that fast). 
- Offset-based pagination makes it harder to reference a particular view in an infinity scroll view (e.g. files in AssetAdmin) since offsets can change over time, but that's no different than 99% of webapps out there. It can easily be fixed by referencing a search result instead.
- By choosing offset-based pagination, we're making it harder to use a SilverStripe Content API with Relay (see [relay issue about offset-based pagination](https://github.com/facebook/relay/issues/540)). And since either pagination approach defines our "public API surface", it's going to be hard to change later.
- For custom data models, devs can choose to implement cursor-based pagination through applying [graphql-relay-php](https://github.com/ivome/graphql-relay-php) in their own resolvers.
- With offset-based pagination, you don't know if you've reached the end of the list unless you can pass back a total count (see [problem description](http://graphql.org/learn/pagination/#end-of-list-counts-and-connections)). In order to show client state based on total count (e.g. numbered pagination), we need nesting within our query responses anyway - what GraphQL calls a "connection object" (`{totalCount: 1, edges: [{id:99},{id:100}]}`). It's questionable if that's still good UX in 2016 though (as opposed to infinity/next-based pagination and powerful search), but cursor-based pagination lets us implement each UX.
- Cursors with custom ordering is explained in this [blog post](https://m.alphasights.com/how-to-do-effortless-pagination-with-relay-connections-and-graphql-ruby-a534ffaf5cbf#.aaeqxmwwb). In short, cursors encode a default sort order (`id---3` vs. `created_at---2016–02–02 02:40:27 UTC`). The average SS dev doesn't need to worry about these details, we can serialise it from the `DataList` query automatically.

~~Overall, while cursor-based pagination is a bit more convoluted in implementation and API signatures, I think it's the best way forward for us - most of this complexity will be abstracted away in the "graphql type scaffolder" (cursor encoding/decoding). I don't think there's much point in supporting multiple pagination approaches in SilverStripe's own APIs, although devs can always build offset-based pagination through the basic `graphql-php` building blocks (rather than the "graphql type scaffolder").~~

Update: We're recommending offset-based pagination as a quicker way to get this feature implemented, with some tradeoffs described in the [comments](https://github.com/chillu/silverstripe-graphql/issues/7#issuecomment-257185230)

wilr:
Fixes #7

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,11 +5,11 @@
 [![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/silverstripe/silverstripe-graphql/badges/quality-score.png?b=master)](https://scrutinizer-ci.com/g/silverstripe/silverstripe-graphql/?branch=master)
 
 This modules serves SilverStripe data as
-[GraphQL](http://facebook.github.io/react/blog/2015/05/01/graphql-introduction.html) representations,
-with helpers to generate schemas based on SilverStripe model introspection.
-It layers a pluggable schema registration system on top of the
-[graphql-php](https://github.com/webonyx/graphql-php) library.
-The APIs are very similar, for example
+[GraphQL](http://facebook.github.io/react/blog/2015/05/01/graphql-introduction.html)
+representations, with helpers to generate schemas based on SilverStripe model
+introspection. It layers a pluggable schema registration system on top of the
+[graphql-php](https://github.com/webonyx/graphql-php) library. The APIs are
+very similar, for example:
 
 ## Installation
 
@@ -21,32 +21,35 @@ composer require silverstripe/graphql
 
 ## Usage
 
-GraphQL is used through a single route which defaults to `/graphql`.
-You need to define *Types* and *Queries* to expose your data via this endpoint.
+GraphQL is used through a single route which defaults to `/graphql`. You need
+to define *Types* and *Queries* to expose your data via this endpoint.
 
 ## Examples
 
-Code examples can be found in the `examples/` folder
-(built out from the configuration docs below).
+Code examples can be found in the `examples/` folder (built out from the
+configuration docs below).
 
 ## Configuration
 
 ### Define Types
 
-Types describe your data. While your data could be any arbitrary structure,
-in a SilverStripe project a GraphQL type usually relates to a `DataObject`.
-GraphQL uses this information to validate queries and allow GraphQL
-clients to introspect your API capabilities. The GraphQL type system is hierarchical,
-so the `fields()` definition declares object properties as scalar types within
-your complex type. Refer to the [graphql-php type definitions](https://github.com/webonyx/graphql-php#type-system)
+Types describe your data. While your data could be any arbitrary structure, in
+a SilverStripe project a GraphQL type usually relates to a `DataObject`.
+GraphQL uses this information to validate queries and allow GraphQL clients to
+introspect your API capabilities. The GraphQL type system is hierarchical, so
+the `fields()` definition declares object properties as scalar types within
+your complex type. Refer to the
+[graphql-php type definitions](https://github.com/webonyx/graphql-php#type-system)
 for available types.
 
 ```php
 <?php
+
 namespace MyProject\GraphQL;
 
 use GraphQL\Type\Definition\Type;
 use SilverStripe\GraphQL\TypeCreator;
+use SilverStripe\GraphQL\Pagination\Connection;
 
 class MemberTypeCreator extends TypeCreator
 {
@@ -67,8 +70,8 @@ class MemberTypeCreator extends TypeCreator
             'Surname' => ['type' => Type::string()],
         ];
     }
-
 }
+
 ```
 
 Each type class needs to be registered with a unique name against the schema
@@ -81,7 +84,8 @@ SilverStripe\GraphQL:
       member: 'MyProject\GraphQL\MemberTypeCreator'
 ```
 
-### Define Queries
+
+## Define Queries
 
 Types can be exposed via "queries". These queries are in charge of retrieving
 data through the SilverStripe ORM. The response itself is handled by the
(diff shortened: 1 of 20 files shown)


chillu:
@wilr Travis build is failing

README.md
```diff
@@ -164,6 +168,198 @@ query {
...
+* `sortDirection`
+
+Additional arguments can be added by providing the `args` parameter (such as
+`Email` in the previous example).
+
+Pagination information is provided under the `pageInfo` type. This object type
+supports the following fields:
+
+* `totalCount` returns the total number of items in the list,
+* `hasNextPage` returns whether more records are available at the end
```
chillu: Missing `hasPreviousPage`?

README.md
```diff
@@ -164,6 +168,198 @@ query {
...
+    public function attributes()
+    {
+        return [
+            'name' => 'member'
+        ];
+    }
+
+    public function fields()
+    {
+        $groups = Connection::create([
```
chillu: Rename to `$groupsConnection` to make it clear that we're not dealing with a `DataList`

README.md
```diff
@@ -164,6 +168,198 @@ query {
...
+                // Optional filtering by properties
+                if(isset($args['Email'])) {
+                    $list = $list->filter('Email', $args['Email']);
+                }
+
+                return $list;
+            }
+        ]);
+    }
+}
```
chillu: Forgot the closing code block, so everything past this is misformatted

src/Pagination/Connection.php
```diff
@@ -0,0 +1,293 @@
...
+            if($offset > 0) {
+                $previousPage = true;
+            }
+        }
+
+        if($list instanceof Sortable) {
+            if(isset($args['sort'])) {
+                $direction = (isset($args['sortDirection'])) ? $args['sortDirection'] : 'ASC';
+
+                if(in_array($args['sort'], $this->sortableFields)) {
```
chillu: I know I'm being a pain in the ass, but could we implement ordering on multiple fields by default, in this notation? `readMembers(sortBy: [{field: Title}, {field: ID, direction: DESC}])`. It's been discussed and approved by some Relay devs over at https://github.com/graphql/graphql-relay-js/issues/20#issuecomment-220494222. I'm using "sort" rather than "order" here, since we've already set that convention in `DataList`.

tests/ConnectionTest.php
```diff
@@ -0,0 +1,141 @@
...
+use SilverStripe\ORM\ArrayList;
+use GraphQL\Type\Definition\Type;
+use SilverStripe\GraphQL\Pagination\Connection;
+use SilverStripe\GraphQL\Tests\Fake\TypeCreatorFake;
+use SilverStripe\GraphQL\Tests\Fake\DataObjectFake;
+use SilverStripe\GraphQL\Tests\Fake\PaginatedQueryFake;
+use GraphQL\Type\Definition\FieldDefinition;
+use GraphQL\Type\Definition\ResolveInfo;
+
+class ConnectionTest extends SapphireTest
```
chillu: Can you test sorting please?

src/Pagination/Connection.php
chillu: If you're sorting by a field that's not whitelisted, it should throw an exception (which will automatically surface in the GraphQL query response)

src/Pagination/Connection.php
```diff
@@ -0,0 +1,293 @@
...
+            if($limit && (($limit + $offset) < $count)) {
+                $nextPage = true;
+            }
+
+            if($offset > 0) {
+                $previousPage = true;
+            }
+        }
+
+        if($list instanceof Sortable) {
```
chillu: Should throw an exception if the list isn't sortable, but you have 'sort' args configured

src/Pagination/Connection.php
```diff
@@ -0,0 +1,293 @@
...
+
+    /**
+     * The maximum limit supported for the connection. Used to prevent excessive
+     * load on the server. To override the default limit, use {@link setLimits}
+     *
+     * @var int
+     */
+    protected $maximumLimit = 100;
+
+    public function __construct($args)
```
chillu: Not a fan of array-based constructor args - too easy to misspell an argument and go on a wild goose chase on why it isn't working. Can we only make required args distinct function args here, and move the rest to (fluent/chainable) setters? Also, each of these should have a getter, which makes it easier to implement subclasses of `Connection`

chillu:
Did you decide to keep sort title a string rather than enum? https://github.com/silverstripe/silverstripe-graphql/pull/21/commits/7652abe1b3a24ad6f3c8f82da4c9d8df2afa542f#diff-911835d9b7bb691adcf79ea0207f04e0R27. I thought we discussed enums in the standup this morning. I’m happy to make this a follow up enhancement though. It’ll change the GraphQL API signature (strings vs. symbols in queries), so we shouldn’t leave it too long.

Otherwise, great work! Merging, enums in sorts can be solved later.

wilr:
👍

PR #24, opened 2016-12-18
Allow sort field mapping

chillu:
Mainly designed to bridge case sensitivity between GraphQL (defaults to lower camel) and the SS ORM (defaults to upper camel). Will also come in handy for custom in-memory sorts, e.g. File.size

/cc @wilr


wilr:
@chillu Travis auth error? Tried restarting the job to no luck.

chillu:
@wilr Passing now

PR #25, opened 2016-12-18
API: Whitelist sortBy field using an EnumType

wilr:
`sortBy` now only accepts the whitelisted field names. Marking as an API change since queries will need to change from quoted to unquoted strings.


chillu:
@wilr I've made two minor modifications, and confirmed sorting still works. Waiting for tests to pass, then will merge

PR #28, opened 2016-12-20
API Standardise OperationResolver::resolve()

tractorcow:
API PaginatedQueryCreator is abstract
PSR2 cleanup
PHPDoc cleanup


src/MutationCreator.php
```diff
@@ -7,7 +7,7 @@
  *
  * @todo Validation support
  */
-class MutationCreator extends FieldCreator
+abstract class MutationCreator extends FieldCreator
```
wilr: zomg..

PR #41, opened 2017-01-05
DOC Add ResolveInfo type hint to resolve() example

robbieaverill:
Prevents fatal error (PHP 7.1) of an incompatible method declaration when running the example in graphiql.

Removed MyDataObject from ReadmembersQueryCreator as it was not used.


chillu:
Nice, thank you - also for cleaning up conventions!

PR #42, opened 2017-01-06
DOCS Implement canView checks into ReadMembersQueryCreator examples

robbieaverill:


robbieaverill:
I'll add these to `examples/` as well, then re-open

Issue #6, opened 2016-10-30
OAuth-based Authentication


chillu:
Session-based authentication relies on a manual mechanism for signing in (CMS login form). While we can already use programmatic authentication with Basic Auth over SSL, it's not very friendly to typical third party integrations. Implement an OAuth provider to make this more flexible, e.g. with http://oauth2.thephpleague.com/. It's unclear which flow we'd use (since we don't want any further user action beyond logging in to the session-based CMS).

See https://medium.com/the-graphqlhub/graphql-and-authentication-b73aed34bbeb#.ys9c7gc32

robbieaverill:
@chillu I'm happy to help out with this if you need some extra resource.

Re: the two auth flows - perhaps you could distinguish the authentication method from a request header. Either match the OAuth header or a custom SS session header.

chillu:
@robbieaverill Thanks for your offer, we can absolutely use the help! :) 

At the moment, the [GraphiQL](https://github.com/graphql/graphiql) browser with the SilverStripe Content API will return empty results filtered by  `canView()`. A quick fix for this would be allowing [basic auth](https://en.wikipedia.org/wiki/Basic_access_authentication) passed in through HTTP headers - which is still a bit finnicky since you need to base64 encode them ready for the HTTP header (can't set user and password separately in GraphiQL). That's more of a missing feature in GraphiQL though, and basic auth is good enough for development work. Plus it already works out of the box in SilverStripe.

I don't think we need a [viewer root node](https://gist.github.com/wincent/598fa75e22bdfa44cf47#how-does-graphql-know-what-type-of-data-is-being-requested-by-a-node-root-call), but maybe I haven't understood the concept completely :)

Regarding an OAuth 2.0 server, this would implicitly add OAuth support to SS core, which is a fairly large security surface - and hence would need some form of RFC. Ideally we'd avoid running an XHR from the client for this, but rather auto-generate a token on the server, stored against the member. I'd rather not reuse the "remember me" token for this. It would have to be removed on logout, so a simple 1:1 relationship with sessions. And should have an expiry date which matches the PHP session. A separate module might also allow multiple API keys (and a UI to manage them), but that's out of scope for now.

The Oauth 2.0 server implementation should be a separate module, which can be applied through the `SilverStripe\Control\RequestFilter` middleware - so the GraphQL module should only need some configuration to apply it to the `/graphql` endpoint.

[...]

PR #44, opened 2017-01-06
API Add support for BasicAuth to be used for native permission checks

robbieaverill:
The BasicAuth implementation will adequately protect the API endpoint, and will validate against a Member, but (by design) it will not be stored in the session and as such cannot be used for native SilverStripe permission checking (`Page::canView`).

This change adds an `AuthenticatorInterface` which can be implemented by various authenticator classes to use the `HTTPRequest` provided to GraphQL to authenticate against members, and return a `Member` instance if successful.

This PR includes a `BasicAuthAuthenticator`.

Example configuration:

```yaml
SilverStripe\GraphQL:
  # Enforce HTTP basic authentication for GraphQL requests
  authenticators:
    - class: SilverStripe\GraphQL\Auth\BasicAuthAuthenticator
      priority: 10
```

The only non-obvious implication here is that even if you have a global basic auth requirement set on your site, you'll still need to configure the `BasicAuthAuthenticator` for GraphQL.

The docs have been updated with authentication information, a configuration example, and an example for GraphiQL (app only) client implementation using the base64 encoded header.

---

Starts #6

### DIFF
--- a/README.md
+++ b/README.md
@@ -457,6 +457,72 @@ This tool is available in **dev mode only**. It can be accessed at
 
 <img src="https://github.com/graphql/graphiql/raw/master/resources/graphiql.png">
 
+## Authentication
+
+Some SilverStripe resources have permission requirements to perform CRUD operations
+on, for example the `Member` object in the previous examples.
+
+If you are logged into the CMS and performing a request from the same session then
+the same Member session is used to authenticate GraphQL requests, however if you
+are performing requests from an anonymous/external application you may need to
+authenticate before you can complete a request.
+
+Please note that when implementing GraphQL resources it is the developer's
+responsibility to ensure that permission checks are implemented wherever
+resources are accessed.
+
+### Basic Authentication
+
+Silverstripe has built in support for [HTTP basic authentication](https://en.wikipedia.org/wiki/Basic_access_authentication).
+It can be configured for GraphQL implementation with YAML configuration (see below).
+This is kept separate from the SilverStripe CMS authenticator because GraphQL needs
+to use the successfully authenticated member for CMS permission filtering, whereas
+the global `BasicAuth` does not log the member in or use it for model security.
+
+#### YAML configuration
+
+You will need to define the class under `SilverStripe\GraphQL.authenticators`.
+You can optionally provide a `priority` number if you want to control which
+Authenticator is used when multiple are defined (higher priority returns first).
+
+Here's an example for implementing HTTP basic authentication:
+
+```yaml
+SilverStripe\GraphQL:
+  authenticators:
+    - class: SilverStripe\GraphQL\Auth\BasicAuthAuthenticator
+      priority: 10
+```
+
+#### In GraphiQL
+
+If you want to add basic authentication support to your GraphQL requests you can
+do so by adding a custom `Authorization` HTTP header to your GraphiQL requests.
+
+If you are using the [GraphiQL macOS app](https://github.com/skevy/graphiql-app)
+this can be done from "Edit HTTP Headers". The `/dev/graphiql` implementation
+does not support custom HTTP headers at this point.
+
+Your custom header should follow the following format:
+
+```
+# Key: Value
+Authorization: Basic aGVsbG86d29ybGQ=
+```
+
+`Basic` is followed by a [base64 encoded](https://en.wikipedia.org/wiki/Base64)
+combination of your username, colon and password. The above example is `hello:world`.
+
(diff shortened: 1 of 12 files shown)


robbieaverill:
Dear Codecov.io - I can't easily test those uncovered lines, because the BasicAuth class is called from the parent of the GraphQL controller and can't be mocked.

I don't like this change a whole lot anyway, so I'm not super willing to write a long winded test for functionality that is tested elsewhere anyway. :)

Best, Robbie

README.md
```diff
@@ -457,6 +457,62 @@ This tool is available in **dev mode only**. It can be accessed at
...
+are performing requests from an anonymous/external application you may need to
+authenticate before you can complete a request.
+
+Please note that when implementing GraphQL resources it is the developer's
+responsibility to ensure that permission checks are implemented wherever
+resources are accessed.
+
+### Basic Authentication
+
+Silverstripe has built in support for basic authentication, and it can be configured
```
chillu: I'd call it "HTTP basic authentication", and link to the wikipedia entry: https://en.wikipedia.org/wiki/Basic_access_authentication. Too easy to misread as "some non-descript form of non-advanced authentication" otherwise :D

README.md
```diff
@@ -457,6 +457,62 @@ This tool is available in **dev mode only**. It can be accessed at
...
+does not support custom HTTP headers at this point.
+
+Your custom header should follow the following format:
+
+```
+# Key: Value
+Authorization: Basic aGVsbG86d29ybGQ=
+```
+
+`Basic` is followed by a [base64 encoded](https://en.wikipedia.org/wiki/Base64)
```
chillu: I'd just add the `bash` example from the wikipedia page here, for ease of use.

chillu: And add a note that authentication is transferred in plain text via this method, so we strongly recommend using TLS for non-development use.

robbieaverill:
@chillu as per our chat on Slack I've re-implemented the authentication structure to be less focussed on the controller.

Now, the controller calls the "auth handler" to decide if any authentication requirements need to be met.

The Handler will load a configurable list of authenticators in a priority order and call the first one's `authenticate` method. If successfully authenticated it will set the Member returned to the Manager for use in the context later, or throw an exception on failure.

Authenticators themselves will implement the `AuthenticatorInterface` and will be given the current request object if they need it.

This is a bit more future proof, and will allow us to plug in an OAuth 2 interface more easily.

robbieaverill:
**Note:** squash on merge and set a nice commit message :-) or I'll rebate first. The previous commits are left here for brevity.

robbieaverill:
## Use case - for extensibility (example only)

mysite configuration:
```yml
SilverStripe\GraphQL:
  authenticators:
    - class: MyProject\GraphQL\DevOnlyAuthenticator
      priority: 100
```

Authenticator:
```php
<?php

namespace MyProject\GraphQL;

use SilverStripe\Control\HTTPRequest;
use SilverStripe\GraphQL\Auth\AuthenticatorInterface;
use SilverStripe\Security\Member;

class DevOnlyAuthenticator implements AuthenticatorInterface
{
    public function authenticate(HTTPRequest $request)
    {
        return Member::get()->filter('Email', $request->getHeader('Email'))->first();
    }
}
```

Then add `Email: my-cms@email-address.com` as a custom header in the GraphiQL app.

src/Auth/Handler.php
```diff
@@ -0,0 +1,133 @@
...
+     * @var Manager
+     */
+    protected $manager;
+
+    /**
+     * @param Manager
+     */
+    public function __construct($manager)
+    {
+        $this->setManager($manager);
```
chillu: Making `Manager` a dependency in this class seems a bit overkill - what's the reasoning? I would've just returned the member from `requireAuthentication()` and set the member in `Controller`? You can infer authentication success from the presence of a member, same as with `BasicAuth::requireLogin()`

robbieaverill: Yep- good call

robbieaverill: @chillu much simpler without it! Thanks for pointing that out. Update pushed.

robbieaverill:
* Squashed

src/Controller.php
```diff
@@ -43,9 +48,13 @@ public function index(HTTPRequest $request)
             $variables = json_decode($variables, true);
         }
 
-        $manager = $this->getManager();
+        $this->setManager($manager = $this->getManager());
```
chillu: Why was this necessary? `setManager()` doesn't do anything but `$this->manager = $manager`, which is already done in `getManager()`?

PR #57, opened 2017-01-16
ENHANCEMENT Provide default authentication rules for /graphql path

### LINKED ISSUE #51: Secure endpoint by default
At the moment, you can run introspection queries on GraphQL without any authentication. Which is fine if you've got a public API, but not necessarily a good default. While this doesn't expose any data directly, it leaks information about installed modules etc - so is a minor security issue.

Secure the endpoint by permission code instead. Permission codes should be defined on the schema configuration, since we'll eventually allow different schemas with different permissions (e.g. a stripped down "public API" with anonymous introspection capabilities)

tractorcow:
Fixes #51

### DIFF
--- a/_config/routes.yml
+++ b/_config/routes.yml
@@ -3,10 +3,11 @@ Name: graphqlroutes
 ---
 SilverStripe\Control\Director:
   rules:
-    # Default admin route stage is draft
     graphql:
       Controller: 'SilverStripe\GraphQL\Controller'
+      # @internal - Experimental config
+      # @todo - move this to a per-schema configuration, and simply register the named schema for this endpoint
+      # https://github.com/silverstripe/silverstripe-graphql/issues/52
       Stage: Stage
       Permissions:
         CMS_ACCESS: CMS_ACCESS
-      Authenticator: SilverStripe\GraphQL\Auth\MemberAuthenticator
--- a/_config/routes.yml
+++ b/_config/routes.yml
@@ -7,3 +7,6 @@ SilverStripe\Control\Director:
     graphql:
       Controller: 'SilverStripe\GraphQL\Controller'
       Stage: Stage
+      Permissions:
+        CMS_ACCESS: CMS_ACCESS
+      Authenticator: SilverStripe\GraphQL\Auth\MemberAuthenticator
--- a/src/Auth/BasicAuthAuthenticator.php
+++ b/src/Auth/BasicAuthAuthenticator.php
@@ -4,6 +4,7 @@ namespace SilverStripe\GraphQL\Auth;
 
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Security\BasicAuth;
+use SilverStripe\Security\Member;
 
 /**
  * An authenticator using SilverStripe's BasicAuth
@@ -12,11 +13,40 @@ use SilverStripe\Security\BasicAuth;
  */
 class BasicAuthAuthenticator implements AuthenticatorInterface
 {
-    /**
-     * {@inheritDoc}
-     */
+
     public function authenticate(HTTPRequest $request)
     {
-        return BasicAuth::requireLogin('Restricted resource');
+        $member = BasicAuth::requireLogin('Restricted resource');
+        if ($member instanceof Member) {
+            return $member;
+        }
+        return null;
+    }
+
+    public function isApplicable(HTTPRequest $request)
+    {
+        if ($this->hasAuthHandler('HTTP_AUTHORIZATION')
+            || $this->hasAuthHandler('REDIRECT_HTTP_AUTHORIZATION')
+        ) {
+            return true;
+        }
+        if (isset($_SERVER['PHP_AUTH_USER']) && isset($_SERVER['PHP_AUTH_PW'])) {
+            return true;
+        }
+        if (Member::currentUserID()) {
+            return true;
+        }
+        return false;
+    }
+
+    /**
+     * Check for $_SERVERVAR with basic auth credentials
+     *
+     * @param $servervar
+     * @return bool
+     */
+    protected function hasAuthHandler($servervar)
+    {
+        return isset($_SERVER[$servervar]) && preg_match('/Basic\s+(.*)$/i', $_SERVER[$servervar]);
     }
 }
--- a/src/Auth/Handler.php
+++ b/src/Auth/Handler.php
@@ -5,7 +5,7 @@ namespace SilverStripe\GraphQL\Auth;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse_Exception;
 use SilverStripe\Core\ClassInfo;
-use SilverStripe\Core\Config\Config;
+use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\ORM\ValidationException;
 use SilverStripe\Security\Member;
@@ -18,6 +18,15 @@ use SilverStripe\Security\Member;
  */
 class Handler
 {
+    use Configurable;
+
+    private static $authenticators = [
+        [
+            'class' => BasicAuthAuthenticator::class,
+            'priority' => 10,
+        ]
+    ];
+
     /**
      * If required, enforce authentication for non-session authenticated requests. The Member returned from the
      * authentication method will returned for use in the OperationResolver context.
@@ -31,7 +40,7 @@ class Handler
      */
     public function requireAuthentication(HTTPRequest $request)
     {
-        $authenticator = $this->getAuthenticator();
+        $authenticator = $this->getAuthenticator($request);
         if (!$authenticator) {
             return false;
         }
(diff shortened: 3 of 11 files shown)


sminnee:
My assumption @tractorcow is that this relies on the session cookie being set in the GraphQL API. Since this is currently being primarily used for communication between the server and a client running on the same host, this will work.

In the long run it would probably be better to have some kind of default authenticator based on a session token of some kind, but this is more work.

I'm inclined to accept this for alpha4 and look to build an alternative default authenticator later.

_config/routes.yml
```diff
@@ -6,4 +6,7 @@ SilverStripe\Control\Director:
     # Default admin route stage is draft
     graphql:
       Controller: 'SilverStripe\GraphQL\Controller'
-      Stage: Stage
\ No newline at end of file
+      Stage: Stage
+      Permissions:
+        CMS_ACCESS: CMS_ACCESS
+      Authenticator: SilverStripe\GraphQL\Auth\MemberAuthenticator
```
sminnee: It's not clear to me why this is mapped in here and not in `SilverStripe\GraphQL\Auth\Handler.authenticators`.

Having 2 separate configuration points just leads to confusion.

src/Auth/Handler.php
```diff
@@ -47,29 +56,46 @@ public function requireAuthentication(HTTPRequest $request)
...
+        }
+
+        // Get list of default authenticators
+        $authenticators = $this->config()->get('authenticators');
         if (empty($authenticators)) {
             return false;
         }
 
+        // Build authenticator from first class
         $this->prioritiseAuthenticators($authenticators);
```
sminnee: It seems a bit odd to have a priority selection mechanism on top of the priority selection mechanism that's built into the config API. Admittedly, it's not strictly part of this PR, but I wonder whether authenticators would be better represented as a list of service references with `%$` – it would be more in keeping with other parts of the system.

src/Auth/Handler.php
```diff
@@ -47,29 +56,46 @@ public function requireAuthentication(HTTPRequest $request)
...
-        $authenticator = false;
         foreach ($authenticators as $authenticatorConfig) {
-            if (!ClassInfo::classImplements($authenticatorConfig['class'], AuthenticatorInterface::class)) {
-                throw new ValidationException(
-                    sprintf('%s must implement %s!', $authenticatorConfig['class'], AuthenticatorInterface::class)
-                );
-            }
-            $authenticator = Injector::inst()->get($authenticatorConfig['class']);
-            break;
+            return $this->buildAuthenticator($authenticatorConfig['class']);
```
sminnee: Authenticators could use two methods:

 * Process my authentication information and give me the current `Member`. This is currently `authenticate()` but I wonder if `getCurrentMember()` would be more self-documenting?
 * A method to indicate "does this request seem to even be trying to use this authentication mechanism?" In the case of MemberAuthenticator, this might return true if the session cookie was set. This might be something like `isApplicable()`.

This would mean that the iteration through different authenticators would take the following form.

 * Iterate through authenticators until `isApplicable()` returns true
 * Call `getCurrentMember()` on that authenticator

This would make it straight forward to implement logic of the form:

 * If basic-auth credentials are provided, use them
 * If an X-Session-Token header is provided, use that
 * Otherwise, use whatever is in the session cookie

src/Auth/Handler.php
```diff
@@ -47,29 +56,46 @@ public function requireAuthentication(HTTPRequest $request)
...
-        return $authenticator;
+        return false;
+    }
+
+    /**
+     * @param string $authenticator
+     * @return AuthenticatorInterface
+     * @throws ValidationException
+     */
+    protected function buildAuthenticator($authenticator)
```
sminnee: Again, if authenticators were specified as an array of named services this would be handled by Injector

src/Controller.php
```diff
@@ -49,10 +51,25 @@ public function index(HTTPRequest $request)
         $this->setManager($manager = $this->getManager());
 
         try {
+            // Check authentication
             $member = $this->getAuthHandler()->requireAuthentication($request);
             if ($member) {
                 $manager->setMember($member);
             }
+
```
sminnee: This will lock the graphql API to be accessible only by users with CMS_ACCESS permissions in a default install. This is an important piece of security neatness, but we'd want to elaborate in the docs (1 line will do for now, even if it's 1 line in the description of the "GraphQL docs" ticket) about how to enable the graphql for users beyond this.

src/Auth/Handler.php
```diff
@@ -47,29 +56,46 @@ public function requireAuthentication(HTTPRequest $request)
...
+     * @param HTTPRequest $request
+     * @return false|AuthenticatorInterface
+     * @throws ValidationException
      */
-    public function getAuthenticator()
+    public function getAuthenticator(HTTPRequest $request = null)
     {
-        $authenticators = Config::inst()->get('SilverStripe\GraphQL', 'authenticators');
+        // Request-specific authenticator
+        $authenticator = $request ? $request->param('Authenticator') : null;
```
chillu: I think we should infer this from the provided information, and just run through registered authenticators by priority until one returns a member. And if it's intended as a debugging capability, I think it's overkill.

src/Auth/Handler.php
chillu: Ah yep, I had the same feedback above, so +1 on iterating through authenticators

chillu:
> My assumption @tractorcow is that this relies on the session cookie being set in the GraphQL API. 

That's currently the case if you're sharing an already started CMS session. Alternatively, you can make stateless requests through `BasicAuthAuthenticator`, and eventually through other systems like OAuth.

> It seems a bit odd to have a priority selection mechanism on top of the priority selection mechanism that's built into the config API. Admittedly, it's not strictly part of this PR, but I wonder whether authenticators would be better represented as a list of service references with %$ – it would be more in keeping with other parts of the system.

Agreed. @robbieaverill the priorities were your approach - do you see any disadvantages of using the config API instead? Created a ticket: https://github.com/silverstripe/silverstripe-graphql/issues/58

> In the long run it would probably be better to have some kind of default authenticator based on a session token of some kind, but this is more work.

@robbieaverill is working on an OAuth2 module for SilverStripe, which could also be wrapped into an authenticator for GraphQL here. He's started by bridging the HTTP request PSR-7 to SS: https://github.com/robbieaverill/psr7-adapters. Which will allow the use of http://oauth2.thephpleague.com/

robbieaverill:
@chillu 

>Agreed. @robbieaverill the priorities were your approach - do you see any disadvantages of using the config API instead? Created a ticket: #58

No, that's fine. As long as they can be ordered in some way so that the BasicAuthAuthenticator can easily be replaced in priority by something meatier, it's fine whichever way it works. Native ordering makes more sense.

>In the long run it would probably be better to have some kind of default authenticator based on a session token of some kind, but this is more work.
@robbieaverill is working on an OAuth2 module for SilverStripe, which could also be wrapped into an authenticator for GraphQL here. He's started by bridging the HTTP request PSR-7 to SS: https://github.com/robbieaverill/psr7-adapters. Which will allow the use of http://oauth2.thephpleague.com/

Yeah - also @sminnee - your [API key module](https://github.com/sminnee/silverstripe-apikey) would be a good intermediary step between basic auth and fully blown OAuth 2.0. The only change would be disabling/ability to disable the RequestFilter and add an AuthenticatorInterface implementation for GraphQL only.

chillu:
@tractorcow said: "I think the security configuration really belongs to the schema, not the endpoint"

I'd say it belongs on the endpoint/routing, to avoid coupling of HTTP transport to the schema (specifically, to `Manager`). It'll likely be quite common to have a private, authenticated `/graphql` endpoint for the CMS, and a more limited `/graphql-public` one for your own SPA API (with a different, narrower schema). Currently I can't see a reason why you'd want to have the same schema exposed through different routes, but even without that use case I think there's value in decoupling HTTP from `Manager`.

tractorcow:
Updated, attempting to address each point above.

src/Auth/BasicAuthAuthenticator.php
```diff
@@ -12,11 +13,40 @@
...
+        $member = BasicAuth::requireLogin('Restricted resource');
+        if ($member instanceof Member) {
+            return $member;
+        }
+        return null;
+    }
+
+    public function isApplicable(HTTPRequest $request)
+    {
+        if ($this->hasAuthHandler('HTTP_AUTHORIZATION')
```
chillu: Why this additional check? Shouldn't this be abstracted away to PHP, which populates `PHP_AUTH_*` in `$_SERVER`?

src/Auth/BasicAuthAuthenticator.php
```diff
@@ -12,11 +13,40 @@
...
+            || $this->hasAuthHandler('REDIRECT_HTTP_AUTHORIZATION')
+        ) {
+            return true;
+        }
+        if (isset($_SERVER['PHP_AUTH_USER']) && isset($_SERVER['PHP_AUTH_PW'])) {
+            return true;
+        }
+        if (Member::currentUserID()) {
```
chillu: Hmm, technically basic auth handling isn't "applicable" when you have a current user (set through PHP sessions). Given you have a separate `MemberAuthenticator`, why was this necessary?

chillu: For example, this makes it impossible to enforce basic auth, without falling back to PHP session

tractorcow: Basic auth failure won't bubble up properly to the user response though, will it? So may as well just fail if not provided.

chillu:
Merging to get the release out, created a follow up ticket on how to call `isApplicable()`: https://github.com/silverstripe/silverstripe-graphql/issues/60

Issue #60, opened 2017-01-17
Don't encode PHP session knowledge in basic auth handler


chillu:
See https://github.com/silverstripe/silverstripe-graphql/pull/57#discussion_r96323371

 * [x] Member handler should be checked last
 * [x] BasicAuth handler should not inspect PHP session through `Member::currentUser()` in `isApplicable()`
 * ~~All handlers returning `isApplicable()` = `true` should be run, but the first one returning a valid member cuts short the loop~~
 * [ ] Handlers have a way to communicate auth failures to the client ("invalid token", "set basic auth challenge")

robbieaverill:
@chillu - not sure I understand the request here:

>All handlers returning `isApplicable()` = `true` should be run, but the first one returning a valid member cuts short the loop

Can you clarify?

The current behaviour is to loop authenticators until one of them says that it is applicable, then stop and authenticate with it.

Are we changing this to loop all authenticators that are applicable and use the first one that successfully authenticates a request?

chillu:
Hmmm, you're right - a basic auth handler will only say it's applicable if basic auth headers are found in the request. And in that case, it should be the *only* one handling authentication. At this point its irrelevant if the request also contains a PHP session cookie, provided you have defined basic auth with a higher priority. I've updated the ticket accordingly, so "communicate auth failures" is the only outstanding item here, right?

robbieaverill:
Cool- yep!

PR #63, opened 2017-01-19
FIX Member handler should be checked last

robbieaverill:
Priority (in its current implementation) is ordered from highest to lowest in order of which authenticators are checked.

`MemberAuthenticator` should be the last class to be checked as it is the fallback authentication option.

Contributes to #60

### DIFF
--- a/src/Auth/Handler.php
+++ b/src/Auth/Handler.php
@@ -32,11 +32,11 @@ class Handler
     private static $authenticators = [
         [
             'class' => MemberAuthenticator::class,
-            'priority' => 20,
+            'priority' => 10,
         ],
         [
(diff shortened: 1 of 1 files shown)


tractorcow:
Will this avoid triggering a basic-auth check header being sent if the member can be logged in automatically via remember-token?

robbieaverill:
@tractorcow this would mean that if basic auth is configured (in GraphQL) that it would be checked first and return the member for context, otherwise try the current member session as a fallback so that same browser use (eg. GraphiQL) would work.

If basic auth were enabled on the whole site then it would actually log you in and you'd end up being authenticated via member in GraphQL rather than basic auth.

PR #64, opened 2017-01-19
FIX Do not check Member context in BasicAuthAuthenticator

robbieaverill:
Contributes to #60


tractorcow:
I'll leave this to @chillu to review.

PR #69, opened 2017-02-06
API Added CORS pre-flight handler (fixes #66)

### LINKED ISSUE #66: Make CORS configurable
GraphQL APIs will be consumed by third party apps on other hosts, e.g. sending XHR requests from your custom web app to a centralised SaaS solution running a SilverStripe-powered GraphQL endpoint. This requires [Cross Origin Resource Sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing)

See https://github.com/colymba/silverstripe-restfulapi/blob/master/doc/RESTfulAPI.md#cors for an example.

PapaBearNZ:
Added code to index() to detect a CORS preflight OPTIONS request.
Added getCorsResponse() to append the correct headers for a CORS request. 
Also checks the origin is an allowed domain for cross-origin requests. 
Included unittests for invalid origin, valid origin, OPTIONS request type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1431,7 +1431,7 @@ Authorization: Basic aGVsbG86d29ybGQ=
 combination of your username, colon and password. The above example is `hello:world`.
 
 **Note:** Authentication credentials are transferred in plain text when using HTTP
-basic authenticaiton. We strongly recommend using TLS for non-development use.
+basic authentication. We strongly recommend using TLS for non-development use.
 
 Example:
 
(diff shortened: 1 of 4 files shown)


PapaBearNZ:
Found the bug in this pull request.  Fixing.  I'll submit a new PR with the fix in place.

PapaBearNZ:
Oops - hit close and comment in error.

PapaBearNZ:
I think CORS should be disabled by default.  This is in line with the idea of assuming all things are as locked down as possible unless the dev chooses to open them up.  If no one has any complaints, I'll update the default config.  This should then (in conjunction with the above modification) fix the travis build issues.

src/Controller.php
```diff
@@ -86,8 +102,12 @@ public function index(HTTPRequest $request)
             ];
         }
 
-        return (new HTTPResponse(json_encode($result)))
-            ->addHeader('Content-Type', 'application/json');
+        $response = new HTTPResponse(json_encode($result));
+        if ($corsEnabled) {
+            $response = $this->getCorsResponse($request, $response);
```
chillu: Good catch about including CORS headers in the actual GraphQL response, that seems to be best practice. If you follow https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests, the actual HTTP response (not the `OPTIONS` call) only has the `Access-Control-Allow-Origin` header - but the spec says "An HTTP response to a CORS request can include the following headers", listing all available headers (https://fetch.spec.whatwg.org/#http-responses) - so I *think* we're good

src/Controller.php
```diff
@@ -123,4 +143,51 @@ public function getAuthHandler()
...
     }
+
+    /**
+     * Process the CORS config options and add the appropriate headers to the response.
+     *
+     * @param HTTPRequest $request
+     * @param HTTPResponse $response
+     * @return HTTPResponse
+     */
+    public function getCorsResponse(HTTPRequest $request, HTTPResponse $response = null)
```
chillu: The dual use with/without a `$response` argument is a bit confusing. Can you create `new HTTPResponse()` outside of this method, make `$response` a mandatory argument, and rename the method to `addCorsHeaders()`?

chillu:
Could you have a look at the failed tests please? Just restarted the build, in case they were temporary

chillu:
> I think CORS should be disabled by default. This is in line with the idea of assuming all things are as locked down as possible unless the dev chooses to open them up. If no one has any complaints, I'll update the default config. This should then (in conjunction with the above modification) fix the travis build issues.

Agreed. So you're planning to set `Enabled: false`, meaning an `OPTIONS` request won't receive any special responses in the controller by default, right?

src/Controller.php
PapaBearNZ: Understood.  I'll make the necessary changes.

PapaBearNZ:
@chillu Re the CORS Enabled comment.  You are correct.  The errors in the tests are due to CORS being enable by default but having no Allowed Origins defined, the CORS response was correctly rejecting the request.  By setting CORS to disabled by default, This check never happens and the response is processed normally without the 403 status.

src/Controller.php
```diff
@@ -51,6 +51,10 @@ public function index(HTTPRequest $request)
             // Process the CORS config and add appropriate headers.
             $response = new HTTPResponse();
             return $this->addCorsHeaders($request, $response);
+        } elseif (!$corsEnabled && $request->httpMethod() == 'OPTIONS') {
+            // CORS is disabled but we have received an OPTIONS request.  This is not a valid request method in this
+            // situation.  Return a 405 Method Not Allowed response.
```
chillu: The comments are getting a little bit verbose (repeating the status code), but otherwise all good :)

chillu:
Sweet, now it's just missing some docs for the README!

README.md
```diff
@@ -1449,6 +1449,107 @@ php -r 'echo base64_encode("hello:world");'
...
+
+By default CORS is disabled in the GraphQL Server. This can be easily enabled via Yaml
+
+```yaml
+   SilverStripe\GraphQL:
+     cors:
+       Enabled: true
+```
+
+Once you have enabled CORS you can then control 4 new headers in the HTTP Response.
```
chillu: "4" should be spelled out as "four" in most common grammar guidelines

README.md
```diff
@@ -1449,6 +1449,107 @@ php -r 'echo base64_encode("hello:world");'
...
+(please refer to the YAML below for examples)
```
chillu: "please refer" sentence is too verbose, given the YAML is two lines below it

chillu:
Can you rebase on master please? There’s conflicts. Can you please link the first mention of CORS to https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS? I found that guide to be helpful. Otherwise the README looks great!

README.md
PapaBearNZ: noted and removed.

README.md
PapaBearNZ: noted and corrected.

README.md
```diff
@@ -1449,6 +1440,126 @@ php -r 'echo base64_encode("hello:world");'
 # aGVsbG86d29ybGQ=
 ```
 
+### Defining your own authenticators
```
chillu: Why did you add this section? Merge error?

PapaBearNZ: Argh - That was to resolve a merge conflict after the rebase.  It's actually @robbieaverill 's work that's somehow gotten credited to me :(  Not sure how to fix that.  (lines 1443-1461)

robbieaverill: Looks like a dodgy merge

PapaBearNZ: The rebase off upstream master worked well but then when trying to push up my branch there were conflicts there as well.  This section got caught in the crossfire :(

What do you suggest to resolve this guys?  Should I create a new branch off master, squash my commit tree and then do one single commit as part of a replacement PR?

chillu: Whatever works for you - the end goal is that only relevant patches make it into this PR. We can squash on merge, or you can squash - either way is fine.

PapaBearNZ: How does this look?

https://github.com/silverstripe/silverstripe-graphql/compare/master...PapaBearNZ:66-cors-handling

PR #70, opened 2017-02-07
POST queries with variables entity causes error

PapaBearNZ:
When submitting a query to GraphQL via a POST message, in the case where the 'variables' entity is non-null, an extra call to json_decode is made which is unnecessary and causes a PHP error because the entity is already decoded earlier in the block.

### DIFF
--- a/src/Controller.php
+++ b/src/Controller.php
@@ -42,7 +42,7 @@ class Controller extends BaseController
             $rawBody = $request->getBody();
             $data = json_decode($rawBody ?: '', true);
             $query = isset($data['query']) ? $data['query'] : null;
-            $variables = isset($data['variables']) ? json_decode($data['variables'], true) : null;
+            $variables = isset($data['variables']) ? $data['variables'] : null;
         } else {
             $query = $request->requestVar('query');
             $variables = json_decode($request->requestVar('variables'), true);


robbieaverill:
Looks fine to me - personally I'd really like that block of logic to be moved into its own method.

PapaBearNZ:
@robbieaverill Can you explain further?  Do you mean separating POST vs GET requests?

robbieaverill:
@PapaBearNZ sure - (in my opinion) [this block of code](https://github.com/PapaBearNZ/silverstripe-graphql/blob/3f131e799492afd9566e90ec5fbb907444809aaf/src/Controller.php#L39-L49) unnecessarily increases the complexity of that method.

It could probably be `$variables = $this->getVariables($request)` and then we can unit test the ... out of the new method to ensure this never happens again (third time we've had problems with this block of code :))

PapaBearNZ:
So something like `list($query, $variables) = $this->parseRequest($request);` perhaps?

chillu:
Looks like the [GraphiQL OSX App](https://github.com/skevy/graphiql-app) used to have an outdated and/or buggy graphiql dependency which doubled encoded the `variables` key in the POST payload as a JSON string (rather than leaving it as JSON). I've reinstalled the app, and this has since been fixed.

Regarding the refactoring, I think it's a good idea, but doesn't block the bugfix at hand

PR #76, opened 2017-02-14
Allow wildcard in allowed-origin CORS settings.

bummzack:
Currently, a wildcard as `Allow-Origin` seems to be ignored. This PR fixes this.


sminnee:
This caused a failure by the time it actually got merged! @bummzack @chillu 

https://travis-ci.org/silverstripe/silverstripe-graphql/jobs/215377609

chillu:
I've fixed this test failure, it was an issue with the config being renamed since that PR was opened.

PR #77, opened 2017-02-14
FIX: Missing edge case for CORS.

PapaBearNZ:
When CORS OPTIONS request is missing the Origin header the
    implementation wrongly sends a valid CORS response.
    This is fixed to now send a more correct 403 response.


chillu:
Resolved a conflict, but builds are breaking.

robbieaverill:
At some stage recently the config namespace changed from `SilverStripe\GraphQL` to `SilverStripe\GraphQL\Controller` - fix at #87. This PR will need to be updated in the same way.

chillu:
Fixed up the config namespace and merged

PR #78, opened 2017-02-15
Force $variables to cast as array

unclecheese:
GraphiQL passes variables as empty string and `GraphQL\Validator\Rules\QueryComplexity::setRawVariableValues()` chokes on a non-array value.


robbieaverill:
@chillu the second change in this PR breaks, will need to be patched. You can't decode json that's an array!

Issue #22, opened 2016-12-14
Scaffold object type inheritance


chillu:
Related to [Scaffold input type inheritance](https://github.com/silverstripe/silverstripe-graphql/issues/23)

In line with ORM conventions, we want to expose fields from all subclasses when querying the base class (e.g. `SiteTree`). In contrast to the ORM, GraphQL exposes a type or interface for fields of the returned objects. While we could flatten all fields to the base type, it means a lot of fields which don't make sense without further context:

```
type SiteTree {
    id
    created
    title
    content
    myField <-- from Page
    redirectionType <-- from RedirectorPage
    externalUrl <-- from RedirectorPage
}
```

Instead, I propose we create distinct types for each subclass, each of which is a superset of its parent class.

```
type SiteTree {
    id
    created
    title
    content
}

type Page {
    id
    created
    title
    content
    myField
}

type RedirectorPage {
    id
    created
    title
    content
    myField
    redirectionType
    externalUrl
}
```

This structure allows us to declare multiple types as a [union](http://facebook.github.io/graphql/#sec-Unions) in queries:

```
union SiteTreeTypes = SiteTree | Page | RedirectorPage

type readSiteTree {
  [SiteTreeTypes]
}
```

The downside is that queries will always have to declare on which type they're expecting fields:

```
query readSiteTree() {
  ...on SiteTree {
    id
    created
    title
    content
  }
  ...on Page {
    myField
  }
  ...on RedirectorPage {
    redirectionType
    externalUrl
  }
}
```

While a type [can implement](https://github.com/facebook/graphql/issues/48#issuecomment-201089763) more than one interface, a field (query response) can only declare a interface response type (or a union of types) - see [discussion on facebook/graphql](https://github.com/facebook/graphql/issues/48). We could use a single [interface](http://facebook.github.io/graphql/#sec-Interfaces), but there's no "union of interfaces" in GraphQL. 

Sam [suggested](https://gist.github.com/chillu/d0069fb0123d0480761e6ba5659bd2e7) types implementing multiple interfaces, but given the lack of "union of interface" support, I don't see much value in this. Also, it we can't have all types implement a `SiteTreeInterface` to get around writing `...on SiteTree` for fields on the base type - it's either a single interface, or a union of types (verified through a [modified ReadFileQueryCreator](https://gist.github.com/chillu/0461fc0c21eb746b12466e548eae8515)).

[...]

PR #79, opened 2017-02-15
Feature/scaffolding inheritance unions

unclecheese:
* Minor, PSR-2 formatting
* Resolve https://github.com/silverstripe/silverstripe-graphql/issues/22


sminnee:
OK looks like the whitespace changes are moving *to* PSR-2. We should get a linter into travis and an editorconfig if we don't already

PR #81, opened 2017-02-27
API Update graphql to use new config

tractorcow:
Update for https://github.com/silverstripe/silverstripe-framework/pull/6641

### DIFF
--- a/tests/ControllerTest.php
+++ b/tests/ControllerTest.php
@@ -2,6 +2,8 @@
 
 namespace SilverStripe\GraphQL\Tests;
 
+use PHPUnit_Framework_MockObject_MockBuilder;
+use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Dev\SapphireTest;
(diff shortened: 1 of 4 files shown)


tests/ControllerTest.php
```diff
@@ -60,17 +64,12 @@ public function testGetGetManagerPopulatesFromConfig()
...
     {
-        Config::inst()->update('SilverStripe\\Control\\Director', 'environment_type', 'live');
+        Director::set_environment_type('live');
 
         $controller = new Controller();
-        $managerMock = $this->getMockBuilder(Schema::class)
+        /** @var Manager|PHPUnit_Framework_MockObject_MockBuilder $managerMock */
+        $managerMock = $this->getMockBuilder(Manager::class)
             ->setMethods(['query'])
-            ->setConstructorArgs([
```
chillu: Why did this need to be removed?

chillu:
Merging even with failing tests, since it blocks the upstream config merge which would allow it to pass.

tests/ControllerTest.php
tractorcow: This mock was incorrectly using a Schema in place of a Manager class.

PR #85, opened 2017-03-23
FIX Correct namespaces in test classes, add PSR-4 autoloader, sort some class imports

robbieaverill:
This pull request:

* implements a dev autoloader
* fixes inconsistent namespaces in test classes
* implements `SilverStripe\Dev\TestOnly` in test stubs
* fixes some PSR-2 compat in stubs
* rearranges some class imports a bit so they're alphabetically ordered

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,6 +19,11 @@
       "SilverStripe\\GraphQL\\": "src/"
     }
   },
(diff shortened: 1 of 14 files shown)


chillu:
Thanks Robbie!

PR #87, opened 2017-03-27
FIX Use FQ controller namespace for config in CORS tests

robbieaverill:


robbieaverill:
Failure unrelated

chillu:
Legend, thanks Robbie!

PR #88, opened 2017-03-27
FIX Update configuration namespace in example YAML

### LINKED ISSUE #84: Example code results in query error
Running the example code using the following query:

`/graphql/?query={readMembers{ID+FirstName+Email}}`

results in an error:

`{"data":null,"errors":[{"message":"Cannot query field \"readMembers\" on type \"Query\".","locations":[{"line":1,"column":2}]}]}`

This is after moving the whole example directory, removing the exclude manifest file and performing a /dev/build to create the example tables.

robbieaverill:
Resolves #84

### DIFF
--- a/examples/_config/config.yml
+++ b/examples/_config/config.yml
@@ -1,4 +1,4 @@
-SilverStripe\GraphQL:
+SilverStripe\GraphQL\Controller:
   schema:
     types:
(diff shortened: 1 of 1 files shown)


robbieaverill:
Tests are fixed in #87

PR #90, opened 2017-03-28
Use new-style travis config

chillu:
- Fixes build failures with missing versioned extension
- Introduces multi database builds


phpunit.xml
```diff
@@ -1,22 +1,26 @@
...
-        </testsuite>
-    </testsuites>
+<phpunit bootstrap="framework/tests/bootstrap.php" colors="true">
+
+    <testsuite name="Default">
+        <directory>tests</directory>
+    </testsuite>
+
+    <listeners>
+        <listener class="SilverStripe\Dev\TestListener" />
```
tractorcow: This class was deleted... not sure how this is passing.

.travis.yml
```diff
@@ -1,24 +1,40 @@
 language: php
 
+sudo: false
+
 env:
   global:
-    - DB=MYSQL CORE_RELEASE=master
+    - COMPOSER_ROOT_VERSION=1.0.x-dev
+    - CORE_RELEASE=master
```
tractorcow: I think CORE_RELEASE can be retired now.

composer.json
```diff
@@ -21,12 +21,15 @@
...
     "psr-4": {
-      "SilverStripe\\GraphQL\\Tests\\": "tests/"
+      "SilverStripe\\GraphQL\\Tests\\": "tests/php/"
     }
   },
   "config": {
     "process-timeout": 600
   },
+  "scripts": {
+    "lint": "phpcs --standard=tests/phpcs/ruleset.xml src/ tests/"
```
tractorcow: I've found that if you rename this file to `phpcs.xml.dist` and put it in the root it's found automatically and doesn't need to be specified.

phpunit.xml
tractorcow: Maybe the file needs to be named `phpunit.xml.dist`?

composer.json
tractorcow: See CMS for example.

sminnee: +1 to using `phpcs.xml.dist` as this will mean that IDEs pick it up too

.travis.yml
```diff
@@ -1,24 +1,40 @@
...
     - php: 7.0
+      env: DB=MYSQL PHPUNIT_TEST=1
     - php: 7.1
+      env: DB=MYSQL PDO=1 PHPUNIT_COVERAGE_TEST=1
 
 before_script:
-    - git clone git://github.com/silverstripe/silverstripe-travis-support.git ~/travis-support
-    - php ~/travis-support/travis_setup.php --source `pwd` --target ~/builds/ss
-    - cd ~/builds/ss
+  - printf "\n" | pecl install imagick
```
sminnee: Is this needed for graphql testing?

.travis.yml
```diff
@@ -1,24 +1,40 @@
...
+    - COMPOSER_ROOT_VERSION=1.0.x-dev
+    - CORE_RELEASE=master
 
 matrix:
+  fast_finish: true
   include:
     - php: 5.6
+      env: DB=MYSQL PHPUNIT_TEST=1 PHPCS_TEST=1
+    - php: 5.6
+      env: DB=SQLITE PHPUNIT_TEST=1
```
sminnee: Should we ditch SQLITE builds in favour of PGSQL ones?

.travis.yml
sminnee: Yep

composer.json
chillu: Right, I was following the framework module "best practice" heh - so much fragmentation in build config!

.travis.yml
chillu: Nope, I've removed it

.travis.yml
chillu: As discussed on Slack, we agree that SQLite and PG can be run through the nightly installer build only

.travis.yml
chillu: Done

phpunit.xml
chillu: Done

sminnee:
Automated codecov/project is benign AFAIK — I'm happy for this to be merged.

PR #94, opened 2017-05-05
Add a table of contents to the readme

flamerohr:
I admit I got lost a few times scrolling through all the titles + descriptions + example code blocks + different styles of code blocks for the same example

This makes it more clear what's available to learn about and with the links in place, you can go straight to where you need.

Also fixed some casing of words in the title.


unclecheese:
Nice. This is long overdue. There was discussion about migrating all of this to a `docs/` section, but the consensus was that everyone liked long READMEs. I'm not in that group. I do think we should revisit that soon, because the API surface is only going to get bigger as we approach stable.

unclecheese (approved):

Issue #47, opened 2017-01-08
Change lazy-loading of types (webonyx 0.8 blocker)


tractorcow:
As of https://github.com/webonyx/graphql-php/issues/35, we need to modify the way that types are lazy-loaded. This is a blocker to upgrading to 0.8 of the webonyx/graphql-php library.

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/474
- [ ] https://github.com/silverstripe/silverstripe-graphql/pull/98

PR #98, opened 2017-05-12
Fix types no longer return callbacks

flamerohr:
This opens up the option to upgrade to `graph-php@0.8`

Fixes https://github.com/silverstripe/silverstripe-graphql/issues/47

Dependencies https://github.com/silverstripe/silverstripe-asset-admin/pull/474

### DIFF
--- a/src/Scaffolding/Scaffolders/MutationScaffolder.php
+++ b/src/Scaffolding/Scaffolders/MutationScaffolder.php
@@ -32,9 +32,7 @@ class MutationScaffolder extends OperationScaffolder implements ManagerMutatorIn
         return [
             'name' => $this->operationName,
             'args' => $this->createArgs(),
-            'type' => function () use ($manager) {
-                return $manager->getType($this->typeName);
-            },
+            'type' => $manager->getType($this->typeName),
             'resolve' => $this->createResolverFunction(),
         ];
     }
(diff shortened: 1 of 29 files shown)


tractorcow:
Wow good job. I thought this work would be a lot harder than it was.

I suggest either @unclecheese or @chillu to review this given their proximity to this module.

chillu:
Yep, as you expected @tractorcow, it's not that easy. I've amended Chris' PR with the other uses where `type()` returns a function rather than a value, and you can see all kinds of unit tests failing.

unclecheese:
Yeah, this change is going to completely hose all the scaffolding stuff. I've been dreading this upgrade.

flamerohr:
Yeah, I didn't think this was the complete work... I was waiting for travis to run unit tests, as they seem to run even slower on my own computer.
I'm still trying out some other things like adding callback wrapper to the `'fields'` property in `ObjectType`.

What I've done so far is only tested with asset-admin currently, and I'm building on top of this as I go along :D Quite new to all these layers of building types and scaffolding in PHP, so it's taking me a bit of time to get the theory into my head.

src/Scaffolding/Scaffolders/MutationScaffolder.php
```diff
@@ -29,6 +29,11 @@ public function addToManager(Manager $manager)
      */
     public function scaffold(Manager $manager)
     {
+        $name = $this->typeName;
+        if (!$manager->hasType($name) && method_exists($this, 'generateInputType')) {
+            $manager->addType($this->generateInputType(), $name);
```
tractorcow: I think we need to move this to an interface, and then use `$this instanceof TheInterface` to promote good type cohesion. On the interface you can explain what this type does. :)

unclecheese: Interfaces can't use non-public methods. :(

PR #106, opened 2017-06-27
FIX namespaces in docs

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/5757


phalkunz:
https://github.com/silverstripe/silverstripe-reports/pull/77

PR #111, opened 2017-07-25
WIP: RFC/API: Refactor authentication to use middleware

robbieaverill:
This PR acts as a WIP RFC, which replaces the authentiation handler and AuthenticatorInterfaces with HTTPMiddleware implementations.

High level changes:

* `Auth\Handler` and `AuthenticatorInterface`s removed
* `SessionAuthenticatorMiddleware` and `BasicAuthenticatorMiddleware` added, the former of which is enabled by default
* `Manager` is now used as a singleton (instead of get/set to preserve the state in `Controller`), and handles the member context instead of `Handler`
* Construction of the `Manager` from configuration is moved to a middleware and stored as an injectable service
* Setting the versioned stage is moved to a middleware

The CORS configuration could probably be moved to a middleware implementation as well.

This change would remove a reasonable amount of complexity from the module, reduces coupling slightly and allows for other types of authentication middleware to be added more easily than the existing AuthenticatorInterface implementations.

If we're happy with this approach I'll add new tests as necessary (there's not much logic to test now!) and update the documentation to reflect the new changes.


_config/config.yml
```diff
@@ -33,3 +33,15 @@ SilverStripe\GraphQL\Controller:
     Allow-Headers: 'Authorization, Content-Type'
     Allow-Methods:  'GET, POST, OPTIONS'
     Max-Age:  86400  # 86,400 seconds = 1 day.
+
+SilverStripe\Core\Injector\Injector:
```
Firesphere: This seems to be unneeded if applied the right way? This looks like an addition that should not be needed?

src/Controller.php
```diff
@@ -69,18 +62,13 @@ public function index(HTTPRequest $request)
...
-            $member = $this->getAuthHandler()->requireAuthentication($request);
-            if ($member) {
-                $manager->setMember($member);
-            }
+            $manager = Manager::singleton();
 
             // Check authorisation
             $permissions = $request->param('Permissions');
             if ($permissions) {
+                $member = $manager->getMember();
```
Firesphere: Why `Manager::getMember` if `Security` has the `getCurrentMember` readily available?

src/Middleware/BasicAuthenticationMiddleware.php
```diff
@@ -0,0 +1,28 @@
...
+namespace SilverStripe\GraphQL\Middleware;
+
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+use SilverStripe\GraphQL\Manager;
+use SilverStripe\Control\Middleware\HTTPMiddleware;
+use SilverStripe\ORM\ValidationException;
+use SilverStripe\Security\BasicAuth;
+
+class BasicAuthenticationMiddleware implements HTTPMiddleware
```
Firesphere: Doesn't the Framework::BasicAuth already handle this? It seems a bit a double up?

src/Middleware/RegisterManagerMiddleware.php
```diff
@@ -0,0 +1,27 @@
...
+use SilverStripe\Core\Config\Config;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\GraphQL\Controller;
+use SilverStripe\GraphQL\Manager;
+
+/**
+ * Instantiates a GraphQL Manager class as a singleton with its own procedural configuration context
+ * so that it can be used further in the request
+ */
+class RegisterManagerMiddleware implements HTTPMiddleware
```
Firesphere: As above, I'm slightly confused as to what the added value is here to using a separate authentication method, while it is already in framework?

robbieaverill:
You're right! Now that middleware runs first we don't need any authenticators any more and can just ask for the current member for context, will test and update

Firesphere:
Heh, that's a great summary of my thoughts indeed, I don't believe the GraphQL module needs it's own authentication anymore with the given Framework Middlewares. 

It would also greatly clean up this module of unnecessary overhead I think?

robbieaverill:
Ok, so for now I'm going to leave this the way it is, because:

* We don't have a basic authentication handler that would be compatible with middleware based auth, since basic auth only provide a "log in" mechanism, but handles its own "login form". Other forms are not relevant, and I'm not sure at a glance whether the `Security` class is able to handle null returns from this method, since it and a few others are required by the `Authenticator` interface
* The doc block in `AuthenticationHandler` references a basic auth handler which doesn't exist, however seeing as the only implementations of this interface are cookie and session it looks like this is the wrong place to be putting basic auth anyway (?)
* It doesn't look like we can protect certain routes via configuration with basic auth at the moment, only theoretically apply it as global middleware (notes above). It could be done via `_config.php` but seems like it should be possible with middleware config in YAML
* I think the GraphQL controller would benefit from some of this middleware abstraction, especially going further than what's in this PR at the moment and separating CORS handling as well

That said, I agree that we should be able to remove the authenticators from this module very soon.

@Firesphere do you have some more insight into the differences between `Authenticator` (interface), `AuthenticatorInterface` and handlers?

My assumption is that we'd need to add an `AuthenticationHandler` (despite note above about it possibly being wrong) since the `AuthenticatorMiddleware` accepts these in `setAuthenticationHandlers`. If this is true, we could add a basic auth handler as a middleware wrapper and apply it via config. The handler might look like this:

```php
class BasicAuthenticationHandler implements AuthenticationHandler
{
    /**
     * Try to authenticate a member with basic authentication
     *
     * {@inheritDoc}
     */
    public function authenticateRequest(HTTPRequest $request)
    {
        try {
            return BasicAuth::requireLogin($request, 'Restricted');
        } catch (HTTPResponse_Exception $ex) {
            throw new ValidationException($ex->getMessage());
        }
    }

    public function logIn(Member $member, $persistent = false, HTTPRequest $request = null)
    {
        // no op
    }

    public function logOut(HTTPRequest $request = null)
    {
        // no op
    }
}
```

This can be applied via route specific config (using the code in this PR too) like so:

```yaml
  GraphQLController:
    class: SilverStripe\Control\Middleware\RequestHandlerMiddlewareAdapter
    properties:
      RequestHandler: %$SilverStripe\GraphQL\Controller
      Middlewares:
        registermanager: %$SilverStripe\GraphQL\Middleware\RegisterManagerMiddleware
        versioned: %$SilverStripe\GraphQL\Middleware\VersionedMiddleware
        basicauth: %$GraphQLAuthenticationMiddleware

  GraphQLAuthenticationMiddleware:
    class: SilverStripe\Security\AuthenticationMiddleware
    properties:
      AuthenticationHandler: %$GraphQLAuthenticationHandler

  GraphQLAuthenticationHandler:
    class: SilverStripe\Security\RequestAuthenticationHandler
    properties:
      Handlers:
        basicauthhandler: %$SilverStripe\Security\MemberAuthenticator\BasicAuthenticationHandler
```

So it is doable, but it's not pretty right now. I don't think I should have to re-define a set of core security components in injector config to be able to achieve this.

Perhaps this needs to be moved to a framework issue for further discussion...?

_config/config.yml
```diff
@@ -33,3 +33,15 @@ SilverStripe\GraphQL\Controller:
...
+  GraphQLController:
+    class: SilverStripe\Control\Middleware\RequestHandlerMiddlewareAdapter
```
unclecheese: We're trying to eliminate the use of arbitrary names for services and make them predictable, using the pattern `FQCN\Interface\Name.variant`, e.g. `League\Flysystem\Filesystem.public`.

In this case, you don't have an interface, so the ideal fix is to create one, but otherwise just use the base class. 
```yml
SilverStripe\Control\Middleware\RequestHandlerMiddlewareAdapter.graphql:
  class: SilverStripe\Control\Middleware\RequestHandlerMiddlewareAdapter
  ...
```

More info:

https://github.com/silverstripe/silverstripe-framework/issues/6642

robbieaverill: Sure, I think the second option is better personally - what do you think?

unclecheese:
Love the new `Manager` middleware. Having the manager as a first class service has huge implications for so much of the scaffolding API that requires a `$manager` instance to be passed around explicitly, creating a lot of unnecessary acrobatics. The next logical step would be a `ManagerAware` trait/interface that injects the inflated `Manager` instance on any classes that need it.

Overall, really nice cleanup. Keen to hear from @Firesphere on the auth stuff, as I'm still learning my way around it too.

src/Controller.php
robbieaverill: Because the member context for a request may not be the same as what is in Security

src/Middleware/BasicAuthenticationMiddleware.php
robbieaverill: It doesn't, but it needs to - logged as https://github.com/silverstripe/silverstripe-framework/issues/7268

src/Middleware/RegisterManagerMiddleware.php
robbieaverill: This is important actually - it's initialising the GraphQL manager as a service early on so it can be used as a singleton by the app, this is not related to authentication

src/Middleware/SessionAuthenticationMiddleware.php
```diff
@@ -0,0 +1,21 @@
...
+class SessionAuthenticationMiddleware implements HTTPMiddleware
+{
+    public function process(HTTPRequest $request, callable $delegate)
+    {
+        /** @var \SilverStripe\Security\Member $member */
+        if ($member = Security::getCurrentUser()) {
+            Manager::singleton()->setMember($member);
+        }
+        return $delegate($request);
+    }
```
robbieaverill: We can probably do away with this class since the Manager will fall back to it anyway

src/Middleware/BasicAuthenticationMiddleware.php
Firesphere: Right, nice catch on that!

src/Controller.php
Firesphere: Hmmm, valid reasoning there. I feel, however, it would be better to split out the authentication part.

Making it two separate modules, one doing the graphql and one doing the graphql security handling.

Admitted, I'm unsure at this moment how to do that though, so at it's current state, I'd say fair enough :D

src/Middleware/SessionAuthenticationMiddleware.php
Firesphere: Seems to be indeed. Another option would be to utilize it in another way, but for now it seems to be just overhead? If I fall back to the default middleware chain it seems to work.

src/Middleware/VersionedMiddleware.php
```diff
@@ -0,0 +1,20 @@
...
+{
+    public function process(HTTPRequest $request, callable $delegate)
+    {
+        $stage = $request->param('Stage');
+        if ($stage && in_array($stage, [Versioned::DRAFT, Versioned::LIVE])) {
+            Versioned::set_stage($stage);
+        }
+        return $delegate($request);
+    }
+}
```
Firesphere: I'm unsure if Versioned should be part of GraphQL?

src/Controller.php
robbieaverill: In fairness, if we decided to remove the member state from Manager and have it always be provided by `Security::getCurrentMember`, *and* the controller becomes middleware aware, then authentication could be completely removed from GraphQL and just handled by middlewares

src/Middleware/VersionedMiddleware.php
robbieaverill: This is existing code, I imagine it's fairly important for the asset admin client

Firesphere: Right. Fair enough, it was marked as green, so I assumed it was new :)

Firesphere:
I've added a few comments based on a quick look and try of the code. In general, I'd say it's definitely a step in the better direction. I would prefer to have any authentication/authorization to be handled by the framework itself, but there are valid reasons to not do that right away.

I'd much prefer Security and any auth related things to be extracted into a separate module, but I don't see a direct solution to that right now that would cover all bases (also, possibly because it's almost 11PM and I'm testing/reading this all while I'm gaming :P )

src/Controller.php
robbieaverill: I think this is feasible. Middlewares can use `Security::setCurrentMember` to provide it, GraphQL could just read from that.

src/Middleware/BasicAuthenticationMiddleware.php
robbieaverill: Ok so the implication of removing the basic auth handler from this module and using a core middleware would be that the request is authenticated outside of the GraphQL controller, which means we can't format the response nicely

robbieaverill: For the record, this middleware doesn't do this either

_config/config.yml
unclecheese: Yeah, that's fine for now. Ideally we want to be injecting on interfaces, but there is plenty of precedent right now for using the base class.

robbieaverill:
I'm gonna close this. It's so old now that it probably needs to be rebuilt from ground up, and was WIP anyway

PR #112, opened 2017-07-25
Fix list scaffolded DataObjects when not paginated

flamerohr:
Otherwise you only get a list of fields all with null values


unclecheese:
Tested, and works. Thanks!

Issue #113, opened 2017-07-26
DBField.graphql_type must be scalar


unclecheese:
In some cases, particularly `DBFile`, the value of the field is a data structure (`FileName`, `Hash`, `Variant`). This causes a fatal error when trying to resolve a field that maps to a `DBFile`, as graphql is casting the value as `String` (the fallback for all DBFields), but it's actually a hash.

Suggested fix: `graphql_type` should accept arrays as values, e.g.

```
....DBFile:
  graphql_type:
    Filename: String
    ...
```

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-graphql/pull/114

PR #114, opened 2017-08-03
Feature/complex db field types

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-graphql/issues/113

## Summary of changes

* Abstract `TypeParser` as an interface, allowing `StringTypeParser` and `ArrayTypeParser` to create types from abstract definitions, e.g. a formatted string (`Int!`) or array for complex object types

* Allow type parsers to be created on `DBField` instances, or any other `ViewableData` object, via the `TypeCreatorExtension`.

* Add custom type definition for `DBFile` which provides access to most of its properties and getters (when `silverstripe/assets` module exists)

* Allow `fixed_types` in scaffolding -- types that should be made available to the GraphQL schema irrespective of the user-defined types, effectively augmenting the list of internal types like `String`, `Int`, etc.

* Only return `DBField::getValue()` if the field's type is an internal type.


src/Scaffolding/Util/StringTypeParser.php
```diff
@@ -32,9 +35,22 @@ class TypeParser
...
+        $types = array_keys(Type::getInternalTypes());
+
+        return in_array($type, $types);
+    }
+
+    /**
      * TypeParser constructor.
      *
-     * @param string $rawArg
+     * @param string|array $rawArg
```
flamerohr: I don't believe `preg_match()` (below) would accept an array input, maybe worth adding a check for that as well?

PR #126, opened 2018-01-05
BUGFIX: Prevent defaultValue from being added to args when null

unclecheese:
Graphiql coerces `null` into a string `"null"`. It shouldn't be in the schema in the first place if it's null.


tractorcow:
@unclecheese can you use open-sausages fork otherwise we can't push changes.

flamerohr:
oh, I thought he had the "allow maintainers to make changes" checkbox checked on github

unclecheese:
I've checked the "allow edits from maintainers" box. Let me know if you need me to push this up to the other fork.

tractorcow:
Ah sorry, I wasn't sure where I could see if that was enabled.

Issue #30, opened 2016-12-20
Support for versioning and stages


sminnee:
Queries and manipulations need to work with the `Versioned` ORM add-on:

## Required operations

Queries

 * Queries by stage: draft/published ("Stage" and "Live")
 * Queries of content as at a historical date
 * Queries of content that are in a specific stage and don't exist on any other stage
 * Query content including deleted items
 * Query a specific version of a single record
 * Query all versions of a single record

Mutations

 * Write to a specific stage
 * Overwrite a specific version? (not sure if this is internal only)
 * Various workflow actions (publish, rollback, etc)

## Usage considerations

 * Powering the asset-admin and campaign-admin UIs in SS4
 * Generic content API for building rich JS apps, mobile apps, etc

## Possible implementation approaches

 * Different GraphQL end-point URLs for different stages.
 * A new GraphQL type for a stage and/or archive date that has the core content types as nested objects.
 * Extra parameters to queries and mutations for selecting stages, dates, versions
  
## Acceptance Criteria (WIP)

* All operations accept an optional parameter that defines the stage for the operation
* Stage parameter is not added if `versioned` is not installed
* No new scaffolded operations are added for reading/writing from stages
* For mutations, a reference to a version should suffice if no fields are being changed
* `Version` is exposed on all versioned data objects when `versioned` is installed
* `versioned` module provides new query for getting historical content
* Existing queries for versioned objects are backward compatible 
* Apollo can distinguish between records with duplicate IDs across different versions.

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-versioned/pull/83
- [x] https://github.com/silverstripe/silverstripe-graphql/pull/128
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/727
- [x] https://github.com/silverstripe/silverstripe-graphql/pull/133

chillu:
> Version is exposed on all versioned data objects when versioned is installed

@unclecheese Can we make this opt-out? I think there'll be public GraphQL endpoints which only want to expose a subset of properties and operations, and mask the fact that the object is versioned in the first place (or that historical versions exist).

PR #128, opened 2018-01-09
ENHANCEMENT: Versioning support for scaffolded operations

unclecheese:
Parent story https://github.com/silverstripe/silverstripe-graphql/issues/30


src/Scaffolding/Scaffolders/OperationScaffolder.php
```diff
@@ -46,20 +44,42 @@
...
+
+        return isset($operations[$class]) ? $operations[$class] : null;
+    }
+
+    /**
+     * Gets a map of operation identifiers to their classes
+     * @return array
+     */
+    public static function getOperations()
+    {
```
tractorcow: I'd like to see fewer static methods and more instance methods on a service. Can we use the OperationScaffolder instance instead?

src/Scaffolding/Util/OperationList.php
```diff
@@ -59,13 +58,17 @@ public function findByName($name)
...
+    public function findByIdentifier($identifier)
     {
-        return $this->findItemByCallback(function (OperationScaffolder $item) use ($id) {
-            return $item instanceof CRUDInterface && $id === $item->getIdentifier();
+        $scaffoldClass = OperationScaffolder::getClassFromIdentifier($identifier);
+        if (!$scaffoldClass) {
+            return false;
+        }
+        return $this->findItemByCallback(function (OperationScaffolder $item) use ($scaffoldClass) {
+            return $item instanceof $scaffoldClass;
```
tractorcow: What if another operation subclasses another? Should it be a proper class comparison, or can this allow subclasses?

composer.json
```diff
@@ -12,7 +12,7 @@
   },
   "extra": {
     "branch-alias": {
-      "1.x-dev": "1.1.x-dev"
+      "dev-master": "2.x-dev"
```
tractorcow: This should be a minor version increment not a master major version, shouldn't it?

composer.json
```diff
@@ -3,7 +3,7 @@
   "description": "GraphQL server for SilverStripe models and other data",
   "type": "silverstripe-vendormodule",
   "require": {
-    "silverstripe/framework": "^4",
+    "silverstripe/framework": "^5",
```
tractorcow: We need this in framework 4

src/Scaffolding/Scaffolders/CRUD/Create.php
```diff
@@ -89,8 +93,12 @@ protected function generateInputType()
                 foreach ($db as $dbFieldName => $dbFieldType) {
                     /** @var DBField $result */
```
tractorcow: I'd do this if you are relying on extension methods

```
/** @var DBField|TypeCreatorExtension $result */
```

src/Scaffolding/Scaffolders/CRUD/ReadOne.php
```diff
@@ -0,0 +1,61 @@
...
+            if (!singleton($this->dataObjectClass)->canView($context['currentUser'])) {
+                throw new Exception(sprintf(
+                    'Cannot view %s',
+                    $this->dataObjectClass
+                ));
+            }
+            // get as a list so extensions can influence it pre-query
+            $list = DataList::create($this->dataObjectClass)
+                ->filter('ID', $args['ID']);
+            $this->extend('updateList', $list, $args, $context, $info);
```
tractorcow: I would make `ReadOne` class take a `Read` instance as a dependency, and use this as the data source.

```php
// Inherit from source Read()
            $list = $parentRead
                ->getResults($args)
                ->filter('ID', $args['ID']);
            $this->extend('updateList', $list, $args, $context, $info);
```

and maybe move the extension point to inside getResults() so that decorations would also be respected.

src/Scaffolding/Interfaces/CRUDInterface.php
```diff
@@ -1,14 +0,0 @@
-<?php
-
-namespace SilverStripe\GraphQL\Scaffolding\Interfaces;
-
-/**
- * Defines the methods required for a class to provide a CRUD scaffold
- */
-interface CRUDInterface
```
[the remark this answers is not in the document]
unclecheese: We need to deprecate the idea of `CRUD`. The semantics of that are incorrect now that the operations are pluggable. It was named that way because when it was built, the idea was that the module would just provide those four operations out of the box, but now with versioning, that set of operations is pluggable (publish, unpublish, etc), so the concept of "CRUD" needs to be changed to something more akin to "Scaffolded Operations."

This interface was effectively used as a way of tagging a class to signal data to the scaffolder, but with these changes, the lines between a scaffolded operation and a custom one have been blurred, and that signaling is no longer necessary. They're all just operations.

Leaving this in would just be bloat. I don't know how conservative you want to be with your API changes on a module this nascent, but I'd like to see it removed.

src/Scaffolding/Scaffolders/CRUD/Create.php
unclecheese: Had no idea you could do that. Is that a union/mixin syntax? That's going to make my life so much better.

src/Scaffolding/Scaffolders/CRUD/ReadOne.php
unclecheese: That's clever, but `Read` has required constructor args. Is there a way for injector to pass in dynamic constructor args?  e.g.

```
  SilverStripe\GraphQL\Scaffolding\Scaffolders\OperationScaffolder.readOne:
    class: SilverStripe\GraphQL\Scaffolding\Scaffolders\CRUD\ReadOne
    properties:
      Reader: '%$SilverStripe\GraphQL\Scaffolding\Scaffolders\OperationScaffolder.read'
  SilverStripe\GraphQL\Scaffolding\Scaffolders\OperationScaffolder.read:
    class: SilverStripe\GraphQL\Scaffolding\Scaffolders\CRUD\Read
    constructor:
      dataObjectClass: $myOwner->operationName
```

Obviously invalid, but can you think of a way to do that?

The required constructor args for those operations make them very difficult to use as dependencies. I'd really like to make them optional and just throw in `scaffold()` if they're not set, but that's another story, and will make a ticket for it.

src/Scaffolding/Scaffolders/OperationScaffolder.php
unclecheese: Yeah, agreed. Previous version of this method was static so I left it that way, but it should be changed.

src/Scaffolding/Util/OperationList.php
unclecheese: Yup, good catch there.

src/Scaffolding/Scaffolders/SchemaScaffolder.php
```diff
@@ -24,13 +22,7 @@
  */
 class SchemaScaffolder implements ManagerMutatorInterface
 {
-    const CREATE = 'create';
```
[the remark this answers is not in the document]
unclecheese: This is explained in a previous comment. The concept of "CRUD" is effectively deprecated. The concept is more just "operations that I can activate for any dataobject type via configuration"

PR #132, opened 2018-01-18
BUG Fix cors breaking if referer header is present

### LINKED ISSUE #118: IE11 does not add CORS header when using the same domain
IE11 and Edge do not add the CORS Origin header, even when explicitly set, when the domain is the same as the domain graphql (asset admin) tries to talk to. 

This breaks the CMS when CORS is enabled. Which means the GraphQL API becomes a lot less useful to anyone who wants to serve it to other domains or standalone apps, and still support IE for it's content authors.

Explicitly setting the Origin header in the asset admin, does not resolve the problem, as IE still does not actually add it.

Background is here:
https://connect.microsoft.com/IE/feedback/details/781303/origin-header-is-not-added-to-cors-requests-to-same-domain-but-different-port
Although the issue applies to the same port as well as a different port

# PRs:

 - [ ] https://github.com/silverstripe/silverstripe-graphql/pull/132

tractorcow:
BUG Prevent un-extendable config by shifting defaults into PHP
BUG Remove dependency on Doctrine module breaking with --prefer-dist
BUG Fix tests not checking cors port
ENHANCEMENT Clean up Controller::index() method and make lovely
ENHANCEMENT Optimise all imports
Fixes #118

### DIFF
--- a/src/Controller.php
+++ b/src/Controller.php
@@ -2,15 +2,16 @@
 
 namespace SilverStripe\GraphQL;
 
+use Exception;
 use SilverStripe\Control\Controller as BaseController;
+use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Core\Config\Config;
-use SilverStripe\Control\Director;
 use SilverStripe\GraphQL\Auth\Handler;
-use SilverStripe\Versioned\Versioned;
-use Exception;
+use SilverStripe\Security\Member;
 use SilverStripe\Security\Permission;
+use SilverStripe\Versioned\Versioned;
 
 /**
  * Top level controller for handling graphql requests.
@@ -19,6 +20,20 @@ use SilverStripe\Security\Permission;
  */
 class Controller extends BaseController
 {
+    /**
+     * Cors default config
+     *
+     * @config
+     * @var array
+     */
+    private static $cors = [
+        'Enabled' => false, // Off by default
+        'Allow-Origin' => [], // List of all allowed origins; Deny by default
+        'Allow-Headers' => 'Authorization, Content-Type',
+        'Allow-Methods' => 'GET, POST, OPTIONS',
+        'Max-Age' => 86400, // 86,400 seconds = 1 day.
+    ];
+
     /**
      * @var Manager
      */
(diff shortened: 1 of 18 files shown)


src/Controller.php
```diff
@@ -155,45 +138,157 @@ public function getAuthHandler()
...
+    protected function getRequestOrigin(HTTPRequest $request)
+    {
+        // Prefer Origin header
+        $origin = $request->getHeader('Origin');
+        if ($origin) {
+            return $origin;
+        }
+
+        // Check referer
+        $referer = $request->getHeader('Referer');
```
Firesphere: I would say this should be a IE/Edge only fallback, not a "fallback because whenever"

tractorcow: Why would it be? Firefox also has this issue in some conditions, and none of the other solutions I've seen around the web target IE.

Referer isn't less trusted than Origin.

tractorcow: Also, user-agent targeting is bad practice. Read up https://webaim.org/blog/user-agent-string-history/ :D

Firesphere: I'm well aware user-agent targeting is bad :)

I wasn't aware that Firefox showed the same behaviour sometimes. It makes sense to fully fall back in that case. it just feels a bit like a security issue, because it opens up the whole reason the CORS header was added.

tractorcow: Yes, older firefox wouldn't send origin on HTTP POST. My feeling is that if we errored on missing Origin when we had a valid Referer would be a bug on our part.

Firesphere: Fair enough, as the change is a "hack" that breaks a part of security anyway. It just makes me uncomfortable ;)

Firesphere: I stand with that it is a potential security issue, but also understand it's the only feasible solution at this moment.

tractorcow: > Fair enough, as the change is a "hack" that breaks a part of security anyway. It just makes me uncomfortable ;)

What part is broken?

Firesphere: Basically, the way CORS is supposed to work. I'm not saying the code itself is broken :)

flamerohr: The origin header is just harder to spoof on browsers than the referrer header, while neither are enforced when using something like `curl`
https://stackoverflow.com/q/21058183/695842

Two of the top answers resonate the same thing: don't rely on CORS for security

> Remember: CORS is not security. Do not rely on CORS to secure your site. If you are serving protected data, use cookies or OAuth tokens or something other than the Origin header to secure that data. The Access-Control-Allow-Origin header in CORS only dictates which origins should be allowed to make cross-origin requests. Don't rely on it for anything more.

src/Controller.php
```diff
@@ -155,45 +138,157 @@ public function getAuthHandler()
...
+     * Response for HTTP OPTIONS request
+     *
+     * @param HTTPRequest $request
+     * @return HTTPResponse
+     */
+    protected function handleOptions(HTTPRequest $request)
+    {
+        $response = HTTPResponse::create();
+        $corsConfig = Config::inst()->get(self::class, 'cors');
+        if ($corsConfig['Enabled']) {
```
flamerohr: this changes the workflow of the CORS headers...

from the original ticket which raised this concern https://github.com/silverstripe/silverstripe-graphql/issues/66#issuecomment-276842929
> Also adds a check for the pre-flight OPTIONS request, and if CORS is enabled, adds the appropriate headers to the response object - **both for the OPTIONS response and for the actual data response**.

src/Scaffolding/Scaffolders/SchemaScaffolder.php
```diff
@@ -34,6 +32,14 @@ class SchemaScaffolder implements ManagerMutatorInterface
 
     const ALL = '*';
 
+    /**
+     * List of fixed types
+     *
+     * @config
+     * @var array
+     */
+    private static $fixed_types = [];
```
flamerohr: where'd this come from?

tractorcow: Oh right, it was an error where it was searching for an undefined config. Probably not related to this PR and can be removed.

src/Controller.php
tractorcow: It's the same workflow; In which situation is it different?

tractorcow:
I've shifted some of the namespace cleanup into a separate commit.

src/Scaffolding/Scaffolders/SchemaScaffolder.php
tractorcow: I've reverted this part.

src/Controller.php
flamerohr: ok, I needed to double check this :)

I overlooked the second `=== 'OPTIONS'` that was removed above

tractorcow: Yeah the original code was hard to follow, hence the refactor. :)

src/Scaffolding/Scaffolders/SchemaScaffolder.php
flamerohr: oh ok, I had a quick search for it just now, seems harmless to be in here and could've been left in :)

PR #135, opened 2018-01-30
Remove ID from Update, add to Input

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-graphql/issues/105


flamerohr:
We've agreed that this goes in master (so not available until SS5)

With the new extension points and other graphql work coming up, if you want this earlier, you could utilise the extension points to mutate the args and get what you need.

PR #138, opened 2018-02-08
BUG Fix broken createArgs() for scaffolded reads

### LINKED ISSUE #137: Adding args using scaffolding on code/config does not work.
https://github.com/silverstripe/silverstripe-graphql/tree/1.1#adding-arguments
I don't think this is working. It should be stated clearly in the documentation so that future users are not confused. 

IF it's in the roadmap, then perhaps note them somewhere as well. 

Use TypeCreator and QueryCreator if you want to use query arguments. 

# PRs:
- [x] https://github.com/silverstripe/silverstripe-graphql/pull/138
- [x] https://github.com/silverstripe/silverstripe-versioned/pull/111

tractorcow:
Fixes #137

After this is merged, you can merge https://github.com/silverstripe/silverstripe-versioned/pull/111

### DIFF
--- a/src/Scaffolding/Scaffolders/CRUD/Create.php
+++ b/src/Scaffolding/Scaffolders/CRUD/Create.php
@@ -2,22 +2,23 @@
 
 namespace SilverStripe\GraphQL\Scaffolding\Scaffolders\CRUD;
 
-use SilverStripe\Core\Extensible;
+use Exception;
+use GraphQL\Type\Definition\InputObjectType;
+use GraphQL\Type\Definition\Type;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\GraphQL\Manager;
+use SilverStripe\GraphQL\Scaffolding\Extensions\TypeCreatorExtension;
+use SilverStripe\GraphQL\Scaffolding\Interfaces\ResolverInterface;
 use SilverStripe\GraphQL\Scaffolding\Scaffolders\MutationScaffolder;
 use SilverStripe\GraphQL\Scaffolding\Traits\DataObjectTypeTrait;
-use GraphQL\Type\Definition\InputObjectType;
-use SilverStripe\Core\Injector\Injector;
-use Exception;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataObjectSchema;
 use SilverStripe\ORM\FieldType\DBField;
-use GraphQL\Type\Definition\Type;
 
 /**
  * A generic "create" operation for a DataObject.
  */
-class Create extends MutationScaffolder
+class Create extends MutationScaffolder implements ResolverInterface
 {
     use DataObjectTypeTrait;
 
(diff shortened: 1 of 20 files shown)


tractorcow:
Nope, all the other operations are broken. Don't merge!

src/Scaffolding/Scaffolders/CRUD/Create.php
```diff
@@ -2,22 +2,23 @@
...
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataObjectSchema;
 use SilverStripe\ORM\FieldType\DBField;
-use GraphQL\Type\Definition\Type;
 
 /**
  * A generic "create" operation for a DataObject.
  */
-class Create extends MutationScaffolder
+class Create extends MutationScaffolder implements ResolverInterface
```
unclecheese: I really like the idea of replacing the explicit assignment of a closure with something more pluggable, but I'd much prefer that these scaffolded operations take a `ResolverInterface` as a dependency.

```
private static $dependencies = [ 
  'Resolver' => '%$SilverStripe\GraphQL\Scaffolding\Interfaces\ResolverInterface.read'
];
```

In addition to allowing the operation to be customisable without  subclassing, it seems more consistent with the change we're trying to make, which is to make customising operations much better. This would put us in a position to make the API more composable, e.g.

```
  SilverStripe\GraphQL\Scaffolding\Interfaces\ResolverInterface.blogPostReadResolver
    class: MyProject\Resolvers\MyBlogResolver
  SilverStripe\GraphQL\Scaffolding\Scaffolders\CRUD\Read.blogPost
    properties:
      Resolver: %$SilverStripe\GraphQL\Scaffolding\Interfaces\ResolverInterface.blogPostReadResolver

Scaffolding:
  types:
     BlogPost:
       operations:
          read:
             service: SilverStripe\GraphQL\Scaffolding\Scaffolders\CRUD\Read.blogPost     
```

... or something to that effect. Just a thought.

src/Scaffolding/Scaffolders/CRUD/Create.php
```diff
@@ -63,20 +47,13 @@ public function addToManager(Manager $manager)
         parent::addToManager($manager);
     }
 
-    /**
-     * @param Manager $manager
-     * @return array
-     */
-    protected function createArgs(Manager $manager)
+    protected function createDefaultArgs(Manager $manager)
```
unclecheese: I wonder if there's a better name for this. For me, it feels a bit in conflict with `setArgDefaults()`, which does something entirely different, but I could see someone thinking it influences this method.

src/Scaffolding/Scaffolders/CRUD/Read.php
```diff
@@ -25,35 +28,7 @@ class Read extends ListQueryScaffolder
...
-     * @param Manager $manager
-     * @return array
-     */
-    protected function createArgs(Manager $manager)
-    {
-        $args = [];
-        $this->extend('updateArgs', $args, $manager);
-
-        return $args;
+        parent::__construct($this->createOperationName(), $this->typeName(), $this);
```
unclecheese: Passing `$this` to the parent constructor feels awfully awkward to me. I'd much rather see `createResolverFunction()` refactored:

```
if ($this instanceof ResolverInterface) {
  // .. just call class member
}
```

.. but as stated above, I think all of this is better served by a dependency.

src/Scaffolding/Scaffolders/CRUD/Update.php
```diff
@@ -80,19 +52,16 @@ public function addToManager(Manager $manager)
      * @param Manager $manager
      * @return array
      */
-    protected function createArgs(Manager $manager)
+    protected function createDefaultArgs(Manager $manager)
     {
-        $args = [
+        return [
```
unclecheese: Just be aware that all of this will throw a merge conflict with the input type refactoring we put into master. https://github.com/silverstripe/silverstripe-graphql/issues/105

src/Scaffolding/Scaffolders/CRUD/Create.php
tractorcow: Sure, but in this case the mutation is it's own resolver. Maybe another story?

src/Scaffolding/Scaffolders/CRUD/Create.php
tractorcow: We can't set it because it depends on the $manager, it can only be late-resolved.

src/Scaffolding/Scaffolders/CRUD/Update.php
tractorcow: I think it can be merged but it'll require a bit of manual work.

PR #151, opened 2018-04-19
ENHANCEMENT Consolidate type / operation name generation

tractorcow:
Cleans up docs, merges duplicate resolver interface (ResolverInterface / OperationResolver). I've made a few variables private to ensure that they cannot be accessed bypassing business logic in failing over to default name scaffolders.


tractorcow:
Darn, we can't merge this without the versioned PR at the same time.

src/Scaffolding/Scaffolders/CRUD/Delete.php
```diff
@@ -29,12 +29,16 @@ public function __construct($dataObjectClass)
         parent::__construct(null, null, $this, $dataObjectClass);
     }
 
+    /**
+     * @return string
+     */
```
tractorcow: I've been put in my place huh. :D

unclecheese: The interface's declaration is sufficient, but you've been beating me over the head with it long enough that things like this are itches that need to be scratched.

tractorcow: See, the power was within you all along! :D haha.

PR #152, opened 2018-04-22
Fix custom nested query class check

michalkleiner:
Use params in the subclass check in the right order.

### DIFF
--- a/src/Scaffolding/Scaffolders/DataObjectScaffolder.php
+++ b/src/Scaffolding/Scaffolders/DataObjectScaffolder.php
@@ -495,9 +495,23 @@ class DataObjectScaffolder implements ManagerMutatorInterface, ScaffolderInterfa
             foreach ($config['nestedQueries'] as $relationName => $settings) {
                 if ($settings === false) {
                     continue;
+                } elseif (is_string($settings)) {
+                    if (is_subclass_of($settings, QueryScaffolder::class)) {
+                        $queryScaffolder = new $settings($relationName);
+                        $this->nestedQuery($relationName, $queryScaffolder);
+                    } else {
+                        throw new InvalidArgumentException(sprintf(
+                            'Tried to specify %s as a custom query scaffolder for %s on %s, but it is not a subclass of %s.',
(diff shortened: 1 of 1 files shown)


flamerohr:
Thanks @michalkleiner!

do you want to shift this to the 1 branch?
going to master will mean this won't be fixed until version 2

michalkleiner:
Against 1.0 the change is a bit bigger as it backports some code added further along the way, but it should be fine.

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
```diff
@@ -495,9 +495,23 @@ public function applyConfig(array $config)
...
+                            'Tried to specify %s as a custom query scaffolder for %s on %s, but it is not a subclass of %s.',
+                            $settings,
+                            $relationName,
+                            $this->dataObjectClass,
+                            QueryScaffolder::class
+                        ));
+                    }
+                } else {
+                    $this->nestedQuery($relationName)
+                        ->applyConfig((array)$settings);
```
flamerohr: I'm a little adverse to add that thrown Exception, as it changes expectations for a patch fix... might be a good idea to remove that for the 1.0 branch.

Can ignore my slack message about this line being skipped :)

unclecheese:
We'll need test coverage for this. `nestedQuery($scaffolderInstance)` isn't tested anywhere from what I can see, so this would be a good time to pay off that technical debt. :)

Likewise, with this PR, the `applyConfig` test should be updated to include a FQCN scaffolder string. https://github.com/silverstripe/silverstripe-graphql/blob/2/tests/Scaffolding/Scaffolders/DataObjectScaffolderTest.php#L211

michalkleiner:
@unclecheese what if we change the PR to target the version where the bug I was fixing has been introduced? Would the unit test be there already? And then when that gets merged to version 1, all will be nice and fluffy already.

maxime-rainville:
We're focusing on the 4 branch right now. We're not looking at doing another 1.x release.

PR #154, opened 2018-05-02
ENHANCEMENT: New static caching of schema types, as well as dynamic endpoint

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-admin/issues/423

### DIFF
--- a/src/Extensions/IntrospectionProvider.php
+++ b/src/Extensions/IntrospectionProvider.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace SilverStripe\GraphQL\Extensions;
+
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Core\Extension;
+use SilverStripe\GraphQL\Scaffolding\StaticSchema;
+
+/**
+ * Class FragmentProvider
+ * @package silversripe/graphql
+ */
+class IntrospectionProvider extends Extension
+{
+    private static $allowed_actions = [
+        'types'
+    ];
+
+    /**
+     * @param HTTPRequest $request
+     * @return HTTPResponse
+     */
+    public function types(HTTPRequest $request)
+    {
+        $manager = $this->owner->getManager();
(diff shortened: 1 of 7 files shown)


src/Extensions/IntrospectionProvider.php
```diff
@@ -0,0 +1,32 @@
...
+namespace SilverStripe\GraphQL\Extensions;
+
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Core\Extension;
+use SilverStripe\GraphQL\Scaffolding\StaticSchema;
+
+/**
+ * Class FragmentProvider
+ * @package silversripe/graphql
```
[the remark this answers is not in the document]
maxime-rainville: `Class FragmentProvider` is not very descriptive of the purpose of this class.

src/Scaffolding/StaticSchema.php
```diff
@@ -49,10 +51,7 @@ class StaticSchema
      */
     public static function inst()
     {
-        if (!self::$instance) {
-            self::$instance = new static();
-        }
-        return self::$instance;
+        return Injector::inst()->get(static::class);
```
maxime-rainville: Are third parties meant to be able to provide their own implementation of StaticSchema and load it via the Injector?

If so, it probably should have an interface.

unclecheese: Ideally, yes, but you could also just plug in a subclass, which is why it's here (for mocking in tests). This Injector call is just longform for `::create()`, which we use all over the place. Not everything that uses `Injectable` has an interface on it.

This is an oddball class that probably doesn't have much of a lifespan, TBH, so I'm going to park this one for now, but the feedback is definitely taken and understood. :)

unclecheese: Actually, scratch that. I'll just add a `setIntance` method to `StaticSchema` and get rid of the Injector call. That's an easy work around.

Issue #164, opened 2018-06-26
RFC Serverside graphql query caching


tractorcow:
Related to https://github.com/silverstripe/silverstripe-graphql/issues/130 (schema caching)

# Use case

When generating responses for DB intensive queries it can be necessary to cache to prevent repeated overhead cost. The challenge is that the query controller doesn't immediately know how to invalidate / validate caches.

# Options

There are two main approaches that could be taken, each with a similar difficulty in their approach:

 - Caches automatically invalidate based on the modified date of the records that back them. The challenge is that graphql queries don't always naturally map strictly to dataobject types, although specific parts of a query response can.
 - Caches are explicitly cached on write of a specific model. This requires user logic (or otherwise a reasonably smart onAfterWrite extension) to trigger invalidation of the query. The inverse of the above is the issue here; Given a model, there's no explicit knowledge of which queries should be invalidated.

# Suggestion

Given that queries are more complex than the underlying data model, my suggestion is to look at the former, where we develop a stronger knowledge of what models back a query. This could be done one of two ways:
 - Static analysis of the query, where we traverse the query on each request to extract the metadata on demand. cc @unclecheese to expand on this idea. :)
- On cache-miss, whenever a record type is queried we register this on the graphql controller / type, so that when the cache response is saved to the cache, we additionally have a list of classes (and perhaps other metadata) that we can include in the cached value. Since this is only done on cache-miss, a cache-hit can extract this from the cache itself rather than performing static analysis, and can do post-hit validation to see if any of the models have been modified since the cache was created.

Pseudocode:

```php
$key = md5($query) . '-' . var_export($args, true);
$cached = $cache->get($key);
if ($cached) {
   foreach ($cached['classes'] as $class) {
      if (DataObject::get($class)->max('LastModified') > $cached['date']) {
        $cached = null; // class modified, fail validation of cache
        break;
      }
   }
}
if (!$cached) {
   $response = $manager->respond($query, $args);
   $classesUsed = $manager->getClassesUsed();
   $cache->save($key, ['response' => $response, 'classes' => $classesUsed, 'date' => DBDateTime::now()->getValue()]);
}
return $response;
```

Whenever a type would perform a query, it would need to register any class it used. Otherwise, we could "hook" into dataquery with an extension and spy on classes queried during that time.

tractorcow:
Pseudocode with spy enabled.

```php
$key = md5($query) . '-' . var_export($args, true);
$cached = $cache->get($key);
[...]

PR #166, opened 2018-06-28
ENHANCEMENT Proof of concept of cached graphql queries

tractorcow:
See #164

Note: This is actually a new minor version (new features) so it should go into a `1` branch (which I haven't created yet), so don't actually merge this please. :)

Tests should come if we think this approach looks good.

I won't post the specifics, but I can confirm that specific queries correctly "spy" on the correct classes queried by the internal graphql.

Another todo is respect versioned deletions (won't affect LastEdited, but will appear as new _Versioned rows). This todo is marked in the code.

@unclecheese / @chillu  / @stevie-mayhew for review.

I've seen performance increases of up to 6x for cached results. ;)

### DIFF
--- a/src/Middleware/QueryMiddleware.php
+++ b/src/Middleware/QueryMiddleware.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace SilverStripe\GraphQL\Middleware;
+
+use GraphQL\Executor\ExecutionResult;
+
+/**
+ * Represents middleware for evaluating a graphql query
+ */
+interface QueryMiddleware
+{
+
+    /**
+     * @param string $query
+     * @param array $params
+     * @param callable $next
+     * @return ExecutionResult|array Result either as an ExecutionResult object or raw array
+     */
+    public function process($query, $params, callable $next);
+}
--- a/src/Middleware/QueryMiddleware.php
+++ b/src/Middleware/QueryMiddleware.php
@@ -3,18 +3,20 @@
 namespace SilverStripe\GraphQL\Middleware;
 
 use GraphQL\Executor\ExecutionResult;
+use GraphQL\Schema;
 
 /**
  * Represents middleware for evaluating a graphql query
  */
 interface QueryMiddleware
 {
-
     /**
+     * @param Schema $schema
      * @param string $query
+     * @param array $context
      * @param array $params
      * @param callable $next
      * @return ExecutionResult|array Result either as an ExecutionResult object or raw array
      */
-    public function process($query, $params, callable $next);
+    public function process(Schema $schema, $query, $context, $params, callable $next);
 }
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -100,22 +100,24 @@ class Manager
     /**
      * Call middleware to evaluate a graphql query
      *
+     * @param Schema $schema
      * @param string $query Query to invoke
+     * @param array $context
      * @param array $params Variables passed to this query
      * @param callable $last The callback to call after all middlewares
      * @return ExecutionResult|array
      */
-    protected function callMiddleware($query, $params, callable $last)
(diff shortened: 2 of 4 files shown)


tractorcow:
Oh well, linting issues are easy to fix :D

tractorcow:
Just in case a nested cache needed to record classes. It just means that the extension works if you nest calls. Otherwise I'd need to throw an error if nested.

I could add it to just the top level, and make sure when we pop the classes, we merge all levels together.

chillu:
Given your recorder is enabled in the main GraphQL `queryAndReturnResult()`, the only way I can see this nesting happening is through unit or functional tests, right?

If we allow caching per resolver (rather than per query), we'd have more potential for nesting (resolvers can call child resolvers).

By way of example, here's a simple nested query where the nested fields might be expensive to calculate (`RelatedPages`), and hence might be cached at a different granularity. 

```
query File() {
  ID
  RelatedPages {
    ID
  }
}
```

Here's how I *think* levels are working at the moment (haven't run the code):

* GraphQL controller calls middlewares, calls resolvers through `recordClasses()` closure
* `FileResolver` resolves within original closure
* Cache levels are set to `[['File']]`
* `RelatedPages` resolves within original closure
* Cache levels are set to `[['Page'], ['File', 'Page']]`
* `recordClasses()` returns `[['File', 'Page'], '<my-result>']`

I'm proposing to change this to:

* GraphQL controller calls resolvers directly
* `FileResolver` creates cache entry based on `File`, without `RelatedPages` (through a "pointer" in the cache entry?)
* `RelatedPages` creates cache entry based on `Page`

Am I completely off my rocker with this? You could argue that we could introduce this level of caching later, but I'd rather only support *one* way of doing it that covers both use cases (full response and resolver-specific caching).

tractorcow:
I'm having trouble visualising your solution @chillu. Maybe if you can explain it to @unclecheese and have him write up his interpretation here? I think he's the best person to critique anything that interacts with the actual query resolution process.

tractorcow:
I have made these changes:
 - Stripped out the actual cache middleware class.
 - Added documentation to the result serialisation to explain why we need to handle both array|ExecutionResult object
 - Added 1.x branch upstream

Hopefully this is enough to allow users to add their own cache middleware for graphql projects as necessary.

src/Middleware/QueryMiddleware.php
```diff
@@ -0,0 +1,20 @@
...
+interface QueryMiddleware
+{
+
+    /**
+     * @param string $query
+     * @param array $params
+     * @param callable $next
+     * @return ExecutionResult|array Result either as an ExecutionResult object or raw array
+     */
+    public function process($query, $params, callable $next);
```
chillu: Should we pass the context to middleware? You might want to set metadata that's relevant for resolvers further down in the execution, right?

tractorcow: Yes that's good. I'll add it into the args.

I will add a unit test as requested also.

tractorcow: I'm going to add in $schema as well in case we build in multi-schema support.

chillu: We've already built that, so great catch :) I guess it ties this API to the Webonyx module API (schema API signature), rather than just a bunch of arrays, but we're already tied pretty heavily to it in other areas, so I'm not concerned.

tractorcow:
Pushed up @chillu

chillu:
Awesome, thanks Damian! I'll write a quick README addition for this as well.

tractorcow:
Code necessary to add in caching itself shivved into https://gist.github.com/tractorcow/fe6571d2d00340a311ca31a677a05a29 for anyone interested.

chillu:
@tractorcow I've created a README pull request over at https://github.com/silverstripe/silverstripe-graphql/pull/178

Issue #153, opened 2018-04-23
OperationScaffolder::addArgs should allow types instead of declarative string inputs


robbieaverill:
Quote @unclecheese "addArgs should allow types instead of just declarative string inputs"

For the current history viewer work I need to be able to apply the `VersionedInputType` argument to the nested Versions subquery that silverstripe/versioned attached to versionable DataObject read operations.

Example of modified HistoryViewer query for the CMS: https://gist.github.com/robbieaverill/261d09b0b34f04e0684d16a5cc213108/revisions

## Pull Request
- [ ] https://github.com/silverstripe/silverstripe-graphql/pull/168

unclecheese:
`->addArgs('MyArg', $myInputType)` should work

We'll have to consider how that affects the Manager, though.  Would probably have to scan the arg list on addToManager and register any input types that have been added.

maxime-rainville:
@robbieaverill Is this still a blocker for you? Should we put an higher priority on it?

robbieaverill:
@maxime-rainville its tied to https://github.com/silverstripe/silverstripe-versioned/issues/147 which combined will be a blocker for anyone on a large project using the history viewer/versioned-admin module

maxime-rainville:
Ok. Throw some angry passive-aggressive comments at Ingo if you feel it needs be an higher priority.

robbieaverill:
As a a similar follow up to this, I can't apply the VersionedInputType to a versioned DataObject. I can apply it to the Versions field underneath it, but for example I can't do this:

```graphql
query getBlocksForPage($id:ID!) {
  readOnePage(ID: $id, Versioning: {
    Mode: LATEST
  }){
    ID
    ElementalArea {
      Elements {
        pageInfo {
          totalCount
        }
        edges {
          node {
            ID
            Title
            Summary
            Publisher # Cannot query field "Publisher"...
          }
        }
      }
    }
  }
}
```

`Published` is only available on the nested `Versions` list underneath that. I'm super hesitant to use that subquery list because it's immediately executed so would kill anyone's website when used in the context of what I'm working on. Related: https://github.com/silverstripe/silverstripe-versioned/issues/147

PR #168, opened 2018-07-04
ENHANCEMENT: Allow non-internal input types passed as args

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-graphql/issues/153

### DIFF
--- a/src/Scaffolding/Scaffolders/ArgumentScaffolder.php
+++ b/src/Scaffolding/Scaffolders/ArgumentScaffolder.php
@@ -7,6 +7,7 @@ use SilverStripe\GraphQL\Manager;
 use SilverStripe\GraphQL\Scaffolding\Interfaces\ConfigurationApplier;
 use SilverStripe\GraphQL\Scaffolding\Interfaces\TypeParserInterface;
 use SilverStripe\Core\Injector\Injector;
+use InvalidArgumentException;
 
 class ArgumentScaffolder implements ConfigurationApplier
 {
@@ -49,7 +50,7 @@ class ArgumentScaffolder implements ConfigurationApplier
             [$typeStr]
         );
         $this->defaultValue = $parser->getDefaultValue();
-        $this->type = $parser->getType();
+        $this->type = $parser->getType(false);
         $this->required = $parser->isRequired();
     }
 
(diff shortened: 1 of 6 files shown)


src/Scaffolding/Scaffolders/OperationScaffolder.php
```diff
@@ -222,6 +225,45 @@ public function setArgDefault($argName, $default)
...
 
+    /**
+     * Sets operation arguments as required or not
+     * [
+     *  'ID' => true
+     * ]
+     * @param $argData
+     * @return $this
+     */
+    public function setArgsRequired($argData)
```
maxime-rainville: Missing expected type for `$argData`.

src/Scaffolding/Util/StringTypeParser.php
```diff
@@ -118,14 +114,16 @@ public function getDefaultValue()
...
     }
 
     /**
-     * @return Type
+     * @param boolean $nullable If true, allow the type to be null. Otherwise,
+     *  return the typename, which may be arbitrary.
+     * @return Type|string
      */
-    public function getType()
+    public function getType($nullable = true)
```
maxime-rainville: Shouldn't `TypeParserInterface::getType()` be updated to match this signature?

maxime-rainville: And shouldn't `ArrayTypeParser::getType()` be updated as well?

unclecheese: Not necessarily. The interface contract says that you should be able to call the method without any parameters. Because we're supplying a default value for `$nullable`, we're not in breech of that contract.

src/Scaffolding/Scaffolders/ArgumentScaffolder.php
```diff
@@ -137,13 +139,28 @@ public function applyConfig(array $config)
 
     /**
      * Creates an array suitable for a map of args in a field
+     * @param Manager $manager
      * @return array
      */
-    public function toArray()
+    public function toArray(Manager $manager = null)
```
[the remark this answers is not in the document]
unclecheese: In testing, yes, it's really useful to be able to pass in different instances. In practice, the idea of Manager being a de facto singleton (even in the proposed multi-schema design) shouldn't be a concern of this class IMO. That would be too much coupling.

I don't see the case for a constructor argument, since the class doesn't require a Manger to function. There's plenty of API surface that's still usable without it, and even the `toArray()` method itself doesn't require it.

maxime-rainville:
I'm satisfied with the provided explanation.

PR #170, opened 2018-07-30
FIX DataObjectScaffolder is now injectable, and typos in readme are fixed

robbieaverill:
The context key is `currentUser`, so updated the examples to reflect this. Also the indentation level for `resolver` is important, and I've made the DataObjectScaffolder instantiation go through injector.

### DIFF
--- a/README.md
+++ b/README.md
@@ -890,7 +890,7 @@ $scaffolder
                 'Title' => 'String'
             ])
             ->setResolver(function($object, array $args, $context, ResolveInfo $info) {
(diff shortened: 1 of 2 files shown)


dhensby:
Ideally these two changes would be separate commits, or even PRs

robbieaverill:
Yeah you're right, sorry =(

dhensby:
Mind splitting the commit and then we can merge :)

robbieaverill:
Done, cheers

dhensby:
thank you

PR #171, opened 2018-07-31
FIX Don't rely on return value of GraphQL scaffolding providers

ScopeyNZ:
Currently the interface doesn't mention a return value on the method `provideGraphQLScaffolding`. This is an issue when the method is actually called as the manager relies on the return value.

This PR removes that reliance (as it probably should be from the get go)

I've targeted 2.0 with this as a patch as the documented API doesn't mention a return value on the interface.

### DIFF
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -148,7 +148,7 @@ class Manager
                         ScaffoldingProvider::class
                     ));
                 }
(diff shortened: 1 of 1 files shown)


unclecheese:
Yeah, this has always bothered me. Thanks.

PR #175, opened 2018-08-06
Feature #167: support persisted query

zzdjk6:

PR #179, opened 2018-08-27
NEW: Persist query support

unclecheese:
Originally here: https://github.com/silverstripe/silverstripe-graphql/pull/175. Have migrated to expedite last few changes.


unclecheese:
This was originally based on `2`. Have rebased against `3`, which is why a lot of other changes are coming though. `2` was due for a merge up, anyway.

chillu:
@zzdjk6 Are you still planning to work on Aaron's feedback?

zzdjk6:
> @zzdjk6 Are you still planning to work on Aaron's feedback?

Well, the original feedbacks and my modifications are put in #175. This PR could be regarded as the final version of our work, and I am very happy with the end result from his hand: it is based on my previous commits, and it becomes more flexible and easier to test than my version in the closed PR (#175).

One thing I have noticed in this PR is that the CI actually failed in his last commits. I may need to take some time to check the reason and fix it.

zzdjk6:
@unclecheese Please take a look at my latest changes based on your repo here (as I have no direct write permission to your repo): https://github.com/open-sausages/silverstripe-graphql/pull/1

It seems that when using the Mock in the `Manager` class, `Injector::inst()->create()` doesn't work correctly (and cause the failure of the unit test), so I replaced it with `Injector::inst()->get()`. Plus, using singleton provider class injection seems more reasonable than create a new provider every time :)

zzdjk6:
Also, I think it might be better to use `$this->config()->schemaToJSON` instead of `$this->schemaToJSON` in `getSchemaMapping()` to utilize the configurable properties on the fly.

unclecheese:
Thanks, @zzdjk6 Can you post your comments inline on the diff so we can have some context?

zzdjk6:
@unclecheese Sure. Below are the screenshots of the major changes to fix the unit test failure. (I also put some comments there: https://github.com/open-sausages/silverstripe-graphql/pull/1/files#diff-e07e4dbec2b7af0db87b7c9a92ddee86R542) 

image

image

As for the **Configuration Properties** suggestion, I think we can forget about that as we are injecting the whole class with its instance properties now.

PR #188, opened 2018-10-06
API Scaffolders can now be more expressive when registering fields

### LINKED ISSUE #162: Type hinting support
Currently the scaffolder relies on introspection for computed "fields", (i.e. custom getters) that are exposed on graphql dataobject types. This introspection is quite crude. The scaffolder invokes the method on a singleton and checks the return type. In many instances, with a singleton lacking so much state, these getters return null, or some other primitive value that isn't really what the schema should be using, for instance:

```php
public function getPrimaryCategory()
{
  return $this->Categories()->filter('Featured', true)->first();
}
```

A singleton will return null in this case, but clearly the intention is that in the schema this field is represented as a `Category`.

## Possible solutions

* Expand the `fields` setting to deeper nesting (currently Name: Description map)

```yaml
fields:
  Title:
    description: The title of the product
    type: String
  PrimaryCategory:
    type: MyApp\Category
  - NonSpecificField #Relies on introspection
```

* Or... Stop treating getters as implicit fields, and create a new setting for them

```yaml
fields:
  - Title
  - Content
computed:
  PrimaryCategory: MyApp\Category
```

ScopeyNZ:
This adds the ability to register a new `FieldDefinition` when adding fields with a scaffolder. This means you can explicitly specify the type AND the resolver for the field rather than relying on it's resolution through the field type of the field on the dataobject.

Part of the reasoning for this PR is it allows scaffolders to define types (ie. boolean, text, int) rather than using `->obj` which falls back to "text" for everything that has no casting (or `$db`) defined. Eg:

```
$scaffolder->addField('MethodAccess', Type::bool(), function($obj) { 
     return $obj->myMethod();
}
```

This PR also tidies up how the fields are retained within the scaffolder, rather than being a non-associative array and then making sure the array is distinct on the field name, I updated it to just be associative by field name. Slight performance gain I suspect.

Fixes #162

### DIFF
--- a/src/FieldDefinition.php
+++ b/src/FieldDefinition.php
@@ -1,0 +1,69 @@
+<?php
+namespace SilverStripe\GraphQL;
+
+use GraphQL\Type\Definition\Type;
+
+class FieldDefinition
+{
+    /**
+     * Self documenting description for this field
+     *
+     * @var string
+     */
+    protected $description;
+
+    /**
+     * The GraphQL type (or string the type can be resolved from) of this field
+     *
+     * @var Type|string
+     */
+    protected $type;
+
+    /**
+     * A callable that can be used to resolve the value of this field. Takes the following parameters:
+     *
+     *  - $object mixed - The object to resolve the field from - eg. the DataObject
+     *  - $args array -
+     *  - $context array - Context for this resolve - includes details like `currentUser` (logged in `Member`)
+     *  - $info GraphQL\Type\Definition\ResolveInfo - Additional info to be used for resolving this
+     *
+     * @var callable
+     */
+    protected $resolver;
+
+    /**
+     * @param string $description
+     * @param Type|string $type
+     * @param callable $resolver
+     */
+    public function __construct($description, $type, callable $resolver)
+    {
+        $this->description = $description;
+        $this->type = $type;
+        $this->resolver = $resolver;
+    }
+
+    /**
+     * @return string
+     */
+    public function getDescription()
+    {
+        return $this->description;
+    }
+
+    /**
+     * @return Type|string
+     */
+    public function getType()
+    {
+        return $this->type;
+    }
+
+    /**
+     * @return callable
+     */
+    public function getResolver()
+    {
+        return $this->resolver;
+    }
+}
--- a/src/FieldDefinition.php
+++ b/src/FieldDefinition.php
@@ -2,6 +2,7 @@
 namespace SilverStripe\GraphQL;
 
 use GraphQL\Type\Definition\Type;
+use SilverStripe\GraphQL\Scaffolding\Extensions\TypeCreatorExtension;
 
 class FieldDefinition
 {
@@ -15,7 +16,7 @@ class FieldDefinition
     /**
      * The GraphQL type (or string the type can be resolved from) of this field
      *
-     * @var Type|string
+     * @var Type|TypeCreatorExtension|string
      */
     protected $type;
 
@@ -33,7 +34,7 @@ class FieldDefinition
 
     /**
      * @param string $description
-     * @param Type|string $type
+     * @param Type|TypeCreatorExtension|string $type
      * @param callable $resolver
      */
     public function __construct($description, $type, callable $resolver)
@@ -52,7 +53,7 @@ class FieldDefinition
     }
 
     /**
-     * @return Type|string
+     * @return Type|TypeCreatorExtension|string
      */
     public function getType()
     {
--- a/src/Scaffolding/Scaffolders/DataObjectScaffolder.php
+++ b/src/Scaffolding/Scaffolders/DataObjectScaffolder.php
@@ -10,7 +10,7 @@ use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Extensible;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Dev\Deprecation;
-use SilverStripe\GraphQL\FieldDefinition;
+use SilverStripe\GraphQL\Scaffolding\Scaffolders\DataObjectScaffolder\FieldDefinition;
 use SilverStripe\GraphQL\Manager;
 use SilverStripe\GraphQL\Scaffolding\Extensions\TypeCreatorExtension;
 use SilverStripe\GraphQL\Scaffolding\Interfaces\ConfigurationApplier;
@@ -28,7 +28,6 @@ use SilverStripe\ORM\DataObjectInterface;
 use SilverStripe\ORM\FieldType\DBField;
 use SilverStripe\ORM\SS_List;
 use SilverStripe\ORM\UnsavedRelationList;
-use SilverStripe\View\ArrayData;
 
 /**
  * Scaffolds a DataObjectTypeCreator.
@@ -51,7 +50,7 @@ class DataObjectScaffolder implements ManagerMutatorInterface, ScaffolderInterfa
     ];
 
     /**
-     * @var ArrayList
+     * @var ArrayList|FieldDefinition[]
      */
     protected $fields;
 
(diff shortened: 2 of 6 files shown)


ScopeyNZ:
I feel like `ArrayData` is the wrong way to store field information. To have objects in an `ArrayData` it needs to extend `ViewableData` but why should a `FieldDefinition` extend `ViewableData`? 

I'm pretty keen to work on making this more intuitive. I might be able to come up with a BC way to implement this before 4.3 but for now this is ready to go. Hopefully it's acceptable as is, although it feels slightly hacky with the `$field->hasField('Definition')` and `$map = $field->toMap(); $map['Definition']` stuff.

ScopeyNZ:
Sounds fair. I'll try and get some time to address this before Friday.

unclecheese:
The `fields` array is already quite overloaded. You might want to break it out into something like `casting`, because there's a reasonable case for overloading the magic casting for native `db` fields, e.g. an `Int` casted as `Bool` or vice versa.

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
```diff
@@ -676,37 +707,50 @@ protected function createFields(Manager $manager)
...
-            $fieldName = $fieldData->Name;
-            if (!StaticSchema::inst()->isValidFieldName($instance, $fieldName)) {
-                throw new InvalidArgumentException(
-                    sprintf(
-                        'Invalid field "%s" on %s',
-                        $fieldName,
-                        $this->getDataObjectClass()
-                    )
-                );
+        foreach ($this->fields as $fieldName => $data) {
```
maxime-rainville: This method is getting kind of big and difficult to follow. Can we split it up into smaller private methods?

src/Scaffolding/StaticSchema.php
```diff
@@ -135,6 +135,27 @@ public function isValidFieldName(ViewableData $instance, $fieldName)
         return ($instance->hasMethod($fieldName) || $instance->hasField($fieldName));
     }
 
+    /**
+     * Throws an InvalidArgumentException given the given field name is not valid for the given instance
+     *
+     * @param ViewableData $instance
+     * @param $fieldName
+     */
+    public function assertValidFieldName(ViewableData $instance, $fieldName)
```
maxime-rainville: The entire point of this method is to throw an exception. We should include a `@throws` statement.

src/FieldDefinition.php
```diff
@@ -0,0 +1,70 @@
+<?php
+namespace SilverStripe\GraphQL;
+
+use GraphQL\Type\Definition\Type;
+use SilverStripe\GraphQL\Scaffolding\Extensions\TypeCreatorExtension;
+
+class FieldDefinition
```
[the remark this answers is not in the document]
maxime-rainville: There's no unit test for this class.

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
```diff
@@ -102,27 +104,38 @@ public function addFields(array $fieldData)
         foreach ($fieldData as $k => $data) {
             $assoc = !is_numeric($k);
             $name = $assoc ? $k : $data;
-            $field = ArrayData::create(
-                [
+
+            // If we're given a field definition we can accept that interface
+            if ($data instanceof FieldDefinition) {
```
maxime-rainville: Can we update the unit test in `tests/Scaffolding/Scaffolders/DataObjectScaffolderTest.php` to test `FieldDefinition`?

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
```diff
@@ -102,27 +104,38 @@ public function addFields(array $fieldData)
...
+     * Add a field to this Data query under the given field name, optionally providing a description or a
+     * `FieldDefinition` that defines how this field is returned.
+     *
      * @param string $field
-     * @param string $description
+     * @param FieldDefinition|string $definition
      * @return $this
      */
-    public function addField($field, $description = null)
+    public function addField($field, $definition = null)
```
maxime-rainville: Can we update the unit test in `tests/Scaffolding/Scaffolders/DataObjectScaffolderTest.php` to test `FieldDefinition`?

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
```diff
@@ -645,6 +663,20 @@ protected function nestedConnections()
         return $queries;
     }
 
+    protected function assertValidFieldData($data, $fieldName)
```
maxime-rainville: Can we document this method?

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
```diff
@@ -102,27 +104,38 @@ public function addFields(array $fieldData)
         foreach ($fieldData as $k => $data) {
             $assoc = !is_numeric($k);
             $name = $assoc ? $k : $data;
-            $field = ArrayData::create(
-                [
+
```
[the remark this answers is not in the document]
unclecheese: Let's use `ArrayLib::is_assoc()` for consistency.

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
```diff
@@ -222,14 +237,11 @@ public function setFieldDescription($field, $description)
...
 
-        $this->fields->replace(
-            $existing,
-            ArrayData::create(
-                [
-                    'Name' => $field,
-                    'Description' => $description
-                ]
-            )
+        $this->fields[$field] = ArrayData::create(
```
unclecheese: I think this should be its own subclass of `ArrayData`. That way you can do your own validation, getters/setters, etc, and everything is more expressive.

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
```diff
@@ -408,7 +420,13 @@ public function cloneTo(DataObjectScaffolder $target)
 
         foreach ($this->getFields() as $field) {
             if (StaticSchema::inst()->isValidFieldName($inst, $field->Name)) {
-                $target->addField($field->Name, $field->Description);
+                $arg = $field->Description;
+                // Work around `ArrayData` not supporting objects as values (unless they extend "ViewableData")
+                if ($field->hasField('Definition')) {
```
unclecheese: Seems to me you wouldn't have to do this hack if you created your own `ArrayData` subclass.

ScopeyNZ: I agree. I don't even think it should be a subclass of ArrayData as it's quite opinionated around holding objects that extend `ViewableData`.

ScopeyNZ:
Thanks for the reviews guys! I've blocked a couple of hours out tomorrow afternoon to run over this some more so I'll try to address everything then.

ScopeyNZ:
I've added a commit - some details on the change are in the commit message.

I refactored this PR so that the only way that fields are maintained within `DataObjectScaffolder` is by using an associative array of `FieldDefinition`s. It occurred to me that this new object is only relevant for `DataObjectScaffolder` so I moved it a more relevant namespace.

I still need to increase the test coverage which I'll try to do before next week but I appreciate any comments you guys may have in the meantime.

ScopeyNZ:
Also this is definitely breaking changes now :( - And isn't actually a dependency for the CCs elemental work in retrospect - We can drop it from the 4.3 release milestone (although it'd be awesome to get this into the new 3.0 release).

ScopeyNZ:
I guess it probably does. This has breaking changes though and it's not high on my list to address this right now.

ScopeyNZ:
I don't have access to this repo for some reason so I can't link PR (nor edit the issue to mention a PR)

robbieaverill:
>I don't have access to this repo for some reason so I can't link PR (nor edit the issue to mention a PR)

Core team should, I've fixed that

maxime-rainville:
@ScopeyNZ Do you still have work to do on this? We've got it in the _blocked_ column on the OS board.

ScopeyNZ:
I can't remember where I left it. Looks like there's at least a failed test to resolve. I'll put aside some time this week to look over it.

unclecheese:
I hate to take this conversation in a totally different direction, but I'm wondering if, since it appears version 4 will use generated code from a dev task, maybe we should rely on statically analysed type hints for this?

Option A: Merge this into 3 only, with a new static analysis implementation in 4
Option B: Merge this into 3 and deprecate it in 4, supporting both for one major release cycle
Option C: Don't merge this. Let's just wait until 4, when we can do it right.
Option D: Merge this and support both long term (because maybe there's a legit case for having a GraphQL type that differs from your PHP type?)

ScopeyNZ:
Considering that the API is changing so quickly and we're jumping to new majors so often I don't think it's worth increasing maintenance burden in the `3` branch. (Option C).

I still haven't found the time to run over this PR - it's not currently relevant for me.

maxime-rainville:
Let's close for now. We can reopen later if we find it's relevant.

PR #194, opened 2018-12-07
BUGFIX: Ensure httpMethod context is applied to all controller actions

unclecheese:


maxime-rainville:
@unclecheese Merge at your convenience.

unclecheese:
Thanks, @maxime-rainville . The issue was mostly in tests. You would do `setManager($someMock)` and then `Controller::getManager()` would see that it already had `$this->manager` assigned and return the cached reference.

The alternative would have been to make it the tester's responsibility to set a manager that already has context, which is a little more burdensome and kind of gross.

robbieaverill:
I think that's a little misleading - the problem was evident in the `types()` method which is added by an extension. That asks for the manager and gets given one without any context, because the context is only provided by the `index()` action - it's not a test-only issue. This fix looks good though! Nice work!

Issue #197, opened 2018-12-18
"continue" targeting switch is equivalent to "break"


ntd:
With PHP 7.3.0 my logs are flooded with the above warnings generated by the `webonyx/graphql-php` dependency. This in turn, at least on my configuration (nginx + php-fpm), breaks the assets section in the back end. If needed, I can provide the full log (it is quite verbose).

That issue has been resolved by `webonyx/graphql-php >= 0.13` (reference: webonyx/graphql-php#419) but `silverstripe/graphql` seems to be stuck in [the 0.8 branch](https://github.com/silverstripe/silverstripe-graphql/blob/3/composer.json#L8).

robbieaverill:
Thanks, I'll triage this as a minor semver enhancement in order to upgrade and test the new version of graphql-php. Would you like to have a go at doing it?

ntd:
Thank you for your answer, that was quick!

If you can wait a couple of days I can make a pull request (I need to close a project before) or if you already have everything in place you can bump the dependency and test for regressions by yourself. Either options are fine from me.

robbieaverill:
I think we'd be happy to wait for you if you're keen, thanks for getting involved!

ScopeyNZ:
Unfortunately we can't upgrade to the newer version _yet_ due to incompatibility with PHP <7. We will be [deprecating support for PHP 5 in a SilverStripe 4.x version at some point next year ](https://github.com/silverstripe/silverstripe-framework/issues/8556) but I think there'll be at least one more major version of this module that will have to support PHP 5 before then.

Happy to receive a PR, but it will _probably_ sit until we know what core release will drop support for PHP 5.

ntd:
I just checked and the [commit](https://github.com/webonyx/graphql-php/commit/fce0e4dd223b0cf746f2b16cca8c86831daa851f) addressing this issue has been included also in 0.12.6 (that still requires [php >= 5.6](https://github.com/webonyx/graphql-php/blob/v0.12.6/composer.json#L11)).

I'll check if `graphql-php >= 0.12.6` works with either php 7.3.0 and 5.6.39.

ntd:
I just created PR#197 with `webonyx/graphql-php=0.12.6`.

ivoba:
@robbieaverill is this already released?
I have this as well on PHP7.3 with:

        "silverstripe/recipe-plugin": "^1.2",
        "silverstripe/recipe-cms": "4.3.3@stable",

[...]

PR #198, opened 2018-12-20
PHP 7.3.0 support (#197)

ntd:
* Pin down required `webonyx/graphql-php` dependency to 0.12.6
* Upgrade codebase accordingly


ntd:
The problem is whenever I locally run the test suite I got 40 errors (`No current controller available`) and 1 failure (`SilverStripe\GraphQL\Tests\Middleware\CSRFMiddlewareTest::testItThrowsIfTokenIsInvalid`). This happens also on the original code.

ntd:
> The problem is whenever I locally run the test suite I got 40 errors ...

Nevermind: after adding `SS_DATABASE_CHOOSE_NAME='true'` to my local `.env` everything seems to be working as expected. Now I can try to clear the results.

composer.json
```diff
@@ -5,7 +5,7 @@
   "require": {
     "silverstripe/framework": "^4.2",
     "silverstripe/vendor-plugin": "^1.0",
-    "webonyx/graphql-php": "~0.8.0"
+    "webonyx/graphql-php": "=0.12.6"
```
kinglozzer: Do we really want to only use this exact release? Isn’t `~0.12.6` more appropriate?

src/Manager.php
```diff
@@ -495,7 +495,17 @@ public static function formatError(Error $exception)
             'message' => $exception->getMessage(),
         ];
 
-        $locations = $exception->getLocations();
+        try {
+            $locations = $exception->getLocations();
+        } catch (Exception $whatever) {
+            /* Ignore the exception to avoid the following error:
```
kinglozzer: This seems _very_ hacky... we’d need to investigate the cause of the error instead of blindly suppressing it

ntd:
> we’ll need to carefully review https://github.com/webonyx/graphql-php/blob/master/CHANGELOG.md for any other breaking changes too

If can be of any help, they summarize API breakages in https://github.com/webonyx/graphql-php/blob/master/UPGRADE.md

src/Manager.php
ntd: Unfortunately this exception is raised by `webonyx/graphql-php` itself and has been introduced by  webonyx/graphql-php@cf276340a4249fb426a6c4fa940fb0e5fcbd455c.

composer.json
ntd: Of course you are right :)

src/Manager.php
kinglozzer: Yuck :( do you know if that’s been raised as an issue on the `webonyx/graphql-php` Github? I think if we suppress this, we run the risk of incorrectly suppressing other errors :/

robbieaverill: A safer disgusting hack might be to do some text checking on the exception message so you can re-throw exceptions that aren't this one

ntd: > do you know if that’s been raised as an issue on the `webonyx/graphql-php` ...

I am pretty sure nothing has been raised because when looking for information I did not find anything. I would have opened that issue by myself but I don't know the caller code well enough.

robbieaverill: Right so it’s a test only problem? That’s good

ntd: Yes. The bug has always been there and webonyx/graphql-php@cf276340a4249fb426a6c4fa940fb0e5fcbd455c exposed it.

ntd:
> [...] we’ll need to carefully review [...] for any other breaking changes [...]

Not sure what to do now. I'm certainly not familiar enough with the code to do a careful review.

robbieaverill:
@ntd I think what you've done already is awesome. We'll need to leave some time for the @silverstripe/core-team to review your proposed changes so we can get it merged in.

robbieaverill:
In terms of @kinglozzer's comment about the underlying GraphQL library's namespaces changing, I'm not concerned about that (in terms of our semver compliance) since it's a third-party library and not part of our API.

I've rebased the PR onto the latest 3 branch.

ScopeyNZ:
I think @unclecheese has also done this work on his POC branch for schema cacking. It'd be good to "compare notes" - especially as there'll eventually be a conflict to resolve 😅

PR #200, opened 2019-01-08
BUG Ensure queries are sorted before limiting

### LINKED ISSUE #199: BUG Connection limit / sort in wrong order
Checking the `Connection` class I noticed that sorting and limiting is done in the incorrect order; If applying both a sort and limit, the list is first limited, and then sorted.

When the base list is an in-memory list (rather than a DataList) this will limit first the un-sorted list, and then apply sorting only to the Nth page of the list.

With a datalist it's fine, because it's lazy-resolved, however.

tractorcow:
Fixes #199

### DIFF
--- a/src/Pagination/Connection.php
+++ b/src/Pagination/Connection.php
@@ -416,6 +416,7 @@ class Connection implements OperationResolver
      * @param SS_List $list
      * @param array   $sortBy
      * @return SS_List Sorted list, if sortable
+     * @throws InvalidArgumentException If an invalid sort column is specified
      */
     protected function applySort($list, $sortBy)
     {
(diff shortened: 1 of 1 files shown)


src/Pagination/Connection.php
```diff
@@ -370,66 +370,84 @@ public function resolve($value, array $args, $context, ResolveInfo $info)
...
+
+        // convert the input from the input format of field, direction
+        // to an accepted SS_List sort format.
+        // https://github.com/graphql/graphql-relay-js/issues/20#issuecomment-220494222
+        $sort = [];
+        foreach ($sortBy as $sortInput) {
+            if (isset($sortInput['field'])) {
+                $direction = isset($sortInput['direction']) ? $sortInput['direction'] : 'ASC';
+                if (!array_key_exists($sortInput['field'], $sortableFields)) {
+                    throw new InvalidArgumentException(sprintf(
```
robbieaverill: I think we should document this exception in the phpdoc

robbieaverill:
I've merged this up to master @tractorcow

Issue #192, opened 2018-12-03
RFC: Make schema serialisable and cachable


unclecheese:
## Pull Requests

- [ ] https://github.com/silverstripe/silverstripe-graphql/pull/202

## Overview

The performance limitations of this module are well documented and long-lived. Attempts to solve the issue have shown limited success. While the recent work to [cache the resolver computations](https://github.com/silverstripe/silverstripe-graphql/pull/166) has been merged and provides notable benefits, this will never be a scalable offering until we can find a way around building the schema dynamically on every request.

The most obvious and least complex way to do this is to cache the schema and invalidate it when it changes. Those are both hard, because, respectively:

* The schema is built using a plethora of anonymous functions and other runtime context
* The schema is designed to take on implicit changes as DataObjects change their definitions.

## Suggested approach

This can be done with three reasonably-sized changes that would not break any APIs.

### 1: Remove all anonymous functions from the schema

An inventory of all the anonymous functions being used in type and query creators shows that most of them are paranoid countermeasures against `Manager` race conditions. Due to the declarative nature of the schema, the manager state is not always ready to define a query, and therefore, wrapping everything in a closure is the easiest safeguard. Most of these functions do not appear to be necessary if we implement lazy loading correctly.

#### Lazy load types properly

A recent addition to the `graphql-php` API was the `typeLoader` property, which allows users to define a global accessor for all types:

```php
new Schema([
  'typeLoader' => function ($name) use ($someState) {
    return $someState[$name];
  }
]);
```

The only constraint is that type instances must be singletons.

#### Eliminate lazy loading of fields

Most of the closures around field definitions are protections against attempting to access type instances too early.  A lazy loaded type layer would eliminate the need for most of these workarounds.

#### Encourage concrete resolvers

Right now, nearly all resolvers that ship with the module are defined at runtime. The benefits of moving these to class definitions would be twofold:

* The fields that use them would be serializable
* They would become injectable.

Before:
```php
class Read extends ListQueryScaffolder
{
  protected function createResolver()
  {
    return function () { ... };
  }
}
```

After:
```php
class Read extends ListQueryScaffolder
{
  $dependencies = [
    'Resolver' => ResolverInterface::class . '.read',
  ];
}
```

Note that it is already possible to do this with the scaffolder. You can simply point the resolver at an instance, or a FQCN of a `ResolverInterface`.

Attempts to serialise a field using a closure as a resolver will throw.

### 2. Make type and field creators serialisable

[...]

PR #202, opened 2019-01-08
WIP: Schema caching

unclecheese:
## Problem

The GraphQL schema is generated on the fly for every request. Depending on how populated your schema is, this can become quite expensive and makes scaling to large projects implausible. [Benchmarks show](https://github.com/silverstripe/silverstripe-graphql/issues/192#issuecomment-443972793) that a response of **30-40** seconds wouldn't be atypical for a decent sized project.

## What this does

This pull request introduces a number of low-level architectural and API changes to the graphql module which afford it layer than can persist the generated schema as a side effect of a build task and subsequently fetched in its fully inflated state on demand with minimal latency. With the benefit of lazy-loaded types, response times scale independently from the density of the schema, and are abated only by common and manageable bottlenecks of the request cycle (e.g. ORM performance). [Benchmarks show](https://github.com/silverstripe/silverstripe-graphql/issues/192#issuecomment-449199812) that a schema with more than 1,000 types (100 dataobjects) produces a `readOne` response **in under 100ms** on a moderately outfitted SS Platform virtual stack.

## Major API Changes

* Schema must be built using a `dev/` task instead of in the request cycle.
* Decoupling from the `webonyx/graphql-php` library. Type creators and scaffolders no longer return `GraphQL\Type\Definition\Type`, but rather our own implementation of graphql type abstractions.
* Resolvers can no longer be anonymous functions. Use static callables, e.g. `[My::class, 'someFunc']` or, for dynamic functions, an implementation of `ClosureFactoryInterface` to build closures lazily.
* `Manager` now has a `SchemaStorageInterface` dependency.
* Middleware callback signature now takes an abstract `Schema` object instead of `GraphQL\Type\Definition\Schema`.

## How it works (high level)

The solution uses code generation to store a monolithic PHP class in a temporary folder that is later fetched, included and given to the schema as a [lazy lookup for types](https://webonyx.github.io/graphql-php/type-system/schema/#lazy-loading-of-types).

This has a major footprint on developer ergonomics. Developers will now have to run a task to generate a new schema every time it changes. To generate the schema, run `/dev/schema/[SCHEMA-KEY]`.

Schema caching is compulsory. There is no backward compatibility for dynamically created and cached schemas. This breakage is completely defensible citing that status quo performs so poorly. It should be viewed as a flaw in the system more than an API worth protecting.

### Current request cycle (status quo)

```
[ HTTPRequest ]
    ↓
    ↓
[ Controller ]
    ↓
    ↓
[ Manager ]  ← ← [ GraphQL Types ] ← ←  [ Scaffolding / Type Creators ]
    ↓
    ↓
[ Manager::query() ]
    ↓
    ↓
[ GraphQL:execute() ]
    ↓
    ↓
[ ExecutionResult ]
    ↓
    ↓
[ HTTPResponse ]
```

This pull request introduces architectural changes that add layers of abstraction between all these points and decouple everything. 

#### GET /dev/schema/[schema-key]

```
[ Manager::regenerate() ]
    ↓
    ↓
[ Manager ]  ← ← [ Type abstractions ] ← ←  [ Scaffolding / Type Creators ]
    ↓
    ↓
[ SchemaStorageInterface::persist() ]
    ↓
    ↓
[ TypeRegistryEncoderInterface::encode() ]
    ↓
    ↓
[ /tmp/cached-schema.php ]
```

#### GET /graphql/?query=...

```
[ HTTPRequest ]
    ↓
    ↓
[ Controller ]
    ↓
    ↓
[ Manager ]
    ↓
    ↓
[ SchemaStorageInterface::load() ]
    ↓
    ↓
[ Abstract schema ]  ← ← [ TypeRegistryInterface ]
    ↓
    ↓
[ SchemaHandlerInterface ] 
    ↓
    ↓
[ SchemaHandlerInterface::query() ]
    ↓
    ↓
[ QueryResultInterface ]
    ↓
    ↓
[ HTTPResponse ]
```

Everything is now decoupled and agnostic of the `graphql-php` library. In theory, you'd only need a new implementation of the encoders and the schema storage service to export your schema to another library, or even another platform, like Node.

### Abstractions

One major API change is that scaffolders and type creators can no longer return concrete `graphql-php` types. Instead, the entire system uses its own library of schema primitives, e.g. `Enum`, `Union`, `Argument`, and `FieldCollection` (for objects), which are essentially value objects with fluent APIs that friendly to the encoder, accepting only those properties which encode easily and enforce internal rules such as the use of `ClosureFactoryInterface` instead of anonymous functions.

Additionally, in type creators, fields should no longer be returned as plain arrays, but rather use the `Field` abstraction. Backward compatibility has been added via `Field::createFromConfig(array $config)`.

### Referential types

One of the challenges with scaffolders writing directly to `graphql-php` types is that the types have to be referentially equal across the whole schema. This introduced all kinds of race conditions and confusing workarounds using lazily executed functions. By working with our own set of abstract types, we can avoid that and make it the job of the encoder to build our references.

The `TypeReference` class allows you to build a reference to a type by name only, not instance, e.g. `->setType(TypeReference::create('File'))` will encode to `$this->getType('File')` and `TypeReference::create('ID')` will encode to `Type::id()`.

For ergonomics, shortcuts are provided to explicitly use internal types, e.g. `InternalType::id()`.

### New interfaces (partial list)

|Interface name|Description|Default implementation|
|--- |--- |--- |
|SchemaStorageInterface|A dependency of `Manager`. All manager instance must store their schemas.|CodeGenerationSchemaStore|
|TypeRegistryEncoderInterface|Takes a list of type abstractions and converts them to executable PHP code (for `graphql-php` or anything else). Supplies an instance of the cached type registry on demand.|GraphQLPHP\SchemaEncoder|
|SchemaHandlerInterface|A dependency of `Manager`. Handles a query against a given abstract schema and gets back a `QueryResultInterface`|GraphQLPHP\SchemaHandler|
|TypeEncoderInterface|Converts a given abstract type into PHP code. These are stored in a TypeEncoderRegistryInterface. The first one to return true on `appliesTo(AbstractType)` gets the job.|e.g. GraphQLPHP\Encoders\ObjectTypeEncoder|
|ClosureFactoryInterface|Anonymous functions are not encodable, for obvious reasons, so for attributes of types that must be callable (e.g. resolvers) but cannot be declared statically, these implementations provide a `createClosure()` method which returns the function just-in-time as part of the creation of the type in the registry. A the time `createClosure()` is executed, it has access to any context the scaffolders have provided to it.|DynamicFunction, StaticFunction|

## New boilerplate for a Manager instance

```yml
  SilverStripe\GraphQL\Manager.default:
    class: SilverStripe\GraphQL\Manager
    constructor:
      schemaKey: default
      schemaStore: '%$SilverStripe\GraphQL\Schema\SchemaStorageInterface.default'
    properties:
      Middlewares:
        CSRFMiddleware: '%$SilverStripe\GraphQL\Middleware\QueryMiddleware.csrf'
        HTTPMethodMiddleware: '%$SilverStripe\GraphQL\Middleware\QueryMiddleware.httpMethod'
  SilverStripe\GraphQL\Schema\SchemaStorageInterface.default:
    class: SilverStripe\GraphQL\Schema\Encoding\CodeGenerationSchemaStore
    constructor:
      registryEncoder: '%$SilverStripe\GraphQL\Schema\Encoding\Interfaces\TypeRegistryEncoderInterface.default'
  SilverStripe\GraphQL\Schema\Encoding\Interfaces\TypeRegistryEncoderInterface.default:
    class: SilverStripe\GraphQL\GraphQLPHP\Encoders\SchemaEncoder
    constructor:
      identifier: default
      encoderRegistry: '%$SilverStripe\GraphQL\Schema\Encoding\Interfaces\TypeEncoderRegistryInterface'
  SilverStripe\GraphQL\Schema\SchemaHandlerInterface:
    class: SilverStripe\GraphQL\GraphQLPHP\SchemaHandler
```

#### Do we really need a homespun library of type abstractions?

Because we're relying on code generation, we have to define our own rules on the composition of a type. If we were to allow the scaffolders to simply export `graphql-php` types, we would have to blacklist a lot of their API surface (e.g. `astNode`, `mapFn` and `resolver` as a Closure). It seems inappropriate to use `graphql-php` types as value objects that are simply exported to encoders  that encode them right back.

#### Can't the schema cache just invalidate itself?

Probably. The schema build should be keyed to the config sha, and `dev/build` should build the schema when it changes.  There are still a number of questions open on how to create the best developer experience. A new endpoint to remember and the conditions that make it a requirement is a nontrivial burden to developers.

#### Is the generated schema a black box? What if I want to load my own custom types in?

You can side load your own raw `graphql-php` types into the type registry using `Manager::setRegistryExtensions(TypeRegistryInterface[])`. This creates a failover system in the type registry, much like `ViewableData::customise()`, where if the type isn't found, it looks it up in the custom extension.

```yml
SilverStripe\GraphQL\Manager.default:
  properties:
    registryExtensions:
      - %$My\TypeRegistryInterface
```

## Related pull requests

 * [ ] https://github.com/silverstripe/silverstripe-frameworktest/pull/60
 * [ ] https://github.com/open-sausages/silverstripe-admin/tree/pulls/4/schemageddon
 * [ ] https://github.com/open-sausages/silverstripe-asset-admin/tree/pulls/1/schemageddon

## Still to do
* Update README
* Finish/fix unit tests
* Get working with admin (need an Interface encoder and some resolver factories)
* Guardrails and deprecations (e.g. automatically upgrading old `graphql-php` types to their corresponding abstractions)


chillu:
Added a few minor changes as commits, and started fixing up the various dependencies:
 * https://github.com/silverstripe/silverstripe-frameworktest/pull/60
 * https://github.com/open-sausages/silverstripe-admin/tree/pulls/4/schemageddon
 * https://github.com/open-sausages/silverstripe-asset-admin/tree/pulls/1/schemageddon

unclecheese:
Superceded by https://github.com/silverstripe/silverstripe-graphql/pull/266

PR #207, opened 2019-01-30
FIX Check readOne permissions against resolved DataObject

ScopeyNZ:
In most cases the canView permission was intended to be checked against the actual object from the database as the object state may have some play on whether the object is accessible by the user.

This was reported in an issue on silverstripe-elemental: https://github.com/dnadesign/silverstripe-elemental/issues/586 . I also have a "custom resolver" fix in that repo but I figure this is a common use case and it's a pretty quick win to get this fixed here.

### DIFF
--- a/src/Scaffolding/Scaffolders/CRUD/ReadOne.php
+++ b/src/Scaffolding/Scaffolders/CRUD/ReadOne.php
@@ -60,16 +60,20 @@ class ReadOne extends ItemQueryScaffolder implements OperationResolver, CRUDInte
      */
     public function resolve($object, array $args, $context, ResolveInfo $info)
     {
-        if (!$this->getDataObjectInstance()->canView($context['currentUser'])) {
+        // get as a list so extensions can influence it pre-query
+        $list = DataList::create($this->getDataObjectClass())
+            ->filter('ID', $args['ID']);
+        $this->extend('updateList', $list, $args, $context, $info);
+
+        $item = $list->first();
+
+        // Check permissions on the individual item as some permission checks may investigate saved state
+        if (!$item->canView($context['currentUser'])) {
             throw new Exception(sprintf(
                 'Cannot view %s',
                 $this->getDataObjectClass()
             ));
         }
-        // get as a list so extensions can influence it pre-query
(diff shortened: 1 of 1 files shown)


src/Scaffolding/Scaffolders/CRUD/ReadOne.php
```diff
@@ -60,16 +60,20 @@ protected function createDefaultArgs(Manager $manager)
...
-        if (!$this->getDataObjectInstance()->canView($context['currentUser'])) {
+        // get as a list so extensions can influence it pre-query
+        $list = DataList::create($this->getDataObjectClass())
+            ->filter('ID', $args['ID']);
+        $this->extend('updateList', $list, $args, $context, $info);
+
+        $item = $list->first();
+
+        // Check permissions on the individual item as some permission checks may investigate saved state
+        if (!$item->canView($context['currentUser'])) {
```
robbieaverill: Possible null pointer error here

ScopeyNZ: Point taken. If it's null I guess I should fallback to the old logic. The expected result should be "can't view" instead of "item doesn't exist"

robbieaverill: I didn’t really think about the context. I guess it’s fine the way it is

ScopeyNZ: I think it still makes sense to defensively program against the null case but I'll do it in a follow up.

robbieaverill: Yeah we should've - kitchen sink just flagged it. I'll make a follow up PR

robbieaverill:
I think I merged this prematurely.

Old behaviour: get a singleton of the class and use that to check canView()

New behaviour (from this patch): get the specific item from the list and use that to check canView(), but don't handle null pointers (valid return from `DataList::first()`)

Patch at #208 will use the new behaviour but fall back to the old behaviour of checking a singleton if no specific instance is available

PR #219, opened 2019-02-19
FIX: Return HTTP 400 for an incomplete query

### LINKED ISSUE #218: 500 error visiting admin/graphql/types
If you just visit the url admin/graphql/types, you get a 500 error, and e.g. a Raygun error.

This should bee a 4xx level error, maybe 400, and not throw an exception.

Occurs in 3.x

sminnee:
This prevents exceptions from being thrown for issues caused by invalid
user-input, preventing spam from being introduced to error monitors
such as Raygun.

This fixes #218

### DIFF
--- a/src/Manager.php
+++ b/src/Manager.php
@@ -301,6 +301,10 @@ class Manager implements ConfigurationApplier
                     }, $this->queries);
                 },
             ]);
+        } else {
+            $schema[self::QUERY_ROOT] = new ObjectType([
+                'name' => 'Query',
+            ]);
         }
 
         if (!empty($this->mutations)) {
@@ -311,6 +315,10 @@ class Manager implements ConfigurationApplier
                         return is_callable($mutation) ? $mutation() : $mutation;
                     }, $this->mutations);
                 },
+            ]);
+        } else {
+            $schema[self::MUTATION_ROOT] = new ObjectType([
+                'name' => 'Mutation',
             ]);
         }
 


sminnee:
I'd suggest a quick 3.0.1 tag for this.

sminnee:
It's worth noting @chillu that the 4.3 ModelAdmin and Security UIs are triggering this, at least when asset-admin/cms aren't installed.

sminnee:
Turns out this change is completely wrong, and the URL should return some sort of value...

sminnee:
OK having seen that this URL should have been returning a value, this is an alternative cut.

With this change, my project now returns this for admin/graphql/types:

```
{"data":{"__schema":{"types":[{"kind":"OBJECT","name":"Query","possibleTypes":null},{"kind":"OBJECT","name":"Mutation","possibleTypes":null},{"kind":"OBJECT","name":"__Schema","possibleTypes":null},{"kind":"OBJECT","name":"__Type","possibleTypes":null},{"kind":"ENUM","name":"__TypeKind","possibleTypes":null},{"kind":"SCALAR","name":"String","possibleTypes":null},{"kind":"SCALAR","name":"Boolean","possibleTypes":null},{"kind":"OBJECT","name":"__Field","possibleTypes":null},{"kind":"OBJECT","name":"__InputValue","possibleTypes":null},{"kind":"OBJECT","name":"__EnumValue","possibleTypes":null},{"kind":"OBJECT","name":"__Directive","possibleTypes":null},{"kind":"ENUM","name":"__DirectiveLocation","possibleTypes":null},{"kind":"ENUM","name":"VersionedStage","possibleTypes":null},{"kind":"ENUM","name":"VersionedStatus","possibleTypes":null},{"kind":"ENUM","name":"VersionedQueryMode","possibleTypes":null},{"kind":"INPUT_OBJECT","name":"VersionedInputType","possibleTypes":null},{"kind":"SCALAR","name":"Int","possibleTypes":null},{"kind":"INPUT_OBJECT","name":"CopyToStageInputType","possibleTypes":null},{"kind":"SCALAR","name":"ID","possibleTypes":null},{"kind":"OBJECT","name":"SilverStripe_Assets_Storage_DBFile","possibleTypes":null},{"kind":"SCALAR","name":"Float","possibleTypes":null}]}}}
```

sminnee:
@unclecheese I don't really know what the "correct" response is here. What I _do_ know is that the my admins are fetching this URL in spite of there being no usable graphql data to provide.

This seems to be because:

 * boot/index.js
 * Always calls boot/apollo/buildClient.js
 * Which always calls boot/apollo/getGrapqhlClients.js
 * Which always fetches the types URL admin/graphql/types (assuming the assets cache ain't avail)

It's a slight inefficiency given that I have neither campaign-admin nor asset-admin installed, but given our general march towards GraphQLization that seems fine.

The question is whether we return a 404 for this URL ("no graphql to see here, ma'am") or whether we return the anaemic result shown above that this PR provides.

sminnee:
@unclecheese FYI this is of relevance

unclecheese:
Right, this makes sense. This issue presents itself in other annoying ways, too. Like when you load the graphiql client before you've added anything to the schema, it just blows up and it makes it look like something is wrong with the tool.

There should be a better empty state for the schema.

unclecheese:
Pretty sure the new version of `graphql-php` fails gracefully, so we should just take care when merging this up.

sminnee:
Yay, thanks!

Is it appropriate to do a cheeky 3.0.1 tag on this module?

unclecheese:
There already is a 3.0.1! :-b

Ideally, we'd issue a new tag, but outside of `silverstripe-config` there's really no precedent for that. All of our other recipe-cms modules are in lockstep with each other, which makes pairing the tags really easy. (e.g. graphql 3.0.1 -> SS 4.3.1, graphql 2.0.4 -> SS 4.2.4).

Not saying it's a great thing, but I'm not sure why we'd just arbitrarily end that now without more discussion.

sminnee:
OK. I kicked off an internal Slack discussion about that.

Issue #68, opened 2017-02-02
Add search params to scaffolded READ operations


unclecheese:
Right now, the scaffolded `READ` operations do not provide any meaningful API for adding `$args`. You can add them, but because the resolver is pre-baked, you can't apply them. You would have to use your own resolver, which defeats the purpose of using scaffolding. 

This would be a great opportunity to use and already established convention such as `SearchContext`.

Possible API:

```
types:
  BlogPost:
    fields: [Title, Content, Author]
    operations:
      read:
        search:
          Title:
            description: Filter the posts by title
          Created:
            description: Filter the posts by their creation date
          Author__Surname:
            description: Search by the author's surname
```

The query would then accept two arguments: `$Filter` and `$Exclude`, each of which would take the above filters.

```
readPosts($Filter, $Exclude) {

}

{
	"Filter": {
	  "Title": { 
	  	"Value": "some search text here",
	  	"Filter": "ExactMatch" // An enum of SearchFilter subclasses
	  },
	  "Created" {
	  	"Value": "2016-01-01",
	  	"Fitler": "GreaterThan"
	  }
	},
	"Exclude": {
		"Author__Surname": {
			"Value": "Smith",
			"Filter": "ExactMatch"
		}
	}
}
```

We could take it a step further to clean up some of the parity with SS `SearchFilter` classes, and alias them.

```
{
	"Filter": {
	  "Title": { 
	  	"Value": "some search text here",
	  	"Comparison": "="
	  },
	  "Created" {
	  	"Value": "2016-01-01",
	  	"Comparison": ">"
	  }
	}
}
```

@chillu has shared some concerns about using `SearchContext` in GraphQL due to complexities like composite filters (e.g. range).

## Pull Requests
- [ ] https://github.com/silverstripe/silverstripe-graphql/pull/220

chillu:
* I'm torn between a strongly typed and verbose API that you propose, and a more flexible and readable DSL (a single `query` string argument). Supporting SearchFilter is definitely an easier goal - and the lack of advanced features like filter grouping isn't a showstopper since you can always create more sophisticated resolvers.
 * The [Github GraphQL API](https://developer.github.com/early-access/graphql/object/query/) has it's own [search syntax](https://help.github.com/articles/search-syntax/) through the `query` argument, getting around this issue. But then again, they use Elastic Search behind the scenes, so don't have to do large DSL translations into an ORM
[...]

PR #220, opened 2019-02-21
NEW: Add search params, filtering service for queries

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-graphql/issues/68

## Summary

This pull request adds an API for adding search parameters to graphql queries. Based on a loosely-coupled service, `DataObjectQueryFilter`, it is compatible with both scaffolded (`read`, `readOne`) operations as well as custom queries generated with `QueryCreator`.

## What it looks like
```yaml
Blog:
  operations:
    read:
      # filters: '*' -> Adds all default filters for every $db field
      filters:
        Title: true ## Adds all default filters for this field type
        CommentCount:
           gt: true # Add a greater than filter
           lt: true # Add a less than filter
        Categories__Title: true
        Hidden: true
```

Query:
```graphql
query readBlogs(
  Filter: {
    Title__contains: "food",
    CommentCount__gt: 5,
    Categories__Title__in: ["Recipes", "Cooking tips"]
  },
  Exclude: {
    Hidden__eq: true
  }
)
```

## What are "default filters"?

Not every filter should be applied to every data type by default. For instance, it doesn't make a whole lot of sense to use `gt` (greater than) on a `Varchar` field. Therefore, a short list of filters are enabled by default on each `DBField` subclass. (See `filters.yml`). Assigning `MyField: true` opts you into those defaults. You can always add `gt` explicitly if you want to.

## Procedural API

While the scaffolded operations (`Read`, `ReadOne`) manage the filter service implicitly, if you're working with a query at a lower level (e.g. `QueryCreator`), you'll have to do this on your own.

```php
$filter = new DataObjectQueryFilter(MyObject::class);
$filter->addAllFilters(); // Add all default filters for every $db field
$filter->addDefaultFilters('MyField');
$filter->addFieldFilterByIdentifier('MyField', 'gt');

// Add the input type as an argument
[
  'Filter' => $filter->getInputType('MyFilterInputTypeName'),
  'Exclude' => $filter->getInputType('MyExcludeInputTypeName'),
]

// Apply it in your resolver
$list = $filter->applyArgsToList($list, $args);
```

## This summary is shit

More detail in the README.


README.md
```diff
@@ -811,6 +939,42 @@ $scaffolder->type(MyDataObject::class)
...
+
+#### Adding search params (read operations only)
+
+You can add all default filters for every field on your dataobject with `filters: '*'`.
+
+```yaml
+read:
+  filters: '*'
+``` 
+> Note: "every field" means every field on the dataobject itself -- not just those exposed on its GraphQL type.
```
chillu: Do you think that's a good default? It seems a bit weird to allow filtering on fields that you can't view. Depending on the context, there might be a good reason not to expose that field. For example, I could start guessing email addresses on `readMember()` even though the developer made the explicit decision not to expose email addresses through the API. I'd simply loop over a list of potential addresses until I get a match.

_config/filters.yml
```diff
@@ -0,0 +1,30 @@
...
+SilverStripe\ORM\FieldType\DBField:
+  graphql_type: String
+  default_filters: [ eq, in ]
+SilverStripe\ORM\FieldType\DBInt:
+  graphql_type: Int
+  default_filters: [ gt, lt, gte, lte ]
+SilverStripe\ORM\FieldType\DBFloat:
+  graphql_type: Float
+  default_filters: [ gt, lt, gte, lte ]
+SilverStripe\ORM\FieldType\DBString:
```
chillu: Did you leave out `DBDate`, `DBTime` and `DBDateTime` on purpose here? They should have `eq, gt, lt, gte, lte` right?

_config/filters.yml
```diff
@@ -0,0 +1,30 @@
...
+      lte: '%$SilverStripe\GraphQL\QueryFilter\Filters\LessThanOrEqualFilter'
+      in: '%$SilverStripe\GraphQL\QueryFilter\Filters\InFilter'
+      endswith: '%$SilverStripe\GraphQL\QueryFilter\Filters\LessThanFilter'
+      startswith: '%$SilverStripe\GraphQL\QueryFilter\Filters\LessThanFilter'
+  SilverStripe\GraphQL\QueryFilter\DataObjectQueryFilter:
+    properties:
+      filterRegistry: '%$SilverStripe\GraphQL\QueryFilter\FilterRegistryInterface'
+SilverStripe\ORM\FieldType\DBField:
+  graphql_type: String
+  default_filters: [ eq, in ]
```
chillu: I think this should be prefixed with `graphql_ default_filters`, because those are different from `ORM/Filters` (in the same namespace as `DBField`)

src/QueryFilter/DataObjectQueryFilter.php
```diff
@@ -0,0 +1,389 @@
...
+        return $this;
+    }
+
+    /**
+     * Adds all the default filters for every field on the dataobject
+     * @return $this
+     */
+    public function addAllFilters()
+    {
+        $fields = array_keys(DataObject::getSchema()->databaseFields($this->getDataObjectClass()));
```
chillu: Like I said above, I think this should filter based on the fields exposed on the equivalent GraphQL type

src/QueryFilter/DataObjectQueryFilter.php
```diff
@@ -0,0 +1,389 @@
...
+    {
+        return !empty($this->filteredFields);
+    }
+
+    /**
+     * @param $fieldName
+     * @param $filterIdentifier
+     * @return $this
+     */
+    public function addFieldFilter($fieldName, $filterIdentifier)
```
chillu: I'm wondering if we need to allow setting instances in addition to filter identifiers, rather than getting them from the FilterRegistry. I might want my `eq` filter to work with `lat+lng` for geospatial fields, but still work with the normal `ExactMatch` filter for all other fields using this identifier. At the moment, I'd need to define `eq_geo` or something, making the API surface a bit less consistent.

src/QueryFilter/DataObjectQueryFilter.php
```diff
@@ -0,0 +1,389 @@
...
+
+        return false;
+    }
+
+    /**
+     * @param string $fieldName
+     * @param string $id
+     * @return $this
+     */
+    public function removeFilterFromField($fieldName, $id)
```
chillu: `addFieldFilter` vs `removeFilterFromField` feels a bit inconsistent?

src/QueryFilter/QueryFilterAware.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $this->queryFilter = $filter;
+
+        return $this;
+    }
+
+    /**
+     * @return DataObjectQueryFilter
+     */
+    public function queryFilter()
```
chillu: Shouldn't this be `getQueryFilter()` for consistency?

src/Scaffolding/Scaffolders/CRUD/Read.php
```diff
@@ -24,6 +41,10 @@ class Read extends ListQueryScaffolder implements OperationResolver, CRUDInterfa
     public function __construct($dataObjectClass)
     {
         parent::__construct(null, null, $this, $dataObjectClass);
+        $filter = Injector::inst()->create(DataObjectQueryFilter::class, $dataObjectClass)
```
chillu: Feels verbose to set the same keys in here which are already defined as the default in `DataObjectQueryFilter`?

src/Scaffolding/Scaffolders/CRUD/ReadOne.php
```diff
@@ -43,11 +54,31 @@ public function getName()
      */
     protected function createDefaultArgs(Manager $manager)
     {
-        return [
+        $args = [
```
chillu: I guess "ID" is a special enough case for `readOne` to elevate it to a toplevel filter.

src/QueryFilter/QueryFilterAware.php
chillu: For fluent purposes

README.md
unclecheese: Have changed to `searchableFields`

_config/filters.yml
unclecheese: Have changed to `graphql_default_filters`

_config/filters.yml
unclecheese: Done

Issue #217, opened 2019-02-14
Nested queries don't respect can* permissions


chillu:
The `canView()` checks are implemented in `Read`, so are checked on top level queries. But not in `ListQueryScaffolder` which is used for `nestedQueries` (through `DataObjectScaffolder->nestedQueries()`.

The solution might be the same as for [GraphQL logic doesn't apply to custom resolvers](https://github.com/silverstripe/silverstripe-versioned/issues/211)

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-graphql/pull/222

PR #222, opened 2019-02-22
NEW: Add universal permission checking for queries

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-graphql/issues/217

### DIFF
--- a/src/Pagination/Connection.php
+++ b/src/Pagination/Connection.php
@@ -9,6 +9,7 @@ use InvalidArgumentException;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\GraphQL\OperationResolver;
+use SilverStripe\GraphQL\Permission\PermissionCheckerAware;
 use SilverStripe\ORM\Limitable;
 use SilverStripe\ORM\Sortable;
 use SilverStripe\ORM\SS_List;
@@ -35,6 +36,7 @@ use SilverStripe\ORM\SS_List;
 class Connection implements OperationResolver
 {
     use Injectable;
+    use PermissionCheckerAware;
 
     /**
      * @var string
@@ -400,6 +402,10 @@ class Connection implements OperationResolver
             if ($offset > 0) {
                 $previousPage = true;
             }
+        }
+        if ($checker = $this->getPermissionChecker()) {
+            $currentUser = $context['currentUser'];
+            $list = $checker->applyToList($list, $currentUser);
         }
 
         return [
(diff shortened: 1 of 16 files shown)


src/Pagination/Connection.php
```diff
@@ -401,6 +403,10 @@ public function resolveList($list, array $args, $context = null, ResolveInfo $in
                 $previousPage = true;
             }
         }
+        if ($checker = $this->getPermissionChecker()) {
+            $currentUser = $context['currentUser'];
+            $list = $checker->applyToList($list, $currentUser);
```
bergice: Alterations to `edges` should also be reflected in `pageInfo`.

src/Pagination/Connection.php
```diff
@@ -401,6 +403,10 @@ public function resolveList($list, array $args, $context = null, ResolveInfo $in
                 $previousPage = true;
             }
         }
+        if ($checker = $this->getPermissionChecker()) {
+            $currentUser = $context['currentUser'];
```
bergice: `$context['currentUser']` returns a `SilverStripe\GraphQL\Pagination\Connection` instance rather than a `Member` for some reason.

bergice: This breaks `admin/assets`

src/Permission/CanViewPermissionChecker.php
```diff
@@ -0,0 +1,40 @@
...
+use SilverStripe\Security\Member;
+
+class CanViewPermissionChecker implements QueryPermissionChecker
+{
+    /**
+     * @param SS_List $list
+     * @param Member|null $member
+     * @return ArrayList|SS_List
+     */
+    public function applyToList(SS_List $list, Member $member = null)
```
bergice:
```suggestion
    public function applyToList(Filterable $filterable, Member $member = null)
```

bergice:
Seems to work as expected, but is currently breaking `admin/assets`

src/Pagination/Connection.php
unclecheese: Yup. This has been a bug in SilverStripe for as long as anyone can remember. You can't have accurate pagination and computational `canView` at the same time. We've always accepted that pagination is incorrect at least one item in the list fails the canView() check. This is true in GridFields, as well. 

Probably won't be resolved until we have a proper ACL.

src/Pagination/Connection.php
unclecheese: The surrogate query patch a few months back broke it. Fixed in latest commit.

unclecheese:
Thanks for the review. Ready for another look when you get a chance, @bergice .

tractorcow:
Doesn't this cause problems when trying to paginate over huge datasets? You'll need to load the entire list into memory to check canView() one at a time.

I guess we just need to override it on a case by case basis; I would hope asset-admin would have a smart one at least, that could optimise based on CanViewType etc. :)

Poor kittens :)

PR #223, opened 2019-02-25
allow CORS headers to be extended via an extension

simonwinter:
I'd like to be able to add specific headers for some CORS headers.

Currently we can only allow certain headers - I'd like to add a header like so:

```
Access-Control-Allow-Credentials: 'true'
```

I've simply added in an extension hook to the `addCorsHeaders` function in the Controller.

We could also have this set up as key-value pairs in the module configuration. 

Any thoughts?


chillu:
As it happens, this particular header was added through another recent PR: https://github.com/silverstripe/silverstripe-graphql/pull/227

Issue #215, opened 2019-02-13
Fix WithDescendants union type regression in scaffolding


chillu:
The docs say that you get `<type>WithDescendants`, but you don't. It's unclear when that happend, but it's a regression which versioned-admin and elemental now rely on. They can't just change their query to use descendant based logic, because that would force them to dynamically create those queries (based on the available types in a particular SilverStripe project).

We could fix this through the [https://github.com/silverstripe/silverstripe-graphql/issues/209](RFC: Replace "descendant unions" with fields to simplify querying), but that seems like a long time to leave this regression in the module.

This has also come up with devs on the #graphql channel just now

chillu:
@unclecheese Just to minimise confusion here, could you validate my assumptions here?
- At some point in the past, `<typeWithDescendants>` worked, and was documented accordingly
- Then it broke, at least in 3.0.0
- versioned-admin and elemental started relying on this broken behaviour without unions
- We fixed this with https://github.com/silverstripe/silverstripe-graphql/pull/182
- Now versioned-admin and elemental are broken. And the only reason this isn't a widespread problem is because we haven't done a graphql 3.x stable release since https://github.com/silverstripe/silverstripe-graphql/pull/182

chillu:
Workaround for this in https://github.com/open-sausages/silverstripe-graphql/commit/6dcc005b0178ae689967198dcf27cb5bbbae7d7c

chillu:
Demo of the current issue in `3.0.1`:

graphql.yml

```yml
SilverStripe\Control\Director:
  rules:
    'graphql': '%$SilverStripe\GraphQL\Controller.default'
SilverStripe\GraphQL\Manager:
  schemas:
    default:
      scaffolding:
        types:
          Page:
            fields: '*'
            operations:
                read: true
           SilverStripe\ErrorPage\ErrorPage:
             fields: [ErrorCode]
```

```graphql
query readPages {
  readPages {
    edges {
      node {
       	Title
        ErrorCode
      }
    }
  }
}
```
Response

```json
{
  "data": null,
  "errors": [
    {
      "message": "Cannot query field \"ErrorCode\" on type \"Page\".",
      "locations": [
        {
          "line": 6,
          "column": 9
        }
      ]
    }
  ]
}
```

chillu:
This doesn't work either

graphql.yml

```yml
SilverStripe\Control\Director:
  rules:
    'graphql': '%$SilverStripe\GraphQL\Controller.default'
[...]

PR #226, opened 2019-03-04
Revert "NEW: Refactor OperationScaffolder subclasses to use resolved type name"

unclecheese:
This introduced a number of regressions due to scaffolded queries no longer returning unions.

How we handle inheritance and unions has turned into a bigger discussion that will likely result in a more aggressive, lower-level change. https://github.com/silverstripe/silverstripe-graphql/issues/215

Given the short shelf life of both this change, or the bug that it was trying to fix, the best option here seems to be to revert this and aim to solve it properly.


chillu:
Confirmed that our kitchen sink still works with this in place (asset admin, elemental blocks viewer and history viewer use GraphQL queries)

PR #227, opened 2019-03-06
Added "Allow-Credentials" in static $cors and addCorsHeaders()

syanaputra:
Added "Allow-Credentials" in static $cors and addCorsHeaders() to allow modification in yaml. 
This is to fix issue when accepting request from other ip address that has "withCredentials" set to true.


chillu:
Awesome, thank you - thanks for writing docs as well :)

TheBnl:
Could we get a new tag for this? 🙏

PR #230, opened 2019-05-30
FIX: 'types' endpoint should catch exceptions

kinglozzer:
I’m only really opening this because I keep getting bugged by exceptions logged in Sentry 😅. Unauthenticated requests to `admin/graphql/types` currently trigger a server error (an exception bubbles up from `getManager()`->`applyManagerContext()`->`getRequestUser()`), this changes it to instead return:

```json
{"errors":[{"message":"Authentication required"}]}
```
This is the same approach as the main query handling: https://github.com/silverstripe/silverstripe-graphql/blob/3.1.0-rc1/src/Controller.php#L99-L120.

### DIFF
--- a/src/Extensions/IntrospectionProvider.php
+++ b/src/Extensions/IntrospectionProvider.php
@@ -2,6 +2,8 @@
 
 namespace SilverStripe\GraphQL\Extensions;
 
+use Exception;
+use SilverStripe\Control\Director;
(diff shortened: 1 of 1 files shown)


unclecheese:
@kinglozzer Looks like this can and should be targeted against `2.0`. Do you see a reason why not?

kinglozzer:
Good point, I’ve rebased against `2.0`

Issue #216, opened 2019-02-13
"No array of fields defined" shows even when nestedQueries adds fields


chillu:
This should be a valid definition (scaffolded query returns fields), but it fails:

```yml
SilverStripe\GraphQL\Manager:
  schemas:
    default:
        scaffolding:
            MyModel:
              fields: []
              nestedQueries:
                MyRelationship: true
```

# PR's

- [x] https://github.com/silverstripe/silverstripe-graphql/pull/233

PR #233, opened 2019-07-12
Fix error when only nestedQueries and no fields are assigned

bergice:
See: https://github.com/silverstripe/silverstripe-graphql/issues/216


ScopeyNZ:
It appears there's a test asserting this condition. You'll have to update the test so both are not defined. It'd probably be worth asserting the exception does not occur with either `fields` or `nestedQueries` defined too...

unclecheese:
Have fixed the test. Merge on greeen.

PR #237, opened 2019-08-08
Add dummy request

### LINKED ISSUE #203: Add cache_types_in_filesystem=true breaking flush
Adding SilverStripe\GraphQL\Controller:cache_types_in_filesystem: true to _config.yml break flush with error
> ERROR [Emergency]: Uncaught Exception: There was an error caching the GraphQL types: Request method must be POST or GET

robrev:
Fixes #203

### DIFF
--- a/src/Controller.php
+++ b/src/Controller.php
@@ -404,7 +404,12 @@ class Controller extends BaseController implements Flushable
      */
     public function writeSchemaToFilesystem()
     {
-        $manager = $this->getManager(new HTTPRequest('GET', ''));
+        if (Injector::inst()->has(HTTPRequest::class)) {
+            $request = Injector::inst()->get(HTTPRequest::class);
+        } else {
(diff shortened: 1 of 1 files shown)


src/Controller.php
```diff
@@ -404,7 +404,7 @@ protected function getRequestUser(HTTPRequest $request)
      */
     public function writeSchemaToFilesystem()
     {
-        $manager = $this->getManager();
+        $manager = $this->getManager(new HTTPRequest('GET', ''));
```
robbieaverill: I think it would be better to ask for a request from Injector and fall back to this if not, e.g.:

```php
$request = Injector::inst()->has(HTTPRequest::class) ? Injector::inst()->get(HTTPRequest::class) : new NullHTTPRequest();
```

robrev: Thanks @robbieaverill! Makes sense and works for me 😄

PR #238, opened 2019-08-09
FIX Give nicer error messages that provide more context

ScopeyNZ:
Spent a little bit of time wondering what the issue is before I realised it was permissions :sweat_smile:


ScopeyNZ:
I'm going to close this. Looks like permissions have changed a bit in `3` and it's probably going to look very different to do the same contribution on that branch

PR #249, opened 2020-02-20
BUG Add asset-admin test to travis build

maxime-rainville:
This adds the asset-admin tests to the graphql module. Asset-admin is the primary module that relies on graphql, so this serves as a sort of integration test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,21 +2,37 @@ language: php
 
 dist: trusty
 
+before_install:
+  - if [[ $BEHAT_TEST ]]; then sudo apt-get update && sudo apt-get install chromium-chromedriver; fi
+
 env:
   global:
-    - COMPOSER_ROOT_VERSION=3.x-dev
+    - COMPOSER_ROOT_VERSION=3.1.x-dev
+    - DISPLAY=":99"
+    - XVFBARGS=":99 -ac -screen 0 1024x768x16"
+    - SS_BASE_URL="http://localhost:8080/"
+    - SS_ENVIRONMENT_TYPE="dev"
+    - BEHAT_RECIPE_CMS_VERSION=4.4.x-dev
+    - ASSET_ADMIN_VERSION=1.4.x-dev
 
 matrix:
   fast_finish: true
   include:
-    - php: 7.1.2
-      env: DB=PGSQL PHPUNIT_TEST=1 PHPCS_TEST=1
-    - php: 7.1.2
-      env: DB=MYSQL PHPUNIT_TEST=1
-    - php: 7.1.2
+    - php: 7.1
       env: DB=MYSQL PDO=1 PHPUNIT_COVERAGE_TEST=1
+    - php: 7.2
+      env: DB=MYSQL PHPUNIT_TEST=1
+    - php: 7.3
+      env: DB=PGSQL PHPUNIT_TEST=1 PHPCS_TEST=1
+    - php: 7.3
+      env: DB=MYSQL BEHAT_TEST=1
+    - php: 7.3
+      env: DB=MYSQL ASSETADMIN_TEST=1
 
 before_script:
+# Extra $PATH
+  - export PATH=/usr/lib/chromium-browser/:$PATH
+
 # Init PHP
   - composer self-update || true
   - phpenv rehash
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -26,13 +40,25 @@ before_script:
   - composer validate
   - composer install --prefer-dist
   - composer require --prefer-dist --no-update silverstripe/recipe-core:4.x-dev silverstripe/versioned:1.x-dev silverstripe/assets:1.x-dev --prefer-dist
+  - if [[ $PHPUNIT_TEST ]]; then vendor/bin/phpunit; fi
   - composer update
   - if [[ $DB == PGSQL ]]; then composer require silverstripe/postgresql:2.x-dev --prefer-dist; fi
 
+  - if [[ $ASSETADMIN_TEST ]]; then composer require silverstripe/asset-admin 1.x-dev --prefer-source; fi
```
robbieaverill: Do you want to consider minor version lock stepping in this requirement?

maxime-rainville: Humm ... the other module aren't locked step, but there's a good argument that asset-admin should be. We'll just to remember to increment the version as we merged this up.

maxime-rainville:
I thought about adding test for PHP 7.4 while I was at it but there's a bunch of old API references. Garion already has a task to update all our builds to be 7.4 compatible, so I'll leave him this devious pleasure.

.travis.yml
```diff
@@ -2,37 +2,61 @@ language: php
...
   - phpenv config-rm xdebug.ini
 
 # Install composer dependencies
   - composer validate
-  - composer install --prefer-dist
-  - composer require --prefer-dist --no-update silverstripe/recipe-core:4.x-dev silverstripe/versioned:1.x-dev silverstripe/assets:1.x-dev --prefer-dist
+  - if [[ ! $BEHAT_TEST ]]; then composer require --prefer-dist --no-update silverstripe/recipe-core:4.x-dev silverstripe/versioned:1.x-dev silverstripe/assets:1.x-dev --prefer-dist ; fi
+  - if [[ $DB == PGSQL ]]; then composer require silverstripe/postgresql:2.x-dev --prefer-dist --no-update; fi
+  - if [[ $ASSETADMIN_TEST ]]; then composer require silverstripe/asset-admin 1.4.x-dev --prefer-source --no-update ; fi
+  - if [[ $BEHAT_TEST ]]; then composer require silverstripe/recipe-testing:^1 silverstripe/recipe-cms 4.4.x-dev --prefer-source --no-update ; fi
```
emteknetnz: put this line directly above (or below) the ! $BEHAT_TEST so it's more like a regular if else statement

.travis.yml
```diff
@@ -2,37 +2,61 @@ language: php
 
 dist: trusty
 
+before_install:
+  - sudo apt-get update
+  - sudo apt-get install chromium-chromedriver
```
emteknetnz: should wrap this in if [[ $BEHAT_TEST ]] ?

.travis.yml
emteknetnz: how come this is using recipe-cms 4.4.x-dev?  this should be 4.5.x-dev?  also would prefer to see this as a BEHAT_RECIPE_CMS_VERSION and put it up the top, otherwise people will not realise that it needs to be updated over time

emteknetnz: why does this and ASSETADMIN_TEST needs --prefer-source?

emteknetnz: how come this is using recipe-cms 4.4.x-dev? this should be 4.5.x-dev? also would prefer to see this as a BEHAT_RECIPE_CMS_VERSION and put it up the top, otherwise people will not realise that it needs to be updated over time

maxime-rainville: I can't really do that because I need to get the dependency installed first. The `--no-update` flag delays the actual update until you manually call update.

.travis.yml
maxime-rainville: That's how the other builds with behat test do it. My guess is that the env variables aren't available at this stage, but I can have a go to see if it they are.

.travis.yml
maxime-rainville: `--prefer-dist` doesn't install the test files normally. 

All of this `--prefer-source/prefer-dist` non-sense might not be making any difference in practice because we're targeting a branch which always install from source any way ... I think.

maxime-rainville: recipe-cms 4.4 shipped with graphql 3.1, which this PR is targeting.

recipe-cms 4.5 shipped with graphql 3.2. We'll need to merge this up and tweak targeted version.

Good suggestion on the BEHAT_RECIPE_CMS_VERSION. I'll do that.

emteknetnz: ok can you put the [[ ! $BEHAT_TEST ]] down a couple of lines then?

.travis.yml
maxime-rainville: Looks like env variable are available there. I've tweak the build to only install `chromium-chromedriver` on BEHAT builds.

PR #253, opened 2020-03-13
NEW: Allow instance override of CORS config

unclecheese:
Now that we support multiple schemas, it doesn't make sense that the CORS settings should be global. This allows you to merge in custom settings per instance and is BC with the current static config.

```
  SilverStripe\GraphQL\Controller.myController:
    properties:
      corsConfig:
        Enabled: false
```

### DIFF
--- a/src/Controller.php
+++ b/src/Controller.php
@@ -251,7 +251,7 @@ class Controller extends BaseController implements Flushable
     /**
      * @return array
      */
-    public function getCorsConfig()
+    public function getCorsConfig(): array
     {
         return $this->corsConfig;
     }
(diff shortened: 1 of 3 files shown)


src/Controller.php
```diff
@@ -242,6 +248,36 @@ public function addCorsHeaders(HTTPRequest $request, HTTPResponse $response)
...
+    {
+        return $this->corsConfig;
+    }
+
+    /**
+     * @return array
+     */
+    public function getMergedCorsConfig(): array
+    {
+        $defaults = Config::inst()->get(static::class, 'cors');
```
dhensby, line 264: Can we not just use `$this->config()->cors`? that might be more "readable"

unclecheese: Not my preference. IDE's hate it.

PR #254, opened 2020-03-16
ENHANCEMENT: Boot typenames outside graphql request

unclecheese:
See: https://github.com/dnadesign/silverstripe-elemental/issues/751


tests/Scaffolding/StaticSchemaTest.php
```diff
@@ -15,6 +15,12 @@
 
 class StaticSchemaTest extends SapphireTest
 {
+    protected function setUp()
```
dhensby, line 18:
```suggestion
    protected function setUp(): void
```

This would be required for ss 5 compat

dhensby:
hmmm - there's a behat failure, not sure why. any thoughts @unclecheese

unclecheese:
This happened last time. I think I just reran it.

PR #255, opened 2020-03-29
Upgrade to PHPUnit 8

dhensby:
Depends on https://github.com/silverstripe/silverstripe-framework/pull/9443


dhensby:
> It looks like the GraphQL API changes you've made maybe should've stayed the way they were (see Travis failures)

Looks like I had v0.8 installed locally 🤦‍♂

PR #261, opened 2020-05-28
BUG Require frameworktest for behat test

maxime-rainville:
The asset-admin behat test now requires frameworktest to work because of https://github.com/silverstripe/silverstripe-asset-admin/pull/1038

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
 
 env:
   global:
-    - COMPOSER_ROOT_VERSION=3.6.x-dev
+    - COMPOSER_ROOT_VERSION=3.3.x-dev
     - SS_BASE_URL="http://localhost:8080/"
     - SS_ENVIRONMENT_TYPE="dev"
     - RECIPE_CMS_VERSION=4.6.x-dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ before_script:
   - if [[ (! $BEHAT_TEST) && (! $ASSETADMIN_TEST) ]]; then composer require --prefer-dist --no-update silverstripe/recipe-core:4.x-dev silverstripe/versioned:1.x-dev silverstripe/assets:1.x-dev --prefer-dist ; fi
   - if [[ $DB == PGSQL ]]; then composer require silverstripe/postgresql:2.x-dev --prefer-dist --no-update; fi
   - if [[ $ASSETADMIN_TEST ]]; then composer require silverstripe/recipe-cms $RECIPE_CMS_VERSION --prefer-source --no-update ; fi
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -19,10 +19,10 @@ install:
 
 env:
   global:
-    - COMPOSER_ROOT_VERSION=3.x-dev
+    - COMPOSER_ROOT_VERSION=3.6.x-dev
```
emteknetnz, line 22:
```suggestion
    - COMPOSER_ROOT_VERSION=3.3.x-dev
```

.travis.yml
```diff
@@ -51,10 +51,10 @@ before_script:
 
 # Install composer dependencies
   - composer validate
-  - if [[ (! $BEHAT_TEST) && (! $ASSETADMIN_TEST) ]]; then composer require --prefer-dist --no-update silverstripe/recipe-core:4.x-dev silverstripe/versioned:1.x-dev silverstripe/assets:1.x-dev --prefer-dist ; fi
+  - if [[ (! $BEHAT_TEST) && (! $ASSETADMIN_TEST) ]]; then composer require --prefer-dist --no-update silverstripe/recipe-core:4.6.x-dev silverstripe/versioned:1.6.x-dev silverstripe/assets:1.6.x-dev --prefer-dist ; fi
   - if [[ $DB == PGSQL ]]; then composer require silverstripe/postgresql:2.x-dev --prefer-dist --no-update; fi
   - if [[ $ASSETADMIN_TEST ]]; then composer require silverstripe/recipe-cms $RECIPE_CMS_VERSION --prefer-source --no-update ; fi
-  - if [[ $BEHAT_TEST ]]; then composer require silverstripe/recipe-testing:^1 silverstripe/recipe-cms $RECIPE_CMS_VERSION --prefer-source --no-update ; fi
+  - if [[ $BEHAT_TEST ]]; then composer require silverstripe/recipe-testing:^1 silverstripe/recipe-cms $RECIPE_CMS_VERSION silverstripe/frameworktest:^0.1.0 --prefer-source --no-update ; fi
```
emteknetnz, line 57: silverstripe/frameworktest:^0.1.0

^ is this normal elsewhere? or do we normally use `dev-master`?

maxime-rainville: Remember, I wanted to use silverstripe/frameworktest:dev-master originally when I was updating the the asset-admin PR for the in-modal editing. But you made the case that if someone updated frameworktest, it might accidentally break the build. So we decided to the a release of frameworktest and target that instead.

I'm extending this principal here.

.travis.yml
maxime-rainville: Good catch.

PR #263, opened 2020-06-06
Make silverstripe-graphql independent of the presence of Versioned mo…

alessandromarotta:
silverstripe-graphql index() function calls a Versioned methods and this makes it impossible to have the silverstripe-graphql module installed without the Versioned module.

I think that it could be usefull to install it without the Versioned module.

Please, could you also introduce in the index() function the verification of the presence of the Versioned class in order to install the silverstripe-graphql module without the Versioned module?

Like in the file commit, for example.
Thank you.

### DIFF
--- a/src/Controller.php
+++ b/src/Controller.php
@@ -103,10 +103,13 @@ class Controller extends BaseController implements Flushable
      */
     public function index(HTTPRequest $request)
     {
-        $stage = $request->param('Stage');
-        if ($stage && in_array($stage, [Versioned::DRAFT, Versioned::LIVE])) {
-            Versioned::set_stage($stage);
+        if (class_exists(Versioned::class)) {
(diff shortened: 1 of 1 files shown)


emteknetnz:
Hi thanks for the PR.

This is a pretty interesting one because the versioned module kind of should be in composer.json given it's reference in the codebase, though it's not.  I guess there's been possibly very few in any instances of people actually using this module without first having installed another silverstripe module that does require versioned.

Another solution here would be to add versioned to composer.json, though that would negate any use cases of using this module without versioned

Are you working on a particular project right where you need to use this module without versioned?

alessandromarotta:
Hi, no for now, but the question should be "why do I have to install the Versioned module if I just want to install the GraphQl module?"

Issue #265, opened 2020-06-15
Search Params/DataObjectQueryFilter Bugs 


HARVS1789UK:
*Affected Version*: 3.2.3

# Issues
Whilst following the documentation for exposing existing DataObjects through procedural code, I have come across the following issues/errors, which I intend to submit a PR to correct, this issue serves as some documentation of the issues and a place to discuss the accuracy of me reporting them as bugs (as I am happy to be corrected and shown how to implement the module correctly in it's current state).

All 3 issues relate to the below section of the docs (or something closely related to them) seen here - https://github.com/silverstripe/silverstripe-graphql/tree/3.2#adding-search-params-read-operations-only

A key snippet for examples to follow:

```
$scaffolder->type(MyDataObject::class)
  ->operation(SchemaScaffolder::READ)
    ->queryFilter()
      ->addDefaultFields('MyField')
      ->addFieldFilter('MyInt', 'gt')
      ->addFieldFilter('MyInt', 'gte')
    ->end()
  ->end();
```

## 1 - Documentation

The documentation refences the use of the following methods with regards to adding search parameters via the `Filter` GraphQL input:

`addFilteredField('Title', 'contains')`
`addDefaultFields('MyField')`

Neither of these seem to exist and I think they are typos? The correct methods seems to be 

`addFieldFilterByIdentifier('Title`,`contains')`
`addDefaultFilters(MyField)`

The docs also show the use of `addFieldFilter('MyInt', 'gt')`, this method does exist, but expects something implementing `FieldFilterInterface` as the second parameter, not a string, so again in this context, I think the intended method is `addFieldFilterByIdentifier('MyInt`,`gt')`

see:
- https://github.com/silverstripe/silverstripe-graphql/tree/3.2#adding-search-params
- https://github.com/silverstripe/silverstripe-graphql/tree/3.2#adding-search-params-read-operations-only

## 2 - DataObjectQueryFilter and the Chainable trait

`DataObjectQueryFilter` implements the `Chainable` trait and is shown in the docs to be chain-able against the `SchemaScaffolder` (see main snippet above). However, this gives me errors claiming `->end()` cannot be called on `null` and I believe this to be due to `Read::__construct()` not calling `->setChainableParent()` on the `DataObjectQueryFilter` when it creates it?

See:

- https://github.com/silverstripe/silverstripe-graphql/blob/3.2/src/Scaffolding/Scaffolders/CRUD/Read.php#L41
[...]

PR #267, opened 2020-08-11
FIX: correctly resolve field names of relations

mleutenegger:
Fixes an error where fields of relations could not be filtered against.

Resolves 1) and 3) in #265

### DIFF
--- a/src/QueryFilter/DataObjectQueryFilter.php
+++ b/src/QueryFilter/DataObjectQueryFilter.php
@@ -420,7 +420,6 @@ class DataObjectQueryFilter implements ConfigurationApplier
         if (stristr($field, self::SEPARATOR) !== false) {
             $relationNames = explode(self::SEPARATOR, $field);
             $relationField = array_pop($relationNames);
-            
             // reverse array so we can use the faster array_pop
             $relationNames = array_reverse($relationNames);
             // initialize current class
--- a/src/QueryFilter/DataObjectQueryFilter.php
+++ b/src/QueryFilter/DataObjectQueryFilter.php
@@ -424,7 +424,7 @@ class DataObjectQueryFilter implements ConfigurationApplier
             // reverse array so we can use the faster array_pop
             $relationNames = array_reverse($relationNames);
             // initialize current class
-            $class = $this->getDataObjectInstance()->getRelationClass($relationName);
+            $class = get_class($this->getDataObjectInstance());
             do {
                 $relationName = array_pop($relationNames);
                 $lastClass = $class;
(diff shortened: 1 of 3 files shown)


emteknetnz:
@mleutenegger Thanks for submitting this PR.  Would it be possible to add a unit test or two to this PR to shows that the bug has been fixed.  Cheers

mleutenegger:
👍  I'll try and add some

mleutenegger:
@emteknetnz I have added tests for the resolving of relations. I stumbled upon another bug in the `getDBField()` function which prevented the definition of deeper relations (>1). I have fixed that one too and added two assertions for this.

Please let me know what you think!

src/QueryFilter/DataObjectQueryFilter.php
```diff
@@ -419,13 +418,23 @@ protected function getDBField($field)
     {
         $dbField = null;
         if (stristr($field, self::SEPARATOR) !== false) {
-            list ($relationName, $relationField) = explode(self::SEPARATOR, $field);
+            $relationNames = explode(self::SEPARATOR, $field);
+            $relationField = array_pop($relationNames);
+
+            $relationName = array_shift($relationNames);
             $class = $this->getDataObjectInstance()->getRelationClass($relationName);
+            while ($class && count($relationNames) > 0) {
```
emteknetnz, line 426: Just wondering if it's possible to rewrite this so that it doesn't have to loop through all the relation names?  Is it possible to reverse this so that it uses array_pop instead?

emteknetnz (changes requested):
Seems good, I've run this through a [cross module integration test](https://travis-ci.org/github/creative-commoners/silverstripe-installer/builds/733920953) which it passed, just wondering if we could make one bit on it a little more efficient. Cheers.

src/QueryFilter/DataObjectQueryFilter.php
dhensby: If it is done this way, then it should use `&& !empty($relationNames)` rather than a count

mleutenegger: I can reverse the array and use `array_pop` instead combined with a `break` statement (or a `do...while` loop). However, the changes would be purely cosmetic, as it seems to me like there is no other way except for traversing the entire relation tree to find the actual final class and field type.

@dhensby Thanks for correcting me on the `count`!

mleutenegger:
@emteknetnz I have changed the condition to use `empty()` and used a do ... while loop with an `array_reverse`. I have to correct my statement from earlier, `array_pop` is faster than `array_shift` 😅 

However, I cannot fix the error with composer 2 on my end...

mleutenegger:
@emteknetnz It seems like the composer issue is fixed, could you re-run the tests for me?

emteknetnz:
@mleutenegger sorry been quite busy at my end, realistically I won't have a chance to look at this for a few weeks, sorry about the delay

mleutenegger:
@emteknetnz @dhensby any news on this?

PR #310, opened 2020-10-24
NEW: field formatting API (GraphQL 4 foward compat)

unclecheese:
We need backward compat for all the core modules, so they can run on either v3 or v4. It's very difficult to do this at run time because the graphql queries are bound to components at build time, so having two separate queries is both a technical challenge and also hard to maintain.

The easier solution is to allow the graphql 3 queries to look like graphql 4 queries, and the main issue there is case sensitivity.

This pull request allows the optional provision of a field formatter function and field accessor service to abstract this away. By default, you still get the naive implementation of DBField == GraphQL Field, but in admin, we'll want to take advantage of a lowerCamelCase transform.

Todo:
* Unit testing
* Documentation (v3 only)

### DIFF
--- a/src/FieldAccessorInterface.php
+++ b/src/FieldAccessorInterface.php
@@ -1,0 +1,35 @@
+<?php
+
+
+namespace SilverStripe\GraphQL;
+
+use SilverStripe\View\ViewableData;
+
+interface FieldAccessorInterface
+{
+    /**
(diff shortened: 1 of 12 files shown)


unclecheese:
Moved to: https://github.com/silverstripe/silverstripe-graphql/pull/313

Issue #282, opened 2020-09-08
GraphQL 4: Upgrade core modules


unclecheese:
## To implement

* [ ] snapshot-admin - https://github.com/silverstripe/silverstripe-versioned-snapshots/issues/40

## To merge

* [x] [graphql](https://github.com/silverstripe/silverstripe-graphql/pull/313)
* [x] [framework](https://github.com/silverstripe/silverstripe-framework/pull/9762) - changelog entry
* [x] [admin](https://github.com/silverstripe/silverstripe-admin/pull/1148)
* [x] [CMS](https://github.com/silverstripe/silverstripe-cms/pull/2607)
* [x] [asset-admin](https://github.com/silverstripe/silverstripe-asset-admin/pull/1154)
* [x] [versioned](https://github.com/silverstripe/silverstripe-versioned/pull/308)
* [x] [versioned-admin](https://github.com/silverstripe/silverstripe-versioned-admin/pull/184)
* [x] [elemental](https://github.com/dnadesign/silverstripe-elemental/pull/837)

## Backward compatibility

These upgrades were done without attention to backward compatibility. This shouldn't have happened. The groundwork for BC had been established (graphql-legacy.yml, and writing `data-graphql-legacy` to the document for runtime client side compat). We'll need to revisit the upgrades for asset-admin and elemental.

Assuming changing the admin graphql schema in GraphQL 3 is not an API change, this should be pretty straightforward.

### Tasks:

- [x] Rewrite GraphQL 3 queries to have the same args, fields, etc as GraphQL 4 does / will. (CaseInsensitiveFieldAccessor will help)
- [x] Implement a solution for including/excluding unit tests based on GraphQL version
- [x] Backport GraphQL 3 unit tests
- [x] Update travis config to test GraphQL 3 and GraphQL 4 (e.g. put all the monkey patching behind an env var condition)
- [x] Regression test with GraphQL v3
- [x] Regression test with GraphQL v4

### Setup

See `.travis.yml` files for new required forks.

### Build status (with backward compat)

✅  silverstripe/admin https://travis-ci.com/github/silverstripe/silverstripe-admin

✅  silverstripe/asset-admin https://travis-ci.org/github/unclecheese/silverstripe-asset-admin/builds/741754720

✅  silverstripe/cms https://travis-ci.org/github/unclecheese/silverstripe-cms

✅  silverstripe/versioned https://travis-ci.org/github/unclecheese/silverstripe-versioned/builds/741841567

[...]

PR #313, opened 2020-11-02
NEW: Field formatting API, forward compatibility

unclecheese:
Moved from: https://github.com/silverstripe/silverstripe-graphql/pull/310

We need backward compat for all the core modules, so they can run on either v3 or v4. It's very difficult to do this at run time because the graphql queries are bound to components at build time, so having two separate queries is both a technical challenge and also hard to maintain.

The easier solution is to allow the graphql 3 queries to look like graphql 4 queries, and the main issue there is case sensitivity.

This pull request allows the optional provision of a field formatter function and field accessor service to abstract this away. By default, you still get the naive implementation of DBField == GraphQL Field, but in admin, we'll want to take advantage of a lowerCamelCase transform.

Todo:
* Unit testing
* Documentation (v3 only)

### DIFF
--- a/src/Pagination/Connection.php
+++ b/src/Pagination/Connection.php
@@ -291,6 +291,10 @@ class Connection implements OperationResolver
             'edges' => [
                 'type' => Type::listOf($this->getEdgeType()),
                 'description' => 'Collection of records'
+            ],
+            'nodes' => [
+                'type' => Type::listOf($this->getConnectionType()),
+                'description' => 'The node at the end of the collections edge',
             ]
         ];
     }
@@ -410,11 +414,12 @@ class Connection implements OperationResolver
 
         return [
             'edges'    => $list,
+            'nodes'    => $list,
             'pageInfo' => [
                 'totalCount'      => $count,
                 'hasNextPage'     => $nextPage,
                 'hasPreviousPage' => $previousPage
-            ]
+            ],
         ];
     }
 
--- a/src/Scaffolding/Scaffolders/DataObjectScaffolder.php
+++ b/src/Scaffolding/Scaffolders/DataObjectScaffolder.php
@@ -341,7 +341,7 @@ class DataObjectScaffolder implements ManagerMutatorInterface, ScaffolderInterfa
 
         if (!$queryScaffolder) {
             // If no scaffolder if provided, try to infer the type by resolving the field
-            $result = $this->getDataObjectInstance()->obj($fieldName);
+            $result = StaticSchema::inst()->accessField($this->getDataObjectInstance(), $fieldName);
 
             if (!$result instanceof DataList && !$result instanceof ArrayList) {
                 throw new InvalidArgumentException(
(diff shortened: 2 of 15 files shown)


chillu (changes requested):
Looking good! Tested this with all the branches mentioned in https://github.com/silverstripe/silverstripe-graphql/issues/282. 

Talked with Aaron, we've agreed to mark the new API surface as `@internal` and encourage usage of GraphQL v4 instead.

We'll want to merge this close to the time as all the other PRs, since it would break common dev usage such as `silverstripe/cms:4.x-dev` + `silverstripe/graphql:3.x-dev` usage otherwise.

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
```diff
@@ -63,6 +63,12 @@ class DataObjectScaffolder implements ManagerMutatorInterface, ScaffolderInterfa
      */
     protected $nestedQueries = [];
 
+    protected function setUp(): void
+    {
+        parent::setUp();
```
chillu, line 68: This class doesn't have a parent?

src/Pagination/Connection.php
```diff
@@ -291,6 +291,10 @@ public function fields()
             'edges' => [
                 'type' => Type::listOf($this->getEdgeType()),
                 'description' => 'Collection of records'
+            ],
+            'nodes' => [
```
chillu, line 295: This breaks `admin/graphql/types/` when not running any of the associated PR branches.

```
GraphQL\Utils\Utils::invariant(, Schema must contain unique named types but contains multiple types named "Result" (see http://webonyx.github.io/graphql-php/type-system/#type-registry).)
TypeInfo.php:109
GraphQL\Utils\TypeInfo::extractTypes(Result, Array)
TypeInfo.php:94
GraphQL\Utils\TypeInfo::extractTypes([Result], Array)
TypeInfo.php:138
GraphQL\Utils\TypeInfo::extractTypes([Result], Array)
TypeInfo.php:138
GraphQL\Utils\TypeInfo::extractTypes(readFilesConnection, Array)
Schema.php:228
```

chillu: `ReadFileQueryCreator->createConnection()` requires patching through it's own PR, which means this PR is coupled to the asset-admin one

src/Scaffolding/Scaffolders/DataObjectScaffolder.php
unclecheese: Ooof, looks like a copy-and-paste that was meant for `DataObjectScaffolderTest`. IDE fail.

src/Pagination/Connection.php
chillu: I don't think it's worth the effort to fix this just to avoid an hour of broken state in a pre-release branch while we merge everything else.

chillu (approved):

PR #325, opened 2020-11-13
NEW Validate schema keys

chillu:
Prevents "forgetting" a level. In this example, I've forgotten the "models" key, which means the schema build will just finish without any output or error.

  schemas:
    default:
      SilverStripe\CMS\Model\SiteTree:
        fields:
          '*': true

The downside of this approach is that we can't easily extend the schema with new keys
through modules building on silverstripe/graphql, but I'd say we cross that bridge when the need arises.


unclecheese:
Outdated. Added to open refactoring PR here: https://github.com/silverstripe/silverstripe-graphql/pull/352/commits/2a2940694469973accf9384c4d8e5709f23cd148

PR #327, opened 2020-11-19
NEW: Stateless plugins, plugin registry as simple list

unclecheese:
This fixes a number of errors that could come up when doing multiple schema builds where plugin state was polluted by preceding schemas. Plugins are no longer singletons, and are instead instantiated each time they are used. State that should be persisted for the duration of the build should be in static members, namespaced by the current schema.

## Interdependent PRs

- [ ] https://github.com/silverstripe/silverstripe-framework/pull/9777
- [ ] https://github.com/silverstripe/silverstripe-cms/pull/2615
- [ ] https://github.com/silverstripe/silverstripe-versioned/pull/314

### DIFF
--- a/_config/plugins.yml
+++ b/_config/plugins.yml
@@ -4,12 +4,12 @@ Name: graphql-plugins
 SilverStripe\Core\Injector\Injector:
   SilverStripe\GraphQL\Schema\Registry\PluginRegistry:
     constructor:
-      paginator: '%$SilverStripe\GraphQL\Schema\Plugin\PaginationPlugin'
-      dataobjectPaginator: '%$SilverStripe\GraphQL\Schema\DataObject\Plugin\Paginator'
-      dataobjectQueryFilter: '%$SilverStripe\GraphQL\Schema\DataObject\Plugin\QueryFilter\QueryFilter'
-      dataobjectQuerySort: '%$SilverStripe\GraphQL\Schema\DataObject\Plugin\QuerySort'
-      dataobjectInheritance: '%$SilverStripe\GraphQL\Schema\DataObject\Plugin\Inheritance'
-      canViewPermission: '%$SilverStripe\GraphQL\Schema\DataObject\Plugin\CanViewPermission'
-      firstResult: '%$SilverStripe\GraphQL\Schema\DataObject\Plugin\FirstResult'
-      inheritedPlugins: '%$SilverStripe\GraphQL\Schema\DataObject\Plugin\InheritedPlugins'
-      sorter: '%$SilverStripe\GraphQL\Schema\Plugin\SortPlugin'
+      - 'SilverStripe\GraphQL\Schema\Plugin\PaginationPlugin'
+      - 'SilverStripe\GraphQL\Schema\DataObject\Plugin\Paginator'
+      - 'SilverStripe\GraphQL\Schema\DataObject\Plugin\QueryFilter\QueryFilter'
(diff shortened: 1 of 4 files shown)


src/Schema/Registry/PluginRegistry.php
```diff
@@ -21,25 +22,26 @@ class PluginRegistry
...
 
     /**
-     * @param PluginInterface[] ...$plugins
+     * @param array ...$plugins
      * @throws SchemaBuilderException
      */
     public function __construct(...$plugins)
     {
         foreach ($plugins as $plugin) {
+            $inst = Injector::inst()->get($plugin);
```
chillu, line 31: Shouldn't this be `create()`? The idea was NOT to create singletons here, right? It's using `create()` further down in `getPluginByID()`

_config/plugins.yml
```diff
@@ -4,12 +4,12 @@ Name: graphql-plugins
...
-      paginator: '%$SilverStripe\GraphQL\Schema\Plugin\PaginationPlugin'
-      dataobjectPaginator: '%$SilverStripe\GraphQL\Schema\DataObject\Plugin\Paginator'
-      dataobjectQueryFilter: '%$SilverStripe\GraphQL\Schema\DataObject\Plugin\QueryFilter\QueryFilter'
-      dataobjectQuerySort: '%$SilverStripe\GraphQL\Schema\DataObject\Plugin\QuerySort'
-      dataobjectInheritance: '%$SilverStripe\GraphQL\Schema\DataObject\Plugin\Inheritance'
-      canViewPermission: '%$SilverStripe\GraphQL\Schema\DataObject\Plugin\CanViewPermission'
-      firstResult: '%$SilverStripe\GraphQL\Schema\DataObject\Plugin\FirstResult'
-      inheritedPlugins: '%$SilverStripe\GraphQL\Schema\DataObject\Plugin\InheritedPlugins'
-      sorter: '%$SilverStripe\GraphQL\Schema\Plugin\SortPlugin'
+      - 'SilverStripe\GraphQL\Schema\Plugin\PaginationPlugin'
```
chillu, line 7: We talked about using identifiers for plugins, rather than `Plugin->getIdentifier()`. It'd make it a bit more obvious how the YAML `plugins` config relates to actual class implementations. A nice side effect of this would be that users could *unregister* plugins from the registry, e.g. for security or performance reasons. And you could replace implementations by YAML config on `SilverStripe\GraphQL\Schema\Registry\PluginRegistry`, rather than the less obvious `Injector` config (dependency injection).

Admittedly, none of those arguments are particularly strong, so I'm happy to keep it as-is if you think that's the best way forward.

chillu (changes requested):

_config/plugins.yml
unclecheese: Looked into this, and it's a bit complicated. There are just too many areas where a plugin needs to know its identifier internally, and discovering that from the config layer just seems really awkward.

The reason I got rid of the keys was because it's just confusing. We have too many of these arbitrary keys that don't mean anything. I was having a hard time envisioning a use case for cancelling out a plugin from the registry all together. 

What about an `excludePlugins` section?

chillu: OK not worth it, we can always introduce `excludePlugins` later.

src/Schema/Registry/PluginRegistry.php
chillu: It's a bit of a code smell, but since you're only using `getIdentifier()` on the instance it doesn't matter too much.

chillu (approved):

PR #333, opened 2020-12-06
NEW: Add test coverage to v4

unclecheese:

PR #335, opened 2020-12-14
NEW: Add test coverage to v4

chillu:
Replaces https://github.com/silverstripe/silverstripe-graphql/pull/333 which got auto-closed because the `4` branch was deleted (it's `master` now)

### DIFF
--- a/tests/Schema/IntegrationTest.php
+++ b/tests/Schema/IntegrationTest.php
@@ -1,0 +1,96 @@
+<?php
+
+
+namespace SilverStripe\GraphQL\Tests\Schema;
+
+
+use SilverStripe\Dev\SapphireTest;
+use SilverStripe\GraphQL\QueryHandler\QueryHandler;
+use SilverStripe\GraphQL\Schema\Schema;
+use SilverStripe\GraphQL\Schema\Storage\CodeGenerationStore;
+use SilverStripe\GraphQL\Schema\Storage\CodeGenerationStoreCreator;
+use Symfony\Component\Filesystem\Filesystem;
+use GraphQL\Type\Schema as GraphQLSchema;
+use Exception;
+
+class IntegrationTest extends SapphireTest
+{
+    protected function tearDownOnce()
+    {
+        parent::tearDownOnce();
+        $this->clean();
+    }
+
+    public function testSimpleType()
+    {
+        $schema = $this->createSchema('_test2');
+        $query = <<<GRAPHQL
+query {
+    readMyTypes {
+        field1
+        field2
+    }
+}
+GRAPHQL;
+
+        $result = $this->querySchema($schema, $query);
+        $this->assertSuccess($result);
+        $records = $result['data'];
+        $this->assertCount(2, $records);
+        $this->assertResults([
+            ['field1' => 'foo', 'field2' => 2],
+            ['field1' => 'bar', 'field2' => 3],
+        ], $records);
+
+    }
+
+    private function createSchema(string $configDir, array $extraConfig = []): GraphQLSchema
+    {
+        $this->clean();
+        $schema = new Schema('test');
+        /* @var CodeGenerationStore $store */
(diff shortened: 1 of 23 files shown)


tests/Schema/IntegrationTest.php
```diff
@@ -0,0 +1,96 @@
...
+    private function createSchema(string $configDir, array $extraConfig = []): GraphQLSchema
+    {
+        $this->clean();
+        $schema = new Schema('test');
+        /* @var CodeGenerationStore $store */
+        $store = (new CodeGenerationStoreCreator())->createStore('test');
+        $store->setRootDir(__DIR__);
+        $schema->setStore($store);
+
+        $schema->applyConfig([
```
chillu, line 56: I don't see how this works - it fails validation right now. Schema calls `bootConfig` on `__construct()`, which resolves all the 'src' references.

```
vendor/bin/phpunit --filter testSimpleType vendor/silverstripe/graphql/tests/Schema/IntegrationTest.php 
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

E                                                                   1 / 1 (100%)

Time: 5.59 seconds, Memory: 22.00MB

There was 1 error:

1) SilverStripe\GraphQL\Tests\Schema\IntegrationTest::testSimpleType
SilverStripe\GraphQL\Schema\Exception\SchemaBuilderException: Your schema must contain at least one type and at least one query

/Users/ingo/Sites/4/vendor/silverstripe/graphql/src/Schema/Schema.php:1145
/Users/ingo/Sites/4/vendor/silverstripe/graphql/src/Schema/Schema.php:684
/Users/ingo/Sites/4/vendor/silverstripe/graphql/src/Schema/Schema.php:592
/Users/ingo/Sites/4/vendor/silverstripe/graphql/tests/Schema/IntegrationTest.php:60
/Users/ingo/Sites/4/vendor/silverstripe/graphql/tests/Schema/IntegrationTest.php:26
/Users/ingo/Sites/4/vendor/sminnee/phpunit/phpunit:52
```

chillu:
Great start. Needs a rebase which I can't push. Our baseline for code coverage on the `3` branch is 86% (https://app.codecov.io/gh/silverstripe/silverstripe-graphql/branch/3). After a local `phpdbg` run we're at 20% now. The purpose of adding test coverage is to avoid regressions which could impact early adoption, so we should focus on a) the most common usage paths and b) the most security sensitive paths. Getting coverage back up to >80% will be part of stabilisation, which in turn is dependant on proven demand for this module over the next months.

We can probably port over at least some tests where the structure/intent stays the same, e.g. `ControllerTest` and `CanViewPermissionCheckerTest`. Others might only provide a starting point but need heavy rewrites, e.g. `DataObjectQueryFilterTest`.

tests/Schema/IntegrationTest.php
unclecheese: Yup, that's right. Could either:

* Detect the `src/` key in the config and take a different path
* Use another API `loadConfigFromSource` (which I think is public?)

The former is good because it creates a consistent dev experience. (`src/` is in the yaml config, why can't it be in an array config?), but it makes the code more complex internally.

The latter would just keep things tidier and require the dev to know that the initial bootConfig() is special.

unclecheese (approved):

unclecheese:
Moved: https://github.com/silverstripe/silverstripe-graphql/pull/342

tests/Schema/IntegrationTest.php
chillu: Lead me down a bit of a rabbit hole, but I have a better understanding of the `Schema` class now. I'm proposing that we get rid of the dual operation of `bootConfig()` and `loadConfig()` which implicitly fixes this issue: https://github.com/silverstripe/silverstripe-graphql/pull/342/commits/58fbcef6914d8cf17238bb3ed5d757c3ae3b1cc0

PR #350, opened 2021-01-17
MINOR: First batch of integration tests

unclecheese:


unclecheese:
Closed because it was using an old `4` PR branch and that's super confusing.

Issue #351, opened 2021-01-18
GraphQL 4: Controller constructor is too greedy


unclecheese:
Seems like the `Controller` class requires parameters that aren't necessarily depedencies.

* schemaKey (string): Mostly done for the ergonomics of the config layer. It's annoying that you can't pass a first-class `Schema` instance. But for that matter, the only method that _requires_ a schema is `index()`. (introspectTypes is a weird one.. that [probably be done away with](https://github.com/silverstripe/silverstripe-graphql/issues/270)

* `schemaConfig` (SchemaConfig): Not sure what the use case is here, since `Schema` can be composed with a `SchemaConfig` instance anyway.

Recommend:

* Only have optional `schemaKey` param, call new public `setSchemaByKey` method
*  Throw on index when no schema set
* If `Schema` has no `SchemaContext` passed to it, automatically try to set it from the store. It's getting too confusing to know the difference between `$schema->getContext()` and `$schema->getStore()->getContext()`.

chillu:
> the only method that requires a schema is index()

That's the *main purpose* of this class though, so I'd say unless we infer this from the request context, declaring this requirement in the constructor is correct. It makes controllers harder to test that it takes charge of creating its own `Schema` instance.

> If Schema has no SchemaContext passed to it, automatically try to set it from the store.

`CodeGenerationStore` doesn't have a reference to a `SchemaStore`. Do you mean `Schema`? In which case I agree that we should remove the default creation of a context from `Controller::__construct()`, because it already happens in `Schema::__construct()`.

unclecheese:
This was cleaned up in 4.0.0-alpha1 https://github.com/silverstripe/silverstripe-graphql/releases/tag/4.0.0-alpha1

Issue #349, opened 2021-01-14
GraphQL v4: Child pages not able to be viewed in DRAFT 


blueo:
We're querying child pages we're not able to view pages that are in DRAFT via graphql eg this query:

```
query ReadTilePage($id: ID!, $stage: VersionedQueryMode!) {
  readOneSiteTree(filter: { id: { eq: $id } }, versioning: { mode: $stage }) {
    id
    metaDescription
    title
    content
    children {
      id
      title
      content
      link
      _extend {
        cardPage {
          icon
        }
      }
    }
    __typename
  }
}
```

with the stage set to DRAFT we're only getting the published pages. 

Its a fairly simple schema to reproduce it:

```
Page:
  fields:
    "*": true
  operations:
    read: true
 ```

chillu:
@unclecheese I think you worked through this with Bernie separately?

blueo:
@chillu we had a chat but it is still unresolved FYI

unclecheese:
On my list to do next

unclecheese:
Resolved in one of the commits to a forthcoming PR: https://github.com/silverstripe/silverstripe-versioned/pull/319/commits/11615a3444174b4aaadf1f8f1ee28144c6ee7efa

unclecheese:
Merged, in 4.0.0-alpha 1 https://github.com/silverstripe/silverstripe-graphql/releases/tag/4.0.0-alpha1

Issue #341, opened 2020-12-16
Remove BuildState global state


chillu:
The `BuildState` class gets passed around half a dozen classes as a "container" for a `Schema`, which is really just a fancy way of using a singleton (which might change over time, e.g. when multiple schemas are built in one process). Global state makes things harder to test and reason about, since it's a somewhat hidden dependency on consumers (e.g. `DataObjectModel` or `Field`).

Aaron suggested that this can be refactored by keeping it in the `$context` argument for resolvers.

unclecheese:
Easiest path:

* `$context` gets a new `schemaConfig` key for easy access to type mapping, etc. No more `BuildState` in resolvers
*  `SchemaModelCreatorInterface` takes a `Schema` param in`createModel`, allowing models to have schema context
*  InheritancePlugin passes `Schema` to `InheritanceChain` service
* `Field` takes new `Schema` constructor arg
* Might as well do `Type`, for consistency

chillu:
This is an example of why globals make things hard to maintain: The controller instanciates a schema, then calls `buildSchema()` with only the schema key, and not passing in the instance. Then it assumes the instance has been affected by the operation without replacing it. https://github.com/silverstripe/silverstripe-graphql/blob/master/src/Controller.php#L149

unclecheese:
Yup, I think the whole reason you need to go through `Build` there is because of the global state it provides. If we got rid of the reliance on global state, that just becomes `$this->getSchema()->save()`.

unclecheese:
Removed `BuildState`in 4.0.0 alpha! https://github.com/silverstripe/silverstripe-graphql/releases/tag/4.0.0-alpha1

Issue #334, opened 2020-12-06
GraphQL 4: filter fields include custom getters


unclecheese:
```
Page: 
  fields:
    '*': true
    link: true
```

Should omit "link" because it is not queryable.

unclecheese:
Fixed in 4.0.0-alpha1 https://github.com/silverstripe/silverstripe-graphql/releases/tag/4.0.0-alpha1

PR #352, opened 2021-01-26
Significant refactor to several annoying inhibitors to stability

unclecheese:
## Summary

Whilst working on the upgrade for silverstripe-gatsby to v4, I came across a lot of things that were slowing me down, and it just ended up being a rabbit hole, but these are all really solid, worthwhile improvements IMO, and to the extent that they break APIs, they're very unlikely to be relied upon by current users.

## Issues addressed/resolved

* https://github.com/silverstripe/silverstripe-graphql/issues/351
* https://github.com/silverstripe/silverstripe-graphql/issues/341
* https://github.com/silverstripe/silverstripe-graphql/issues/334
* https://github.com/silverstripe/silverstripe-graphql/issues/328
* https://github.com/silverstripe/silverstripe-graphql/issues/307
* https://github.com/silverstripe/silverstripe-graphql/issues/270
* https://github.com/silverstripe/silverstripe-graphql/issues/349
* .. And potentially a few others that we'll have to test after merge

## Dependent PRs

- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/1173 (asset-admin)
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/1171 (admin)
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/9835 (framework, docs)
- [ ] https://github.com/silverstripe/silverstripe-versioned/pull/319 (versioned)
- [ ] https://github.com/dnadesign/silverstripe-elemental/pull/870 (elemental)
- [ ] https://github.com/silverstripe/silverstripe-graphql-devtools/pull/35 (graphql-devtools)
- [x] https://github.com/silverstripe/silverstripe-frameworktest/pull/84 (frameworktest)

## TODO

- [x] Aaron: Reconcile two admin.graphql.types commits
- [x] Aaron: Check in with CMS Squad about merge
- [x] Aaron: Remaining elemental fix (unless we can split that out of the PR)
- [x] Fix upgrade guide (you mentioned something on community Slack)
- [x] Ingo: Ensure other module builds are passing
- [ ] ~Ingo: Ensure Behat tests are passing~ They're not, but its likely unrelated. Let's see what CI says.
- [x] Aaron: Answer Ingo’s question about getter side effects (https://github.com/silverstripe/silverstripe-graphql/pull/352#discussion_r575952918)
- [x] Ingo: Write up something about the new deployment complexities of writing files into PUBLIC_PATH in main docs
- [ ] Aaron: Regression test a bit (elemental, asset admin)
- [ ] Ingo: Merge!
- [ ] Alpha1 changelog and release

## Changelog

* Refactor context provider API so that they now plug into `QueryHandlerInterface`. They
can set and get their own context based on their own intnernal logic. Before:
`$user = $context['currentUser'];` after: `$user = UserContextProvider::get($context)`

* `BuildState` API is gone. Use new `SchemaContextProvider` in resolvers. Pass `SchemaContext`
objects to the small number of classes that needed it in their internals.

* `AbstractNestedInputPlugin` is gone. This never made any sense to tie something
so utilitarian to an arbitrary inheritance chain. It has been refactored as `NestedInputBuilder`
and is far more single-purpose. No longer concerned with model and field mapping, but rather
just building recursive input types and adding them to the schema.

* Refactor `QuerySort`, `QueryFilter` to use input builder

* Field mapping is no longer exhaustively encoded in the resolvers. New 
`SchemaContextProvider` allows resolvers to look at the schema as persisted to a
PHP array in the code gen and invoke methods like `$schema->mapField($type, $fieldName)`

* No more `resolveContext` for DataObject\Resolver since field mapping is handled externally now

* Migrate type `Controller::introspectTypes` to standalone service (`SchemaTranscriber`)
invoked through event dispatcher

* Remove `IntrospectionProvider` extension. Static only. Leaving it to the user to decide never made any sense, and it just slowed down the UI in the CMS. Now that we're doing code gen, it seems like the static solution is more consistent. 

* Remove `ModelConfigurationProvider`: Didn't make any sense. Models are now
responsible for fetching their own configurations.

* Remove `defaultResolver`: Didn't make any sense. Just set a default `resolver`
property and allow it to be overriden.

* Resolver discovery is now lazy to avoid race conditions in config. If resolvers
are null at code gen time, discovery is invoked and applied.

* Controller now only accepts a schema key and uses `SchemaFactory` to boot.

* `Schema::quiet()` is gone. Schema is quiet by default to better support autobuilding
and test suites. Opt into verbose with `setVerbose(true)`.

* `CSRFMiddleware` and `HTTPMethodMiddleware` now rely on new ContextProviders
to better encapsultate their state concerns.

* Better error reporting:
  * QueryHandler now pretty prints the stack trace
  * Resolver level errors now throw `ResolverFailure` which express tons of useful
  context, including the query path, the field, the type, the args, and most importantly,
  the resolver chain. No more guessing which closure failed.
  * ResolverFailure relies on new `resolverComposition` field in encoded types.
  * Updates to code gen templates to support the above, and other tidyness like
  variable assignment to save code execution and comments to show closing bracket relationships.

* ComposedResolver is now stateful, mostly for diagnostic purposes. Before: 
`ComposedResolver::create(): Closure` after: `ComposedResolver::create($list)->toClosure()`

* All resolvers use `ComposedResolver` for consistency, even when no middle/afterwares

* `ComposedResolver` no longer needs separate arguments for primary, before, and after
resolvers. Just a stack that leaves ordering a concern of the code composing the class.
(this was a @todo annotation FWIW)

* `SchemaModelInterface` now must declare `getPropertyForField` (gql field -> object property mapping)

* `SchemaModelInterface` now must declare `getSchemaContext` and `getModelConfiguration` 

* New `hasNativeField` in `FieldAccessor` helps with knowing what fields are
filterable/sortable. Basically ORM insight.

* Fields using the `model` property (a bit of a polymorphic hack to support
cases like CMS needing to know what the SiteTree model is named), will
get just-in-time assignment before code generation time to prevent race conditions
in the config.

* New suite of relational `FakeProduct` objects for testing nested queries, filters, mutations.

* New top-to-bottom integration test that creates, updates, and reads nested data structures with filters/sorting, field aliases, etc.
 
* **NB**: `CodeGenerationStore` is now responsible for type mapping and field mapping. This
is a leaky abstraction as it's tightly coupled to SchemaContext now.

* New `StorableSchema` layer eliminates unpredictable state of `Schema` through `process()`. The readonly value object returned by `createStorableSchema()` is expected to
be a complete set of primitive objects that are consumable by a storage service.

* `SchemaFactory` is now `SchemaBuilder`, which has four clearly defined functions, and eliminates the confusion around using `Schema` at query time:
  * `getConfig(string $schemaKey): ?SchemaContext` -- Get the cached configuration of a Schema. Useful for resolvers
  * `boot(string $schemaKey): Schema` -- Load all the configuration in to a Schema object. Useful for the configuration/build process.
  * `getSchema(string $schemaKey): ?GraphQLSchema` -- Retrieve a queryable graphql-php Schema object from the storage service. Useful at query time.
  * `build(Schema $schema, bool $clear = false)`: GraphQLSchema -- put the `Schema` through a storage service and return its queryable schema

* SchemaStorageInterface is now the concern of `SchemaBuilder` rather than `Schema`, which results in a diminished API surface for `Schema`.

* Validation of Schema moved to `StorableSchema`, as this is the only time the schema truly needs to be valid (before storage)

* Type/field mapping done just-in-time for schema storage

* Prefer event dispatcher over extend() hooks in controller

* `SchemaContext` is now `SchemaConfig`

* `Schema` has new `getState()` for persisting state during build

* Improved handling of "empty" schemas that shouldn't be built, but also shouldn't error (e.g. `default` out of the box)


src/Schema/Services/NestedInputBuilder.php
```diff
@@ -0,0 +1,353 @@
...
+use SilverStripe\GraphQL\Schema\Exception\SchemaBuilderException;
+use SilverStripe\GraphQL\Schema\Field\Field;
+use SilverStripe\GraphQL\Schema\Schema;
+use SilverStripe\GraphQL\Schema\Type\InputType;
+use SilverStripe\GraphQL\Schema\Type\ModelType;
+use SilverStripe\GraphQL\Schema\Type\Type;
+use SilverStripe\GraphQL\Schema\Type\TypeReference;
+use SilverStripe\ORM\ArrayLib;
+
+class NestedInputBuilder
```
chillu, line 17: Needs a description of the class purpose

src/Schema/Services/NestedInputBuilder.php
```diff
@@ -0,0 +1,353 @@
...
+            if (!$this->shouldAddField($type, $fieldObj)) {
+                continue;
+            }
+            $namedType = $fieldObj->getNamedType();
+            $nestedType = $this->schema->getTypeOrModel($namedType);
+            if ($nestedType) {
+                if ($level > $this->config()->get('max_nesting')) {
+                    continue;
+                }
+                // Prevent stupid recursion in self-referential relationships, e.g. Parent
```
chillu, line 139: Seems like a good candidate for a test?

tests/Schema/Services/NestedInputBuilderTest.php
```diff
@@ -0,0 +1,90 @@
...
+    protected static $extra_dataobjects = [
+        FakeProductPage::class,
+        FakeProduct::class,
+        FakeReview::class,
+        Member::class,
+    ];
+
+    protected static $fixture_file = '../fixtures.yml';
+
+    public function testNestedInputBuilder()
```
chillu, line 28: Awesome to see the test setup being so self explanatory

_config/events.yml
```diff
@@ -0,0 +1,7 @@
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\EventDispatcher\Dispatch\Dispatcher:
+    properties:
+      handlers:
+        graphqlTranscribe:
+          on: [ graphqlSchemaBuild ]
```
chillu, line 6: This is technically only required for Apollo usage in the CMS, right? Can we limit this to the `admin` schema by default? I'm assuming there either are other GraphQL clients who ignore the type ambiguity, or simpler schemas where that ambiguity is not an issue?

src/Schema/DataObject/Plugin/QuerySort.php
```diff
@@ -90,52 +94,61 @@ protected function buildAllFieldsConfig(ModelType $modelType, Schema $schema, in
...
         };
     }
 
+    /**
+     * @param NestedInputBuilder $builder
+     */
+    protected function updateInputBuilder(NestedInputBuilder $builder): void
+    {
+        parent::updateInputBuilder($builder);
+        $builder->setFieldFilter(function (Type $type, Field $field) {
```
chillu, line 142: Wouldn't this need to be serialised out to generated PHP? I don't understand how this works :D

src/Schema/Schema.php
```diff
@@ -471,34 +520,32 @@ private function collectSchemaUpdaters(array $components): array
      */
     public function save(): void
     {
+        // Fill in lazily defined type properties, e.g. fields with a classname as a type
+        $this->processTypes();
```
chillu, line 524: Given this is a build-time step, where performance is (somewhat) less important than during query execution: What is the purpose of this lazy mechanism? It feels like it's adding another layer of complexity, where you have to assume inconsistent types which need processing before the schema can be persisted?

src/Schema/SchemaContext.php
```diff
@@ -130,4 +122,84 @@ public static function pluralise(string $typeName): string
         $typeName .= 's';
         return $typeName;
     }
+
+    /**
+     * @param array $typeMapping
```
chillu, line 127: Can you define which type the array should contain? Same with `setFieldMapping()` below.

src/Schema/Storage/templates/type.inc.php
```diff
@@ -32,12 +32,21 @@ public function __construct()
...
+                $fields = [];
                 <?php foreach ($type->getFields() as $field) : ?>
-                    [
+                    <?php $resolver = $field->getEncodedResolver($type->getName()); ?>
+                    $resolverInst = <?=$resolver->encode(); ?>;
+                    $fields[] = [
                         'name' => '<?=$field->getName(); ?>',
                         'type' => <?=$field->getEncodedType()->encode() ?>,
-                        'resolve' => <?=$field->getEncodedResolver($type->getName())->encode(); ?>,
+                        'resolve' => $resolverInst->toClosure(),
```
chillu, line 42: Does this introduce new limits about how resolvers can be defined? I guess they were always static, so there's really no opportunity to close over values that can't be expressed in PHP code?

chillu (changes requested):
Amazing work as usual, thanks for chipping away at this! Definitely a tough one to peer review as a single commit, but I appreciate the effort to create a changelog - that helps a bit ;)

> SchemaModelInterface now must declare getPropertyForField (gql field -> object property mapping)
> SchemaModelInterface now must declare getSchemaContext and getModelConfiguration

I don't understand this layer enough to determine if that's a good change. Might need to write a bit of an architecture overview doc to get my head around it.

> BuildState API is gone. Use new SchemaContextProvider in resolvers. Pass SchemaContext
objects to the small number of classes that needed it in their internals.

Yep that's a great change, a lot more explicit and testable.

chillu:
> Remove IntrospectionProvider extension. Static only. Leaving it to the user to decide never made any sense, and it just slowed down the UI in the CMS. Now that we're doing code gen, it seems like the static solution is more consistent.

How will that work when the schema build is performed outside of the production environment, e.g. during CI when creating a deployment package? That environment won't have access to the production `assets/` folder.

src/Controller.php
```diff
@@ -327,28 +285,21 @@ protected function validateOrigin($origin, $allowedOrigins)
...
-                $provider->setMemberContext($member);
-            }
+        $request = $this->getRequest();
+        $user = $this->getRequestUser($request);
+        $token = $this->getToken();
+
+        $handler->addContextProvider(UserContextProvider::create($user))
+                ->addContextProvider(TokenContextProvider::create($token))
+                ->addContextProvider(RequestContextProvider::create($request));
+        if ($schema) {
```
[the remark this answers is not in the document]
unclecheese: Yup, there's no reason Schema should be passed around at query time. That's why the context provider is just giving you the `SchemaContext` object. That class is designed to encapsulate only that schema state that might be relevant to you in resolvers.

I don't know why it takes Schema as its constructor. Maybe the idea is that `provideContext` could at some point provide more keys, like:

```php
    public function provideContext(): array
    {
        return [
            self::KEY => $this->schema->getStore()->getContext(),
            'types' => array_keys($this->schema->getTypes())
        ];
    }
```

But as far as I can tell, this should be constructed with a SchemaContext object.

_config/events.yml
unclecheese: Checked in on this. Apollo needs it when you have fragments that are pinned to ambiguous types (Unions and Interfaces). We do that in asset-admin.

We could just forget the whole thing if we ditched the fragments and duplicated some code, or we could just press forward and make this work only on admin.

Easiest thing is to just pin it to the admin schema using the event dispatcher's inbuilt dot syntax:

```
   properties:
      handlers:
        graphqlTranscribe:
          on: [ graphqlSchemaBuild.admin ]
```

src/Schema/DataObject/Plugin/QuerySort.php
unclecheese: The way these classes are named makes this very hard to understand. Basically there are three kinds of these query plugins:

* Something super opinionated that you apply to a model operation (MyObject read: true)
* Something more flexible that you can apply to a custom query that returns dataobjects
* Something very generic that leaves you to do all the wiring (e.g. just adds the input types to the schema, and adds an argument)

We need to embrace better naming conventions for all these cases. Right now, the space is very overloaded with generic terminology.

But in this case, this is basically #2 on the list asserting that in addition to what #3 does, it only works on fields that are model-backed.

This is a bit messy ATM, and there's a hackday PR in the future. :)

src/Schema/Schema.php
unclecheese: It ensures that plugins are always dealing with a schema state that is as close to the original as possible, so minimise the impact of order of execution in the plugin list.

Because plugins have the ability to mutate the schema, you can have things like `query->setType()` to some new thing (e.g. pagination), and that can obfuscate important information to a plugin.

I wish I could think of the example now, but I think it was pagination, where you had a query like `readPages` returning type `SilverStripe\CMS\Model\SiteTree`.

src/Controller.php
```diff
@@ -327,28 +285,21 @@ protected function validateOrigin($origin, $allowedOrigins)
...
-            /* @var MemberContextProvider $provider */
-            $member = $this->getRequestUser($request);
-            if ($member) {
-                $provider->setMemberContext($member);
-            }
+        $request = $this->getRequest();
+        $user = $this->getRequestUser($request);
+        $token = $this->getToken();
+
+        $handler->addContextProvider(UserContextProvider::create($user))
```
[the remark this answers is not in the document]
unclecheese: > Also, any middleware is receiving $context as an array. Do we need a ContextProvider::set($myArray, $myValue) equivalent so that middleware and resolvers can deal with the same level of abstraction here?

If a middleware wants to mutate context, it should create a ContextProvider to go with it that contains internals for get and set. The idea is to get away from having to guess what the name of the key is. Classes are easier to discover than strings. :-)


> That's a pretty roundabout way to wrap a strongly typed API around a weakly typed one inherited by the library.

I think, as we discussed earlier, modules make a good case for doing it this way, e.g. fluent / subsites, which might want to provide context in some kind of middleware layer.

src/Schema/Services/NestedInputBuilder.php
unclecheese: It actually is tested. `FakeProduct` has a self-referential relation to `relatedProducts`. If this functionality didn't work, the `assertSchema` test would fail because there would be more inputs.

src/Schema/Storage/templates/type.inc.php
unclecheese: Seems unlikely. `CompsedResolver` should never be used externally.

unclecheese: Actually, I think we need to validate that the callables are not instance of Closure. Good call.

unclecheese:
> How will that work when the schema build is performed outside of the production environment, e.g. during CI when creating a deployment package? That environment won't have access to the production assets/ folder.

OK, now:

* Remove `GeneratedAssetHandler` dependency from `SchemaTranscriber` (injector hell for one little file)
* `admin` module adds event for `graphqlSchemaBuild.admin` (only applies to admin schema)
* `SchemaTranscriber` has new `rootDir` property that defaults to `PUBLIC_PATH`
* Graphql special `assets.yml` config (behind moduleexists) sets `rootDir` of `SchemaTranscriber` from `PUBLIC_PATH` to `ASSETS_PATH`
* Document edge case for running SS without assets installed -- block those `.graphql` files to outside traffic

src/Schema/DataObject/FieldAccessor.php
```diff
@@ -99,6 +99,42 @@ public function hasField(DataObject $dataObject, string $field): bool
...
 
+    /**
+     * Returns true if the field is part of the ORM data structure
+     * @param DataObject $dataObject
+     * @param string $field
+     * @param bool $includeUnary
+     * @param bool $includeList
+     * @return bool
+     */
+    public function hasNativeField(
```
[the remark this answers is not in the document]
unclecheese: We now have full test coverage on `FieldAccessor`! Easy win.

chillu:
I've sent a (very straightforward) devtools PR for this: https://github.com/silverstripe/silverstripe-graphql-devtools/pull/33

src/Controller.php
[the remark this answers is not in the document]
chillu: Sweet, I've done this in 3dfcca7e778cd23f04058d0bcdc1cdee260e61cc

chillu:
Also fixed frameworktest with https://github.com/silverstripe/silverstripe-frameworktest/pull/84

src/Controller.php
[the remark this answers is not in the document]
chillu: OK that's done now in https://github.com/silverstripe/silverstripe-graphql/pull/352/commits/3dfcca7e778cd23f04058d0bcdc1cdee260e61cc

src/Controller.php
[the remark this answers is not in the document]
chillu: Yeah modules with their own context providers validate that level of composition. I've documented that through https://github.com/unclecheese/silverstripe-framework/commit/1be5f471de7bce661f1db18c78bbcf9371dfefa5

unclecheese:
Test failure is due to the `silverstripe/versioned` dependency.

src/Schema/SchemaBuilder.php
```diff
@@ -11,50 +11,98 @@
...
+     * which does not require booting. Useful for getting data from
+     * a saved schema at request time.
      * Returns null when no stored schema can be found.
      *
      * @param string $key
-     * @return Schema|null
+     * @return SchemaContext|null
      */
-    public function get(string $key): ?Schema
+    public function read(string $key): ?SchemaContext
```
chillu, line 47: This feels like a poor naming choice: `read()` returns `SchemaContext`, `fetch()` returns `GraphQLSchema`. How about `getContext()` and `getSchema()`? Using "fetch" rather than "get" feels a bit arbitrary, especially since it deviates from the "inner" methods in `getStoreCreator()`

src/Schema/SchemaContext.php
```diff
@@ -165,7 +165,16 @@ public function setFieldMapping(array $fields): self
      */
     public function getTypeNameForClass(string $class): ?string
     {
-        return $this->get(['typeMapping', $class]);
+        $name = $this->get(['typeMapping', $class]);
+        if ($name) {
+            return $name;
+        }
+        $model = $this->createModel($class);
```
chillu, line 172: Getters with side effects? Shouldn't that be an explicit decision by the caller? The `createModel` call creates a cache entry, and the creator gets a reference to this context (`$creator->createModel($class, $this)`) - so execution order might matter here. Could be that the opaqueness is a feature here, just checking.

chillu (changes requested):
Refactoring the `StorableSchema` as a value object is great, much cleaner. I'm still a bit confused why we need a `Schema` that's *separate* from a `SchemaBuilder` during the build phase. Feels like this could be inlined into `SchemaBuilder` to avoid confusion between the two "schema" types which perform very different objectives. But I don't think it's worth extra effort at this point, you describe the responsibilities pretty clearly in PHPDoc - and we've already identified the need for an architecture overview doc.

> How will that work when the schema build is performed outside of the production environment, e.g. during CI when creating a deployment package? That environment won't have access to the production assets/ folder.

I don't think your explanation covers it. The "run cms without assets module" edge case is worth covering, but I'm more concerned about building the schema during CI. If this `<schema>.types.graphql` file is created *anywhere other than* `PUBLIC_PATH`, it simply won't be accessible by the browser. You removed the dynamic fallback which is fine in general (since creation is part of a build that's a necessary precondition anyway). But you also removed the dynamic passthrough of a file's contents when they don't live in `PUBLIC_PATH`.

That's important to resolve, because it means that CI-based builds would break `admin/assets`. The schema cache files exist on the webserver(s) because they're part of the production deployment. The types file doesn't, because it's either stuck in CI on `PUBLIC_PATH` (which will get filtered out of the deployment artefact), or because it's in a path where the webserver can't directly serve it without going through `file_get_contents()`.

src/Schema/Schema.php
chillu: Not sure I'm following, but I trust your judgement on this :) Thanks for taking the time to explain!

chillu:
Just noted the overlap with https://github.com/silverstripe/silverstripe-framework/issues/9688 - by merging this PR we're making a stronger (not yet semver binding) commitment to supporting `silverstripe/event-dispatcher`

src/Schema/SchemaBuilder.php
unclecheese: Yeah this gets a bit tricky. So the intention here is to make it clear to the user what they can do. In resolvers interacting with the schema, you're performing readonly operations. 

Latest commit uses `getConfig(string $key): SchemaConfig` 👍

`getSchema()` is ambiguous, because `boot` returns a Schema object, but I think we can live with that.

tests/Schema/SchemaBuilderTest.php
```diff
@@ -39,7 +39,7 @@ public function testFetch()
             [
                 'properties' => [
                     'handlers' => [
-                        'graphqlTranscribe' => [
+                        'graphqlSchemaBuild' => [
```
unclecheese, line 42: These identifiers are actually arbitrary. 🤷‍♂️

src/Schema/StorableSchema.php
```diff
@@ -118,6 +118,28 @@ public function getConfig(): SchemaConfig
         return $this->config;
     }
 
+    /**
+     * @return bool
+     */
+    public function exists(): bool
+    {
+        $queryType = $this->types[Schema::QUERY_TYPE] ?? null;
```
chillu, line 126: Technically you could have a schema with *just* mutations, right?

src/Schema/SchemaContext.php
chillu: Aaron has refactored this now

src/Schema/StorableSchema.php
unclecheese: Nope. Believe it or not, at least one query is required.

unclecheese: Actually, i _think_ the reason for this is for the IDE. It works based on `query {  __schema }` which can't exist if there's no query type, and you can't just leave an empty query type in there with no fields, either. So `query` is implicitly required.

chillu (approved):

PR #363, opened 2021-02-20
FIX: NestedInputBuilder correctly builds filters for models with multiple relations of the same type

### LINKED ISSUE #362: GraphQL 4: Filter plugin fails to produce complete input types if multiple filtered connections with the same type are present
`addInputTypesToSchema` in `NestedInputBuilder` tries to resolve nested input types by recursing into them, applying the  field on the parent type only when a new field was added to the current type. Considering the following definition:
```yml
Member:
  fields:
    likedCats:
      plugins:
        filter: true
        paginateList: true
    adoptedCats:
      plugins:
        filter: true
        paginateList: true
  operations:
    readOne:
      plugins:
        filter: true

Cat:
  fields:
    likedBy: true
    adoptedBy: true
```
where `likedCats` and `adoptedCats` are a:N or N:M relations. The generated input type for `MemberFilterFields` is produced as such:
```graphql
input MemberFilterFields {
  id: QueryFilterIDComparator
  likedCats: CatFilterFields
}
```
This is because on the first pass through Member, the `likedCats` field on the `MemberFilterFields`type will generate all fields for the `CatFilterFields` type. When it reaches the `adoptedCats` field, it will skip through the child fields, never applying `adoptedCats` as `CatFilterFields` to the `MemberFilterFields`type.

The reason for this is here: https://github.com/silverstripe/silverstripe-graphql/blob/4011f0d815b47702e733506ad9e4f9785c111fda/src/Schema/Services/NestedInputBuilder.php#L183-L186

I have already come up with a fix and opened a PR.

mleutenegger:
Closes #362


mleutenegger:
Added a test to check for a complete filter type

unclecheese (approved):

PR #368, opened 2021-02-23
MINOR: Make paginator injectable

unclecheese:


unclecheese:
This is a really low-impact PR that is blocking another critical fix, so I'm self-merging.

PR #371, opened 2021-03-01
FIX: input building in cyclic graphs

mleutenegger:
This PR aims to fix a bug in the `NestedInputBuilder` where a cyclic graph prohibits the building of all necessary nested types on filter and sort input-types.

### DIFF
--- a/src/Schema/Services/NestedInputBuilder.php
+++ b/src/Schema/Services/NestedInputBuilder.php
@@ -123,8 +123,8 @@ class NestedInputBuilder
      * @return array
      * @throws SchemaBuilderException
      */
-     protected function buildAllFieldsConfig(Type $type, int $level = 0, array &$seenConnections = []): array
-     {
+    protected function buildAllFieldsConfig(Type $type, int $level = 0, array &$seenConnections = []): array
+    {
         $existing = $this->fetch($type->getName());
         if ($existing) {
             return $existing;
@@ -159,7 +159,7 @@ class NestedInputBuilder
         }
         $this->persist($type->getName(), $map);
         return $map;
-     }
+    }
 
     /**
      * @param Type $type
--- a/src/Schema/Services/NestedInputBuilder.php
+++ b/src/Schema/Services/NestedInputBuilder.php
@@ -119,20 +119,15 @@ class NestedInputBuilder
 
     /**
      * @param Type $type
-     * @param int $level
      * @return array
      * @throws SchemaBuilderException
      */
-    protected function buildAllFieldsConfig(Type $type, int $level = 0, array &$seenConnections = []): array
+    protected function buildAllFieldsConfig(Type $type): array
     {
         $existing = $this->fetch($type->getName());
         if ($existing) {
             return $existing;
         }
-        if (!isset($seenConnections[$type->getName()])) {
-            $seenConnections[$type->getName()] = [];
-        }
-        $level++;
         $map = [];
         foreach ($type->getFields() as $fieldObj) {
             if (!$this->shouldAddField($type, $fieldObj)) {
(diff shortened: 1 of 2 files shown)


src/Schema/Services/NestedInputBuilder.php
```diff
@@ -123,12 +123,15 @@ public function populateSchema()
...
+    protected function buildAllFieldsConfig(Type $type, int $level = 0, array &$seenConnections = []): array
     {
         $existing = $this->fetch($type->getName());
         if ($existing) {
             return $existing;
         }
+        if (!isset($seenConnections[$type->getName()])) {
+            $seenConnections[$type->getName()] = [];
+        }
         $level++;
```
unclecheese, line 135: If we're getting rid of `max_nesting`, I think we can do away with this parameter
```suggestion

```

src/Schema/Services/NestedInputBuilder.php
```diff
@@ -138,21 +141,23 @@ protected function buildAllFieldsConfig(Type $type, int $level = 0): array
             $namedType = $fieldObj->getNamedType();
             $nestedType = $this->schema->getTypeOrModel($namedType);
             if ($nestedType) {
-                if ($level > $this->config()->get('max_nesting')) {
-                    continue;
-                }
+                // if ($level > $this->config()->get('max_nesting')) {
```
unclecheese, line 144: Is it the conclusion from this issue that the edge cases around `max_nesting` make implementation too problematic to support? I'm fine with that. It's mostly a build performance optimisation.

src/Schema/Services/NestedInputBuilder.php
```diff
@@ -138,21 +141,23 @@ protected function buildAllFieldsConfig(Type $type, int $level = 0): array
...
+                //     continue;
+                // }
                 // Prevent stupid recursion in self-referential relationships, e.g. Parent
                 if ($namedType === $type->getName()) {
                     $map[$fieldObj->getName()] = self::SELF_REFERENTIAL;
+                } elseif (isset($seenConnections[$type->getName()][$fieldObj->getName()])) {
```
unclecheese, line 150: Let's use the new `Schema::getState()` API for this. There are too many cases in plugins where you need to track state across a build, which is why I added this recently.

```php
$seen = $this->schema->getState()->get([
  static::class,
  'seenConnections',
  $type->getName(),
  $fieldObj->getName()
]);
```

src/Schema/Services/NestedInputBuilder.php
```diff
@@ -138,21 +141,23 @@ protected function buildAllFieldsConfig(Type $type, int $level = 0): array
...
+                    continue;
                 } else {
-                    $map[$fieldObj->getName()] = $this->buildAllFieldsConfig($nestedType, $level);
+                    $seenConnections[$type->getName()][$fieldObj->getName()] = true;
```
unclecheese, line 153:
```suggestion
$this->schema->getState()->set([
  static::class,
  'seenConnections',
  $type->getName(),
  $fieldObj->getName()
], true);

File: src/Schema/Services/NestedInputBuilder.php [hunk 5]
```diff
@@ -123,12 +123,15 @@ public function populateSchema()
      * @return array
      * @throws SchemaBuilderException
      */
-    protected function buildAllFieldsConfig(Type $type, int $level = 0): array
+    protected function buildAllFieldsConfig(Type $type, int $level = 0, array &$seenConnections = []): array
```
unclecheese, line 126:
```suggestion
    protected function buildAllFieldsConfig(Type $type): array
```

src/Schema/Services/NestedInputBuilder.php
```diff
@@ -138,21 +141,23 @@ protected function buildAllFieldsConfig(Type $type, int $level = 0): array
...
+                // }
                 // Prevent stupid recursion in self-referential relationships, e.g. Parent
                 if ($namedType === $type->getName()) {
                     $map[$fieldObj->getName()] = self::SELF_REFERENTIAL;
+                } elseif (isset($seenConnections[$type->getName()][$fieldObj->getName()])) {
+                    continue;
                 } else {
-                    $map[$fieldObj->getName()] = $this->buildAllFieldsConfig($nestedType, $level);
+                    $seenConnections[$type->getName()][$fieldObj->getName()] = true;
+                    $map[$fieldObj->getName()] = $this->buildAllFieldsConfig($nestedType, $level, $seenConnections);
```
unclecheese, line 154:
```suggestion
                    $map[$fieldObj->getName()] = $this->buildAllFieldsConfig($nestedType);
```

unclecheese (changes requested):
Looks good. Just some tidy up and a new API call I would recommend using.

src/Schema/Services/NestedInputBuilder.php
unclecheese: If so, let's remember to remove the config setting and update any documentation that points to it.

mleutenegger: :+1: I'd even argue that using a solution which terminates on known edges is faster. Not going to do the math though, as its a prebuild step and really not impacting performance of the application.

unclecheese (approved):
Beautiful. Excellent work!

Issue #376, opened 2021-05-06
Graphql 3.x-dev compatibility with upcoming CWP 2.8


emteknetnz:
Installing kitchen-sink 2.x-dev with graphql 3.x-dev

File manager:

![image](https://user-images.githubusercontent.com/4809037/117229925-195c9480-ae70-11eb-91b8-6510de42a47d.png)

Installing kitchen-sink 2.7.1 and manually using graphql 3.x-dev instead of 3.4.1

![image](https://user-images.githubusercontent.com/4809037/117230017-385b2680-ae70-11eb-9ef9-6d4428253e37.png)

Some debugging notes:

```
- Manager:408
- UnionScaffolder:80 - $this->type has CWPBasePage
- $_name = SilverStripeSiteTreeWithDescendants (for both exc's)
- UnionScaffolder:36
- InheritanceScaffolder:118
- :127 $tree[2] = 'CWP\CWP\PageTypes\BasePage' ($class)
- StaticSchema:109
- :112 $customTypeName = $this->mappedTypeName($class); - null
- :118 $typeName = 'CWPBasePage'
- $parts = explode('\\', $class)
    array (
        0 => 'CWP',
        1 => 'CWP',
        2 => 'PageTypes',
        3 => 'BasePage',
    )
    $typeName = sizeof($parts) > 1 // (true)
      ? $parts[0] . end($parts) // <<< CWPBasePage
      : $parts[0];
```

Possibly this commit? https://github.com/silverstripe/silverstripe-graphql/commit/25f654c9fd33327b6107158868a636a6090d4d13#diff-b60efc42f855dbb4c758a6216ded97c98c345b5606e1820014afc5e174ebf473

emteknetnz:
I'm not sure if this is the greatest solution, though it does stop asset-admin from complaining about graphql types

**myproject.yml**
```
SilverStripe\GraphQL\Manager:
  schemas:
    admin:
      typeNames:
        CWP\CWP\PageTypes\BasePage: Page
```

Trying to use `CWP\CWP\PageTypes\BasePage: BasePage` results in the error "Type 'BasePage' is not a registered GraphQL type"

My understanding is that Page will usually inherit BasePage on CWP projects, so this seems near enough

This config is really just to get graphql to not complain.  I don't think there are any real world use case where graphql is used to read BasePage's specifically

maxime-rainville:
[...]

PR #377, opened 2021-05-06
DEP Require framework ^4.8

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/416

Travis failures are [existing](https://travis-ci.com/github/silverstripe/silverstripe-graphql/builds/224955671)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,7 @@
     "type": "silverstripe-vendormodule",
     "license": "BSD-3-Clause",
     "require": {
-        "silverstripe/framework": "^4.2",
+        "silverstripe/framework": "^4.8",
         "silverstripe/vendor-plugin": "^1.0",
         "webonyx/graphql-php": "~0.12.6"
     },


maxime-rainville:
Can you explain why we need to bump this to framework 4.8? Is GraphQL 3.5 using an API that got introduced in framework 4.8? 

If not, what's the value in explicitly requiring 4.8?

emteknetnz:
To the best of my knowledge, it's because there was a lot of core modules that are in lockstep (not just framework) that had graphql3 changes made to them ([beta changelog](https://docs.silverstripe.org/en/4/changelogs/beta/4.8.0-beta1/)) as part of supported both graphql3 and graphql4.  Because graphql4 queries are a bit different (use of lowercase, etc) graphql3 was updated to allow for graphql4 style queries

The Graphql3 module seems like it also had some changes made to it to work with this way https://github.com/silverstripe/silverstripe-graphql/compare/636c212..3.4.1

Because of all these changes, I'm really not sure exactly how compatible graphql ^3.5 would be with 4.7.3 and below (its current requirement is very loose at ^4.2).  We know of at least some issues with non core modules.  It may well be that adding framework ^4.8 isn't strictly necessary, though at this stage it seems like the safest option

maxime-rainville:
If the problem is that a bunch of module had changes made to them to work with GraphQL 3.5, forcing the installation of framework  4.8 is not going to solve this. You can still install admin 1.7 with framework 4.8.

emteknetnz:
We know there's [issues](https://github.com/silverstripe/silverstripe-graphql/issues/376#issue-876993441) with CWP kitchen-sink 2.7.1 running graphql 3.5

The current [proposal](https://github.com/silverstripe/silverstripe-graphql/issues/376#issuecomment-833200950) to patch CWP would only be available on CWP 2.8+

We need some way to ensure that CWP 2.7 sites that stay on framework ~4.7 with a ^3 graphql requirement don't auto update to graphql 3.5 which will probably break their site :-/  We cannot alter the requirements of existing tagged versions and put an upper bounds on the graphql 3 requirement.  Requiring framework ^4.8 on graphql should meet this goal?

maxime-rainville:
I'm wondering if it would be better to use `conflict` here? https://getcomposer.org/doc/04-schema.md#conflict

Basically we could say this version of graphql is not compatible with any version of `asset-admin` smaller than 1.8, etc. This seems more expressive than putting a semi-random constraint on framework.

But I agree that this kind of sucks and that we'll have to make the best of a bad situation.

emteknetnz:
`conflict` seems more correct than framework: ^4.8 since there's no requirement on API in framework 4.8

The non-recipe requirements of cwp are loose e.g. cwp/cwp-core 2.7.1 requires silverstripe/admin:^1.7 - using conflict would ensure that if asset-admin 1.7.x was installed (which requires admin ^1.7) it would stay on graphql 3.4.1

maxime-rainville:
`conflict` also has the advantage that it will prevent projects who have inlined all their recipes from accidentally upgrading themselves to graphql 3.5.

so we probably need to do something like:
```json
"conflict": {
   "silverstripe/admin": "<1.8",
   "silverstripe/versioned": "<1.8",
   "silverstripe/asset-admin": "<1.8",
   "dnadesign/silverstripe-elemental": "<4.6"
}
```

emteknetnz:
Closing this PR in favor of the `conflict` approach

PR #379, opened 2021-05-10
API Block older version module from using this minor

maxime-rainville:
Version 3.5 of GraphQL expects different cases for some queries. Older version of modules that depend on this module will not work with 3.5.

This PR aims to make this conflict explicit.

# Parent issue
- https://github.com/silverstripeltd/product-issues/issues/416

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,6 +8,12 @@
         "silverstripe/vendor-plugin": "^1.0",
         "webonyx/graphql-php": "~0.12.6"
     },
+    "conflict": {
+        "silverstripe/admin": "<1.8",
+        "silverstripe/versioned": "<1.8",
+        "silverstripe/asset-admin": "<1.8",
+        "dnadesign/silverstripe-elemental": "<4.6"
+    },
     "require-dev": {
         "sminnee/phpunit": "^5.7",
         "sminnee/phpunit-mock-objects": "^3.4.5",


maxime-rainville:
I created fake releases on the creative commoners repo to test this:
- `3.4.0` tracks the head of the `3.4` branch
- `3.5.0` includes this fix.

I use this composer file to test.
```json
{
    "name": "maxime/boom",
    "type": "project",
    "authors": [
        {
            "name": "Maxime Rainville",
            "email": "maxime@silverstripe.com"
        }
    ],
    "minimum-stability": "dev",
    "require": {
      "silverstripe/graphql": "^3.4",
      "silverstripe/admin": "~1.7.0"
    },
    "prefer-stable": true,
    "repositories": [
        {
            "type": "vcs",
            "url": "git@github.com:creative-commoners/silverstripe-graphql.git"
        }
    ]
}
```

I've confirmed that changing the `admin` constraint between `~1.8.0` and `~1.7.0` will upgrade/downgrade `graphql`.

maxime-rainville:
I'll double check that the modules included the conflict require `graphql` 3.5.0 or greater.

composer.json
```diff
@@ -8,6 +8,12 @@
         "silverstripe/vendor-plugin": "^1.0",
         "webonyx/graphql-php": "~0.12.6"
     },
+    "conflict": {
```
maxime-rainville, line 11: I used the list of dependent module from packagist to determine what should show in here https://packagist.org/packages/silverstripe/graphql/dependents

I only included supported module.

maxime-rainville: Actually, that list looks wrong. `admin` and `versioned` shouldn't be here. `versioned-admin` should.

emteknetnz (approved):
Travis failures are unrelated

PR #391, opened 2021-06-07
[CVE-2020-26136] Disable basic-auth for graphql requests

emteknetnz:


maxime-rainville (approved):

emteknetnz:
Both the 3.5 and 3 branches have this change already

https://github.com/silverstripe/silverstripe-graphql/blob/3.5/src/Auth/Handler.php#L31
https://github.com/silverstripe/silverstripe-graphql/blob/3/src/Auth/Handler.php#L31

PR #393, opened 2021-06-14
Pulls/master/ extendsive refactor

unclecheese:
See https://github.com/silverstripe/silverstripe-graphql/pull/374


src/Schema/DataObject/Plugin/InheritedPlugins.php
```diff
@@ -36,6 +36,7 @@ public function getIdentifier(): string
      */
     public function apply(ModelType $type, Schema $schema, array $config = []): void
     {
+        return;
```
maxime-rainville, line 39: Was that a mistake?

PR #407, opened 2021-10-05
[CVE-2021-28661] Add a CanViewPermissionChecker if permission checker is null

maxime-rainville:
Backporting CVE-2021-28661


maxime-rainville:
This commit has already been independently approved.

Issue #19, opened 2016-11-29
Handle file uploads through GraphQL


marcello-silverstripe:
Handle file uploads through something like (https://medium.com/@danielbuechele/file-uploads-with-graphql-and-apollo-5502bbf3941e#.8md29rk3f)

blueo:
I've done this before in JS using [apollo-upload-network-interface](https://github.com/HriBB/apollo-upload-network-interface) and [graphql-server-express-upload](https://github.com/HriBB/graphql-server-express-upload). I've created a quick and dirty demo of a similar thing using the silverstripe module [here](https://github.com/blueo/silverstripe-graphqldemo) if you're interested.

PR #418, opened 2021-12-08
Implement support for query batching (closes #177)

### LINKED ISSUE #177: Support query batching
# Overview

Query batching allows for multiple queries to be processed in a single HTTP request, reducing network overheads, and allowing in-process optimisations. It is already part of the webonyx PHP library we're using (see http://webonyx.github.io/graphql-php/executing-queries/#query-batching). 

# Acceptance Criteria

 * A GraphQL client can submit [apollo style](https://dev-blog.apollodata.com/query-batching-in-apollo-63acfd859862) query batches
 * Query batching executes all queries in parallel (through async PHP?).
 * Query batching works with the upcoming persisted queries feature
 * Query batching is turned on for CMS requests

# Notes

 * Without async PHP, this feature will make overall UI performance *slower*
 * Assuming enough server capacity (workers, memory), this feature isn't as important with HTTP/2 (less network overhead)

kinglozzer:
See #177

This adds support for query batching using [Apollo’s `BatchHttpLink`](https://www.apollographql.com/docs/react/api/link/apollo-link-batch-http/), while preserving support for regular, non-batched queries and persisted queries. I’m not sure if it’s possible to support _both_ batched and persisted queries in the one request, I can’t find anything in the docs to suggest that it is.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,8 @@
         "silverstripe/framework": "^4.10",
         "silverstripe/vendor-plugin": "^1.0",
         "webonyx/graphql-php": "^14.0",
-        "silverstripe/event-dispatcher": "^0.1.2"
+        "silverstripe/event-dispatcher": "^0.1.2",
+        "ext-json": "*"
     },
     "require-dev": {
         "phpunit/phpunit": "^9.5",
--- a/src/Controller.php
+++ b/src/Controller.php
@@ -11,6 +11,7 @@ use SilverStripe\Control\Controller as BaseController;
 use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Control\HTTPResponse_Exception;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\EventDispatcher\Dispatch\Dispatcher;
@@ -26,6 +27,7 @@ use SilverStripe\GraphQL\QueryHandler\SchemaConfigProvider;
 use SilverStripe\GraphQL\QueryHandler\TokenContextProvider;
 use SilverStripe\GraphQL\QueryHandler\UserContextProvider;
 use SilverStripe\GraphQL\Schema\SchemaBuilder;
+use SilverStripe\ORM\ArrayLib;
 use SilverStripe\Security\Member;
 use SilverStripe\Security\Permission;
 use SilverStripe\Versioned\Versioned;
@@ -92,7 +94,8 @@ class Controller extends BaseController
      *
      * @param HTTPRequest $request
      * @return HTTPResponse
-     * @throws InvalidArgumentException
+     * @throws BadMethodCallException
+     * @throws HTTPResponse_Exception
      */
     public function index(HTTPRequest $request): HTTPResponse
     {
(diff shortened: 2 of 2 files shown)


kinglozzer:
I’ve had a look at this with fresh(er) eyes today. This PR is the bare minimum in terms of API changes to offer batch query functionality, but it could certainly be improved - e.g. the schema builder only needs to run once, not once per batched query. However, it piqued my curiosity in how other projects handle the HTTP part of GraphQL queries/mutations.

@unclecheese I’d like to preface this next section by saying that I’m sure you must have a lot on your plate with getting GraphQL v4 stable-release-ready, as it seems you’ve been working on this alone. If you think what I’m proposing below is too large a change, or too late for v4, I understand. We’ve not had the opportunity until very recently to start working on a new project with React + GraphQL, and predictably I now need both batched queries & file upload support!

---

I’d like to propose a bit of an overhaul to our HTTP request handling.

There’s a request parser in `laragraph/utils` that we could possibly borrow some ideas from for the initial parsing of request data: https://github.com/laragraph/utils/blob/master/src/RequestParser.php. The `application/graphql` content type looks fairly easy to add support for.

We’re lacking support for `multipart/form-data` file uploads (https://github.com/silverstripe/silverstripe-graphql/issues/19). This is obviously a larger undertaking than batched queries, but perhaps there’s some inspiration that could be taken from https://github.com/rebing/graphql-laravel#file-uploads. I’m only really raising this here because if I’m adjusting our HTTP request handling in this PR, it makes sense to plan ahead if file upload support is a long-term goal. If the transport-level support is already baked in, hopefully it could be a future addition without any API changes.

`webonyx/graphql-php` now has a [helper class](https://github.com/webonyx/graphql-php/blob/master/src/Server/Helper.php) that contains a bunch of logic for helping with persisted queries, I know this didn’t exist when persisted query support was originally added to this module. Perhaps we could offload some code to the webonyx library here - the existing providers offered (file, JSON, HTTP) could be retained and adjusted to fit the [webonyx API](https://github.com/webonyx/graphql-php/blob/fe747025212b4a93fe81caf91b2b55f04ae57a15/src/Server/Helper.php#L331-L337), which appears to just be a callable function.

Actually reaching a point where we could offload logic to the webonyx library would mean a reasonable refactor of the controller and `QueryHandler` classes, and probably some API changes to boot as I think we’d need to parse request data into `OperationParams` objects. At a glance I think the ideal entry points would be [`Helper::executeOperation()`](https://github.com/webonyx/graphql-php/blob/fe747025212b4a93fe81caf91b2b55f04ae57a15/src/Server/Helper.php#L199-L209) and [`Helper::executeBatch()`](https://github.com/webonyx/graphql-php/blob/fe747025212b4a93fe81caf91b2b55f04ae57a15/src/Server/Helper.php#L221-L238), but of course that would need more investigation/planning.

What I openly admit I’ve not investigated at this stage is how much customisation the module currently offers this early in the request lifecycle (with injector, extensions etc). I’m just floating ideas!

I have a in-development project that ideally needs batch query support, and in the new year I’ll be starting work on a component that I think is an ideal candidate for GraphQL file uploads, so I’d be looking to carry out this work myself. I think the next step would be for me to actually start work on this, it should then become a bit clearer what the impacts of any changes would be. If you disagree that a refactor like this would be a good idea that’s fine, I’ve got planned workarounds for both the features I need anyway.

unclecheese:
Awesome. Thanks for writing all this up. I'll go through it in detail as I pick through all these remaining issues on the path to v4 stable. 

This does make me really glad we abstracted away the request handler. This could potentially be done without breaking changes.

kinglozzer:
I’ll close this PR for now as I don’t think it’s really up to scratch. If the UK is indeed in lockdown after Xmas I may have a crack at some of this just for something to do!

Issue #403, opened 2021-09-02
Schema does not provide any safeguards on duplicate model names


unclecheese:
With the new, much more forgiving naming strategy of `ClassInfo::short_name()` versus `VendorNameModelName` in graphql3, there is a much greater chance of duplicate type names, and we should be encouraging the use of `type_mapping` in this case. 

Common example is multiple `Page` types (the unnamespaced one and the namespaced one)

unclecheese:
Fix: https://github.com/silverstripe/silverstripe-graphql/pull/421

PR #421, opened 2021-12-20
BUGFIX: Throw on duplicate derived model names

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-graphql/issues/403


kinglozzer:
I get an infinite loop somewhere when running a (web) `dev/build?flush` after this change - I’ve narrowed it down to this PR.

I _do_ have a duplicate derived name, which I was working around with this:

```yml
  schemas:
    admin:
      config:
        modelConfig:
          DataObject:
            type_mapping:
              App\Model\Page: AppPage
```

I tried removing that, but I still get an infinite loop rather than a warning about duplicate names 😞

unclecheese:
Can you move that type_mapping to the main schema config, eg config.yml and use "typeMapping" instead?

kinglozzer:
So the config in e.g. app/_graphql/config.yml? It’s for the admin schema
though, so will that work or will I have to set up a second config
directory for the admin schema?

I’ll give it a shot tomorrow, thanks!

kinglozzer:
Okay that does fix it:

```yml
  schemas:
    admin:
      config:
        typeMapping:
          App\Model\Page: AppPage
```

I’ll create an issue for the infinite loop thing. Thanks and Merry Christmas @unclecheese 🎄

Issue #389, opened 2021-06-04
Schema builder should have varying levels of verbosity


unclecheese:
We have one entry point for outputting to the console at the moment, `Schema::message()`, which is only configurable though `Schema::setVerbose()`.

Let's use a proper logger and allow the task to take a verbosity level, so we can have silent builds when we need them and get more detailed messages when debugging.

unclecheese:
Pull request: https://github.com/silverstripe/silverstripe-graphql/pull/422

PR #422, opened 2021-12-20
NEW: Schema logger, set verbosity

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-graphql/issues/389

Future -- add more logging throughout the entire build process for operations at the `DEBUG` level.

### DIFF
--- a/src/Schema/Logger.php
+++ b/src/Schema/Logger.php
@@ -1,0 +1,178 @@
+<?php
+
+
+namespace SilverStripe\GraphQL\Schema;
+
+use Psr\Log\LoggerInterface;
+use SilverStripe\Control\Director;
+use SilverStripe\Core\Injector\Injectable;
+
+class Logger implements LoggerInterface
+{
+    use Injectable;
+
+    const BLACK = "\033[30m";
+    const RED = "\033[31m";
+    const GREEN = "\033[32m";
+    const YELLOW = "\033[33m";
+    const BLUE = "\033[34m";
+    const MAGENTA = "\033[35m";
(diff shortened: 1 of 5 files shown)


src/Schema/Logger.php
```diff
@@ -0,0 +1,178 @@
+<?php
+
+
+namespace SilverStripe\GraphQL\Schema;
+
+use Psr\Log\LoggerInterface;
+use SilverStripe\Control\Director;
+use SilverStripe\Core\Injector\Injectable;
+
+class Logger implements LoggerInterface
```
michalkleiner, line 10: I could take a guess why you went with a custom logger class, but could we avoid adding a new logger and leverage whatever devs/DevOps folks set up for their environment?
Or is there something I'm missing/now knowing why this is needed?

unclecheese: Yeah, in the initial phase of development a year ago, I attempted to use a proper `LoggerInfterface` implementation and I just found it impossible. So many layers of abstraction, and all I wanted to do was just print some text on the console. I'd love to see how it's done properly, if you have an idea. :)

I'm also being a bit permissive with stuff like this simply because it only runs in the build task. Most of the code in graphql never executes at request time.

michalkleiner: Alright, that's all good.

As long as it can be replaced via Injector (should be when creating it through `::singleton`) and it's named in a way that makes it clear what it is, then fair enough -> `SchemaBuildLogger`? Not a major though, was just concerned about adding yet another thing doing the same thing but different etc. The same Steve or Max would probably raise :-D

PR #423, opened 2021-12-20
MINOR: change graphql codegen dir, force clear

unclecheese:
* Change `.graphql/` to `.graphql-codegen/`
* Force `clear = 1` because the new dir will hose your build without that. Caching is unreliable anyway.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,1 @@
-tests/Schema/.graphql
+tests/Schema/.graphql-codegen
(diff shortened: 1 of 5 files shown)


michalkleiner:
The folder name is better with something added to it, but it's not clear whether it generates something or whether it is generated, so how about `.graphql-generated` or `.gql-code-generated` or similar?

Happy to have `clear=1` as the default.

unclecheese:
`.graphql-generated` it is!

michalkleiner:
@unclecheese you probably didn't push some change here as the dir name is not the right one.

Issue #367, opened 2021-02-23
RFC: Model bulk loading


unclecheese:
## Summary

In the interest of the 80% case that just want "a content API", we should make exposing a group of models with common configurations as simple as possible.

```yaml
schemas:
  default:
    moduleLoader:
       - 'MyProject\Models\*'
```

## Proposed implementation

This would hinge on two new abstraction layers:

* `ModelCollectorInterface`
* `ModelFilterInterface`

Implementations for collections might include:

* Filesystem path matching (avoid ClassLoader madness)
* `subclassesFor` matching
* `hasExtension` matching
* etc..

Implementations for filters might include:

* Filesystem path matching
* `is_subclass_of` matching
* Exact matching (give me everything that's versioned except File)

## Example

```yaml
schemas:
  default:
    moduleCollectors:
      namespaceLoader:
        include: [ 'MyProject\Models\*' ]
        filters:
          matchFilter: [ 'My\Poject\Models\MySpecialModel' ]
      filesystemLoader:
        include:
          -  'app/src/**/Models' 
          -  'somevendor/somemodule: src/Models/*'
        filters:
          subclassFilter: [ 'MyApp\Test\TestObject' ]
      extensionLoader:
        include: ['SilverStripe\Versioned\Versioned']
```

Granted, most of the time this would be pretty straightforward -- subclass of DataObject collector, filter a few exact matches, but there's a good case to make it pretty flexible.

michalkleiner:
Yep, looks sensible to me. I'd add an `exclude` mechanism as well so that the include part would create the set and then exclude would remove from it. Or add a `not: true` negation.
Would also want to allow multiple instances of the same loader, so they wouldn't be able to be the keys of the list of collectors.

LiamKearn:
+1 for this. Could be really really nice QoL.

unclecheese:
@michalkleiner Working on this now and trying to find a reason why you'd need multiple instances of the same loader if they allow `include` and `exclude` as arrays.

michalkleiner:
@unclecheese didn't have a particular use case in mind but everytime I see a class being the key of an array, it triggers a warning light for me because it means it can only be used once.

The difference would only be this (note the extra dashes):
```
schemas:
  default:
    modelCollectors:
      - namespaceLoader:
          include: [ 'MyProject\Models\*' ]
          filters:
            matchFilter: [ 'My\Poject\Models\MySpecialModel' ]
      - filesystemLoader:
          include:
[...]

PR #432, opened 2022-01-10
NEW: Bulk loader

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-graphql/issues/367
Docs: https://github.com/silverstripe/silverstripe-framework/pull/10203

```
elemental: # arbitrary key to define what this set of loaders is doing
  load:
    inheritanceLoader:
      include:
        - DNADesign\Elemental\Models\BaseElement
  apply:
    fields:
      '*': true
    operations:
      read: true
      readOne: true
app:
  load:
    namespaceLoader:
      include:
        - 'App\Models\*'
      exclude:
        - 'App\Models\Secret\*'
  apply:
    fields:
      '*': true
    operations:
      '*': true
```


src/Schema/BulkLoader/CollectionFactory.php
```diff
@@ -0,0 +1,22 @@
...
+{
+    /**
+     * @param string $service
+     * @param array $params
+     * @return Collection
+     * @throws Exception
+     */
+    public function create($service, array $params = []): Collection
+    {
+        $loader = require(BASE_PATH . '/vendor/autoload.php');
```
kinglozzer, line 19: I think this will only work for valid PSR-4 classes, or classes manually loaded in composer.json? Annoyingly we still have our own class loader that picks up classes anywhere :/

https://github.com/silverstripe/silverstripe-framework/blob/8a73a4ad22c993e5fef99a6e8e96854437f68345/src/Core/Manifest/ClassLoader.php#L76-L79

unclecheese: Yeah, found that out the hard way. There are several classes it can't find. The only reason we need this is to get the file path for classes, so we can do a `filesystemLoader`, e.g. `include: [src/Models/*]`.

Wondering:

* Is that something we even want?
* Is there an alternative way to get a filename for a class?

michalkleiner: @unclecheese https://www.php.net/manual/en/reflectionclass.getfilename.php

kinglozzer: I’m not sure, I think we probably _do_ want that because I know some devs don’t namespace their app code at all - presumably a path-based `include:` would be the only way to bulk load those classes?

Reflection is an option. Alternatively we could grab the `ClassLoader` from `CoreKernel`, from there get a list of `ClassManifest` instances (I think there’s typically only ever one instance?) and then use [`ClassManifest::getItemPath($class)`](https://github.com/silverstripe/silverstripe-framework/blob/8a73a4ad22c993e5fef99a6e8e96854437f68345/src/Core/Manifest/ClassManifest.php#L333-L346).

If this is only called during schema build, Reflection is probably the better option - a slight performance hit is acceptable during builds

unclecheese: Yeah I ended up finding the relflection option yesterday and went with that approach in the new `filepathLoader` option. Thanks!

src/Schema/BulkLoader/AbstractBulkLoader.php
```diff
@@ -0,0 +1,96 @@
+<?php
+
+
```
emteknetnz, line 3: Remove extra blank line

src/Schema/BulkLoader/AbstractBulkLoader.php
```diff
@@ -0,0 +1,96 @@
...
+namespace SilverStripe\GraphQL\Schema\BulkLoader;
+
+use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\GraphQL\Schema\Exception\SchemaBuilderException;
+use SilverStripe\GraphQL\Schema\Interfaces\ConfigurationApplier;
+use SilverStripe\GraphQL\Schema\Interfaces\Identifiable;
+use SilverStripe\GraphQL\Schema\Schema;
+
+/**
+ * Provides base functionality to all bulk loaders. Must define a collect() method.
```
emteknetnz, line 13: `Must define a collect() method.` - This is a bit confusing since there's collect is a `public function` while we also have `abstract public static function getIdentifier(): string;` which does actually need to be defined

src/Schema/BulkLoader/BulkLoaderSet.php
```diff
@@ -0,0 +1,129 @@
+<?php
+
+
```
emteknetnz, line 3: Remove extra blank line

src/Schema/BulkLoader/BulkLoaderSet.php
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        return $collection;
+    }
+
+    /**
+     * @param $loaders
+     * @return $this
+     */
+    public function setLoaders($loaders): self
```
emteknetnz, line 106: $loaders should have a type of array?

src/Schema/BulkLoader/ExtensionLoader.php
```diff
@@ -0,0 +1,89 @@
+<?php
+
+
```
emteknetnz, line 3: Remove extra blank line

src/Schema/BulkLoader/FilepathLoader.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+
```
emteknetnz, line 3: Remove extra blank line

src/Schema/BulkLoader/InheritanceLoader.php
```diff
@@ -0,0 +1,90 @@
+<?php
+
+
```
emteknetnz, line 3: Remove extra blank line

(and do same on all subsequent files)

src/Schema/BulkLoader/ExtensionLoader.php
```diff
@@ -0,0 +1,89 @@
...
+use SilverStripe\Core\Extensible;
+use InvalidArgumentException;
+use SilverStripe\Core\Extension;
+
+/**
+ * Loads classes that have a given extension assigned to them.
+ */
+class ExtensionLoader extends AbstractBulkLoader
+{
+    const IDENTIFIER = 'extensionLoader';
```
emteknetnz, line 15:
```suggestion
    public const IDENTIFIER = 'extensionLoader';
```
Also update it for all subsequent IDENTIFIER's in following files

src/Schema/BulkLoader/ExtensionLoader.php
```diff
@@ -0,0 +1,89 @@
...
+                    $class,
+                    static::class
+                ));
+            }
+        }
+
+        return parent::include($include);
+    }
+
+    public function exclude(array $exclude): AbstractBulkLoader
```
emteknetnz, line 75: add docblock to be consistent with other methods

src/Schema/Interfaces/Identifiable.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+
+namespace SilverStripe\GraphQL\Schema\Interfaces;
+
+/**
+ * Used by any class that delcares an identifier
```
emteknetnz, line 7:
```suggestion
 * Used by any class that declares an identifier
```

emteknetnz (changes requested):
Looks good, some very minor things

src/Schema/Interfaces/Identifiable.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+
+namespace SilverStripe\GraphQL\Schema\Interfaces;
+
+/**
+ * Used by any class that declares an identifier
+ */
+interface Identifiable
```
michalkleiner, line 9: I generally support using interfaces, but the name here and that it's used to populate the registry of available loaders is a bit misleading in my opinion. Can we achieve the same looking for subclasses of `AbstractBulkLoader`?

unclecheese: The idea is that this could someday become reusable. We're using this pattern more and more where we just have a holding tank for a given set of implementations. At a bare minimum, a registry should be able to find implementations by an identifier. In practise, it should be a thing with an identifier that also implements or inherits from a given contract, e.g. `AbstractBulkLoader` or `ModelQueryPlugin`.. I have yet to figure that last part out. 

But what I'm trying to avoid is lighting up a new concretion for every single registry use case we have -- BulkLoaders, Plugins, etc... seems like a good case for abstraction.

src/Schema/BulkLoader/AbstractBulkLoader.php
unclecheese: Yeah, used to be abstract and now it's basically a noop. Have updated.

PR #433, opened 2022-01-17
NEW: Schema initialise task

unclecheese:
There's too much boilerplate in setting up a default schema. What this does:

* Creates a `_graphql` dir with `models.yml`, `types.yml`, `queries.yml`, `mutations.yml`, 
* Creates a `_config/graphql.yml` file that routes and configures your default schema (Director rule, `src:` property)
* Creates an empty `Resolvers` class where you can define custom resolvers 
* Registers the resolver class with `schemaConfig.resolvers`

### Example

`$ vendor/bin/sake dev/graphql/init namespace="MyAgency\MyApp`"

### Arguments

* **namespace**: The root app namespace (required)
* **name**: The schema name (default: `default`)
* **graphqlConfigDir**: The folder where the flushless graphql files will go (default: `_graphql/`)
* **graphqlCodeDir**: The subfolder of `src/` where your GraphQL code (the resolver class) will go. Follows PSR-4 based on the `namespace` argument (default: `GraphQL`)

### File permissions

Inherits whatever you're using on the `app/` directory.

### It's idempotent

Any of the above artefacts that already exist will not be over written, and it will tell you that.
 
## Related PRs
- silverstripe/silverstripe-framework#10218

### DIFF
--- a/_config/dev.yml
+++ b/_config/dev.yml
@@ -13,3 +13,7 @@ SilverStripe\GraphQL\Dev\DevelopmentAdmin:
       controller: SilverStripe\GraphQL\Dev\Build
       links:
         build: Build the GraphQL schema
+    init:
+      controller: SilverStripe\GraphQL\Dev\Initialise
+      links:
+        init: Initialise a GraphQL schema in your project directory (CLI only)
--- a/src/Dev/Initialise.php
+++ b/src/Dev/Initialise.php
@@ -1,6 +1,5 @@
 <?php
 
-
 namespace SilverStripe\GraphQL\Dev;
 
 use SilverStripe\Control\Controller;
@@ -8,18 +7,11 @@ use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Core\Manifest\ModuleManifest;
 use SilverStripe\Core\Path;
-use SilverStripe\Dev\DebugView;
-use SilverStripe\GraphQL\Schema\DataObject\FieldAccessor;
-use SilverStripe\GraphQL\Schema\Exception\EmptySchemaException;
-use SilverStripe\GraphQL\Schema\Exception\SchemaBuilderException;
-use SilverStripe\GraphQL\Schema\Exception\SchemaNotFoundException;
 use SilverStripe\GraphQL\Schema\Schema;
-use SilverStripe\GraphQL\Schema\SchemaBuilder;
 use SilverStripe\ORM\Connect\NullDatabaseException;
 
 /**
- * Class Initialise
- * @package SilverStripe\GraphQL\Dev
+ * A task that initialises a schema with boilerplate config and files.
  */
 class Initialise extends Controller
 {
@@ -65,7 +57,7 @@ class Initialise extends Controller
     /**
      * @var string
      */
-    private $projectDir = 'app';
+    private $projectDir = '';
 
     /**
      * @var string
@@ -75,7 +67,7 @@ class Initialise extends Controller
     /**
      * @var string
      */
-    private $perms = '0777';
+    private $perms = '';
 
     /**
      * @param HTTPRequest $request
@@ -87,15 +79,22 @@ class Initialise extends Controller
             !$isBrowser,
             'This task can only be run from CLI'
         );
+
+        if ($request->getVar('help')) {
+            $this->showHelp();
+            return;
+        }
+
         $appNamespace = $request->getVar('namespace');
-        Schema::invariant(
-            $appNamespace,
-            'Please provide a base namespace for your app, e.g. "namespace=App" or "namespace=MyVendor\MyProject"'
-        );
+
+        if (!$appNamespace) {
+            echo "Please provide a base namespace for your app, e.g. \"namespace=App\" or \"namespace=MyVendor\MyProject\".\nFor help, run \"dev/graphql/init help=1\"\n";
+            return;
+        }
+
         $this->appNamespace = $appNamespace;
 
         $this->projectDir = ModuleManifest::config()->get('project');
-
 
         $schemaName = $request->getVar('name');
         if ($schemaName) {
@@ -140,13 +139,14 @@ class Initialise extends Controller
             echo "Graphql config directory already exists. Skipping." . PHP_EOL;
             return;
         }
-            echo "Creating graphql config directory: $this->graphqlConfigDir" . PHP_EOL;
-            mkdir($absGraphQLDir, $this->perms);
+
+        echo "Creating graphql config directory: $this->graphqlConfigDir" . PHP_EOL;
+        mkdir($absGraphQLDir, $this->perms);
         foreach (['models', 'config', 'types', 'queries', 'mutations'] as $file) {
             touch(Path::join($absGraphQLDir, "$file.yml"));
         }
-            $configPath = Path::join($absGraphQLDir, 'config.yml');
-            $defaultConfig = <<<YAML
+        $configPath = Path::join($absGraphQLDir, 'config.yml');
+        $defaultConfig = <<<YAML
 resolvers:
   - $this->appNamespace\Resolvers
 YAML;
@@ -172,6 +172,7 @@ SilverStripe\GraphQL\Schema\Schema:
     $this->schemaName:
       src:
         - $this->projectDir/$this->graphqlConfigDir
+
 YAML;
         file_put_contents($absConfigFile, $defaultProjectConfig);
     }
@@ -206,16 +207,44 @@ namespace $graphqlNamespace;
  */
 class Resolvers
 {
-
-    public static function resolveMyQuery(\$obj, array \$args, \$context): array
+    public static function resolveMyQuery(\$obj, array \$args, array \$context): array
     {
        // Return the result of query { myQuery { ... } }
        return [];
     }
-
 }
 
 PHP;
         file_put_contents($resolverFile, $resolverCode);
+    }
+
+    /**
+     * Outputs help text to the console
+     */
+    private function showHelp(): void
+    {
+        echo <<<TXT
+
+****
+This task executes a lot of the boilerplate required to build a new GraphQL schema. It will
+generate a few files in your project directory. Any files that already exist will not be
+overwritten. The task can be run multiple times and is non-destructive.
+****
+
+-- Example:
+
+$ vendor/bin/sake dev/graphql/init namespace="MyAgency\MyApp"
+
+-- Arguments:
+
+[namespace]: The root namespace. Required.
+
+<name>: The name of the schema. Default: "default"
+
+<graphqlConfigDir>: The folder where the flushless graphql config files will go. Default: "_graphql"
+
+<graphqlCodeDir>: The subfolder of src/ where your GraphQL code (the resolver class) will go. Follows PSR-4 based on the namespace argument (default: "GraphQL")
+
+TXT;
     }
 }
--- a/src/Dev/Initialise.php
+++ b/src/Dev/Initialise.php
@@ -1,0 +1,221 @@
+<?php
+
+
+namespace SilverStripe\GraphQL\Dev;
+
+use SilverStripe\Control\Controller;
+use SilverStripe\Control\Director;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Core\Manifest\ModuleManifest;
+use SilverStripe\Core\Path;
+use SilverStripe\Dev\DebugView;
+use SilverStripe\GraphQL\Schema\DataObject\FieldAccessor;
+use SilverStripe\GraphQL\Schema\Exception\EmptySchemaException;
+use SilverStripe\GraphQL\Schema\Exception\SchemaBuilderException;
+use SilverStripe\GraphQL\Schema\Exception\SchemaNotFoundException;
+use SilverStripe\GraphQL\Schema\Schema;
+use SilverStripe\GraphQL\Schema\SchemaBuilder;
+use SilverStripe\ORM\Connect\NullDatabaseException;
+
+/**
+ * Class Initialise
+ * @package SilverStripe\GraphQL\Dev
+ */
+class Initialise extends Controller
+{
+    /**
+     * @var string[]
+     */
+    private static $url_handlers = [
+        '' => 'initialise'
+    ];
+
+    /**
+     * @var string[]
+     */
+    private static $allowed_actions = [
+        'initialise'
+    ];
+
+    /**
+     * @var string
+     */
+    private $appNamespace;
+
+    /**
+     * @var string
+     */
+    private $schemaName = 'default';
+
+    /**
+     * @var string
+     */
+    private $graphqlConfigDir = '_graphql';
+
+    /**
+     * @var string
+     */
+    private $graphqlCodeDir = 'GraphQL';
+
+    /**
+     * @var string
+     */
+    private $endpoint = 'graphql';
+
+    /**
+     * @var string
+     */
+    private $projectDir = 'app';
+
+    /**
+     * @var string
+     */
+    private $srcDir = 'src';
+
+    /**
+     * @var string
+     */
+    private $perms = '0777';
+
+    /**
+     * @param HTTPRequest $request
+     */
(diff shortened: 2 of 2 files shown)


src/Dev/Initialise.php
```diff
@@ -0,0 +1,221 @@
...
+/**
+ * Use this class to define custom resolvers. Static functions in this class
+ * matching the pattern resolve<FieldName> or resolve<TypeNameFieldName>
+ * will be automatically assigned to their respective fields.
+ *
+ * More information: https://docs.silverstripe.org/en/4/developer_guides/graphql/working_with_generic_types/resolver_discovery/#the-resolver-discovery-pattern
+ */
+class Resolvers
+{
+
```
emteknetnz, line 209: Remove this blank line, it'll fail PHPCS

src/Dev/Initialise.php
```diff
@@ -0,0 +1,221 @@
...
+    public static function resolveMyQuery(\$obj, array \$args, \$context): array
+    {
+       // Return the result of query { myQuery { ... } }
+       return [];
+    }
+
```
emteknetnz, line 215: May as well remove this blank line too

src/Dev/Initialise.php
```diff
@@ -0,0 +1,221 @@
...
+        Schema::invariant(
+            $appNamespace,
+            'Please provide a base namespace for your app, e.g. "namespace=App" or "namespace=MyVendor\MyProject"'
+        );
+        $this->appNamespace = $appNamespace;
+
+        $this->projectDir = ModuleManifest::config()->get('project');
+
+
+        $schemaName = $request->getVar('name');
```
emteknetnz, line 100: You need to put docs for CLI arguments somewhere more accessible to end-users i.e. not the PR description.

src/Dev/Initialise.php
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /**
+     * Creates the SS config in _config/graphql.yml
+     */
+    private function createProjectConfig(): void
+    {
+        $absConfigFile = Path::join(BASE_PATH, $this->projectDir, '_config', 'graphql.yml');
+        if (file_exists($absConfigFile)) {
+            echo "Config file $absConfigFile already exists. Skipping." . PHP_EOL;
```
emteknetnz, line 163: Would it be better to suffix the config if the graphql.yml file already exists?

src/Dev/Initialise.php
```diff
@@ -0,0 +1,221 @@
...
+    private function createGraphQLConfig(): void
+    {
+        $absGraphQLDir = Path::join(BASE_PATH, $this->projectDir, $this->graphqlConfigDir);
+        if (is_dir($absGraphQLDir)) {
+            echo "Graphql config directory already exists. Skipping." . PHP_EOL;
+            return;
+        }
+            echo "Creating graphql config directory: $this->graphqlConfigDir" . PHP_EOL;
+            mkdir($absGraphQLDir, $this->perms);
+        foreach (['models', 'config', 'types', 'queries', 'mutations'] as $file) {
```
emteknetnz, line 145: indentation in this method is all over the place

src/Dev/Initialise.php
```diff
@@ -0,0 +1,221 @@
...
+            touch(Path::join($absGraphQLDir, "$file.yml"));
```
emteknetnz, line 146: I don't think creating a bunch of empty files adds much value.  I think it's fine having only the config.yml file here.

src/Dev/Initialise.php
```diff
@@ -0,0 +1,221 @@
...
+
+    /**
+     * @var string
+     */
+    private $srcDir = 'src';
+
+    /**
+     * @var string
+     */
+    private $perms = '0777';
```
emteknetnz, line 78: I think values like this and $projectDir, which will get dynamically set during the `initialise()` method, are better off being left as blank strings. It's confusing having '0777' whens it's never used as it just get changed straight away

src/Dev/Initialise.php
```diff
@@ -0,0 +1,221 @@
...
+use SilverStripe\GraphQL\Schema\Exception\EmptySchemaException;
+use SilverStripe\GraphQL\Schema\Exception\SchemaBuilderException;
+use SilverStripe\GraphQL\Schema\Exception\SchemaNotFoundException;
+use SilverStripe\GraphQL\Schema\Schema;
+use SilverStripe\GraphQL\Schema\SchemaBuilder;
+use SilverStripe\ORM\Connect\NullDatabaseException;
+
+/**
+ * Class Initialise
+ * @package SilverStripe\GraphQL\Dev
```
emteknetnz, line 22: Do we even need this Class / @package docblock anymore?  I can't image there's any IDE's out that there can't figure this out based on the actual namespace and the class name

src/Dev/Initialise.php
```diff
@@ -0,0 +1,221 @@
+<?php
+
+
```
emteknetnz, line 3: Remove blank line

src/Dev/Initialise.php
```diff
@@ -0,0 +1,221 @@
...
+        $appNamespace = $request->getVar('namespace');
+        Schema::invariant(
+            $appNamespace,
+            'Please provide a base namespace for your app, e.g. "namespace=App" or "namespace=MyVendor\MyProject"'
+        );
+        $this->appNamespace = $appNamespace;
+
+        $this->projectDir = ModuleManifest::config()->get('project');
+
+
```
emteknetnz, line 99: Remove blank line

src/Dev/Initialise.php
```diff
@@ -0,0 +1,221 @@
...
+    {
+        $isBrowser = !Director::is_cli();
+        Schema::invariant(
+            !$isBrowser,
+            'This task can only be run from CLI'
+        );
+        $appNamespace = $request->getVar('namespace');
+        Schema::invariant(
+            $appNamespace,
+            'Please provide a base namespace for your app, e.g. "namespace=App" or "namespace=MyVendor\MyProject"'
```
emteknetnz, line 93: extremely ugly user experience

```
www-data@www:/var/www$ vendor/bin/sake dev/graphql/init 
ERROR [Emergency]: Uncaught SilverStripe\GraphQL\Schema\Exception\SchemaBuilderException: Please provide a base namespace for your app, e.g. "namespace=App" or "namespace=MyVendor\MyProject"
IN GET dev/graphql/init
Line 1226 in /var/www/vendor/silverstripe/graphql/src/Schema/Schema.php                                                                                                             

Source
======
  1217:     * @param $test
  1218:     * @param string $message
  1219:     * @param mixed ...$params
  1220:     * @throws SchemaBuilderException
  1221:     */
  1222:    public static function invariant($test, $message = '', ...$params): void
  1223:    {
  1224:        if (!$test) {
  1225:            $message = count($params) > 0 ? sprintf($message, ...$params) : $message;
* 1226:            throw new SchemaBuilderException($message);
  1227:        }
  1228:    }
  1229:}

Trace
=====
SilverStripe\GraphQL\Schema\Schema::invariant(, Please provide a base namespace for your app, e.g. "namespace=App" or "namespace=MyVendor\MyProject")
Initialise.php:93

SilverStripe\GraphQL\Dev\Initialise->initialise(SilverStripe\Control\HTTPRequest)
RequestHandler.php:323

SilverStripe\Control\RequestHandler->handleAction(SilverStripe\Control\HTTPRequest, initialise)
Controller.php:286
```

(very long strack trace follows)

_config/dev.yml
```diff
@@ -13,3 +13,7 @@ SilverStripe\GraphQL\Dev\DevelopmentAdmin:
       controller: SilverStripe\GraphQL\Dev\Build
       links:
         build: Build the GraphQL schema
+    init:
```
emteknetnz, line 16: Is there any docs that links to `vendor/bin/sake dev/graphql/init`?  Right now it's very hidden

src/Dev/Initialise.php
```diff
@@ -0,0 +1,221 @@
...
+<?php
+
+namespace $graphqlNamespace;
+
+/**
+ * Use this class to define custom resolvers. Static functions in this class
+ * matching the pattern resolve<FieldName> or resolve<TypeNameFieldName>
+ * will be automatically assigned to their respective fields.
+ *
+ * More information: https://docs.silverstripe.org/en/4/developer_guides/graphql/working_with_generic_types/resolver_discovery/#the-resolver-discovery-pattern
```
emteknetnz, line 205: Lines too long, will fail PHPCS

src/Dev/Initialise.php
```diff
@@ -0,0 +1,221 @@
...
+ */
+class Resolvers
+{
+
+    public static function resolveMyQuery(\$obj, array \$args, \$context): array
```
emteknetnz, line 210: $context has no type - shouldn't it have something defined?

src/Dev/Initialise.php
```diff
@@ -0,0 +1,221 @@
...
+            $defaultProjectConfig = <<<YAML
+SilverStripe\Control\Director:
+  rules:
+    $this->endpoint: '%\$SilverStripe\GraphQL\Controller.$this->schemaName'
+SilverStripe\GraphQL\Schema\Schema:
+  schemas:
+    $this->schemaName:
+      src:
+        - $this->projectDir/$this->graphqlConfigDir
+YAML;
```
emteknetnz, line 175: Needs a blank line at end of file

emteknetnz (changes requested):
Concept seems OK?  Feels like it needs more work around CLI documentation and experience

src/Dev/Initialise.php
unclecheese: Have added a `help` option

src/Dev/Initialise.php
unclecheese: Ideally, this would do a similar check that the GraphQL devtools controller does, where it finds what schemas are actually routed. That's in a separate module, and it's got some GraphQL 3 backward compat stuff in it.

I'm thinking that if you have a `graphql.yml` file, you're past the point of needing this tool? I'd rather be paranoid of clobbering someone's existing config.

src/Dev/Initialise.php
unclecheese: I actually think it does. There's a lot of naming conventions to learn in GraphQL 4, and this just saves a trip to the docs. At a bare minimum, we need `config.yml` and `models.yml`, but really, I'd prefer to just do empty files. You're going to need them at some point, and for users who don't know about custom types/queries, it creates a nice footprint that shows you a bit of what the module can do without forcing you down the documentation path.

src/Dev/Initialise.php
unclecheese: Yup, fair enough.

src/Dev/Initialise.php
unclecheese: Changed to standard output.

src/Dev/Initialise.php
unclecheese: Yeah, good point. I think it's always an array? Hope it's not nullable. :)

src/Dev/Initialise.php
unclecheese: The link itself is enough to break the limit. What would you recommend?

unclecheese:
Docs are here: https://github.com/silverstripe/silverstripe-framework/pull/10218

_config/dev.yml
unclecheese: https://github.com/silverstripe/silverstripe-framework/pull/10218

unclecheese:
Revisions made per feedback.

src/Dev/Initialise.php
michalkleiner: Having worked with GQL4 I tend to agree that any help in code/prebaked configs is closer to the hands-on dev and therefore better than having to go and read docs to then do the same thing anyway. Even from the perspective of naming the files and what they may contain.

src/Dev/Initialise.php
GuySartorelli: I think this is a good way of handling it - if the file already exists, we inform the developer of that fact and skip it. They can then make a judgement call to either keep the existing schema definition or delete it and run this again.

maxime-rainville:
I'm wondering if maybe we want to move this to the GraphQL dev tools module instead? That way we can just call it experimental and merge it.

GuySartorelli:
I don't necessarily think that we should push it off to another module just to get it merged if it's something that we think _should_ be in core at the end of the day... that said, this _is_ ultimately something that helps at dev-time and shouldn't be used in a production environment (if I understand it correctly) so dev tools does genuinely seem to be a better place for it.

unclecheese:
I don't see it as a good fit for dev tools. I'll try to explain why.

There are essentially two use cases for this module:

* Make the CMS work
* Provide a bespoke GraphQL schema for my app

In the case of the latter, your entry point is now creating a bunch of boilerplate yaml before you can request a single byte of data. The time when this feature is most useful is during that time. Your journey will start with the docs or a Slack message to get the tl;dr on how you do it . If we can deliver a single CLI task to get you into your schema in a few minutes rather than hours, we should not pass on the opportunity to provide that value. With devtools, by the time you've got your hands dirty and are seeking the features of that module, it's too late for this to be effective. 

You could certainly make the case that if you're building a bespoke GraphQL server in Silverstripe CMS, you're highly unlikely to do it without the devtools module (imagine building a schema and not testing out any queries in the IDE!), and this is just another reinforcement of the criticality of that module when doing a custom GraphQL schema, but I think that speaks more to the case for renaming the devtools module to "silverstripe-graphql-custom-schema" or something -- in other words, a specific module that provides tools to break GraphQL out of the conceptual boundaries of the CMS, rather than framing it as a power-user addon.

I'm more wondering what's left to do on this PR? It's been stale for months, which is fine, but I don't think the fact that its been sitting for a while should be a pretext for changing its intended domain.

GuySartorelli:
> Your journey will start with the docs or a Slack message to get the tl;dr on how you do it .
> ...
> With devtools, by the time you've got your hands dirty and are seeking the features of that module, it's too late for this to be effective.

This seems to me like it's ultimately a matter of documentation - if the documentation says something along the lines of "An easy way to get started creating your schema is using the schema initialise task in the devtools module" with a link to that module, then it brings not only this functionality but also the dev tools module in general to the attention of people who are just starting out with making their custom graphql schema.

> I'm more wondering what's left to do on this PR? 

There were a few discussion points that someone just needs to make a call on, I think - maybe @emteknetnz could take another look now that his questions have been answered.

> I don't think the fact that its been sitting for a while should be a pretext for changing its intended domain.

Agreed - but now that the question has been asked I think it does behoove us to decide if this code might be better suited to devtools. If the only reason for it not to be there is that we want to make people aware of it, I think that's a matter of documentation as I mentioned above. In general it would be better to keep code intended for use at development time in a module intended to be a dev dependency.

unclecheese:
The thing is, 90% of the code in this module is _not intended for production_. The whole thing is a giant task that creates an artefact that _is_ intended for production. So shipping dev code is part of the purchase at this point.

Reading a "Quick Start" documentation for a module and getting told that you need to install another module to do the quick start is probably a good indication that we've crippled the DX. I just don't see any positive outcomes from that.

maxime-rainville:
Yeah ... to me this sounds more like a documentation concern. Having a quick way to scaffold a graphql schema is pointless if you don't also have a quick way to run a dummy query against it. The doc should push you to install the dev tools early.

On a side note, we probably need to start providing more formal support for those dev-tools.

unclecheese:
In theory, the devtools module shouldn't have to exist. The graphql module is itself, for devs. It produces a single deployable artefact. Really everything but the controller could be left out of the deployment. So it makes sense that the IDE and all of its furnishings should ship with the module.

It was this way for a short time, until we realised that GraphQL Playground came with massive JS bundle that bloated this repository. We decided to prune the commits and move the playground to a separate module. That minor detail is what put us on this path of two separate modules, and it's never really felt right.

So I guess I'm fine with it either way. I think I'm just a bit annoyed that devtools has to exist. It feels like unnecessary documentation and maintenance overhead. But here we are. 🤷‍♂️

emteknetnz:
Closed in favor of https://github.com/silverstripe/silverstripe-graphql-devtools/pull/53

Issue #443, opened 2022-02-14
Unknown graphql type: Query when rebuilding schema after deleting .graphql-generated


maxime-rainville:
# Affects
- GraphQL v4

# Steps
- Create a brand new project and config your env
- Run a `sake dev/build`
- Schema gets build normally
- run `rm -r .graphql-generated/`
- Run `sake dev/graphql/build`

Expected: results schema is regenerated
Actual result: Exception is thrown on [AbstractTypeRegistry.php:55](https://github.com/silverstripe/silverstripe-graphql/blob/0fb4768d0b87ccc739cc3ddb86a45499205f2917/src/Schema/Storage/AbstractTypeRegistry.php#L55).

```
[INFO]: --- Building schema "admin" ---
[INFO]: Creating new schema
[INFO]: Migrating new schema
[INFO]: Deleting temp schema
[INFO]: Total types: 59
[INFO]: Types built: 0
[INFO]: Types deleted: 0
ERROR [Emergency]: Uncaught Exception: Unknown graphql type: Query
IN GET dev/graphql/build
Line 55 in /home/max/Projects/innovation-week/vendor/silverstripe/graphql/src/Schema/Storage/AbstractTypeRegistry.php

Source
======
  46:                  $cls = static::getSourceNamespace() . '\\' . $obfuscatedName;
  47:                  if (class_exists($cls)) {
  48:                      $type = new $cls();
  49:                  }
  50:              }
  51:              static::$types[$typename] = $type;
  52:          }
  53:          $type = static::$types[$typename];
  54:          if (!$type) {
* 55:              throw new Exception("Unknown graphql type: " . $typename);
  56:          }
  57:          return $type;
  58:      }
  59:  
  60:      public static function ID(): ScalarType
  61:      {

Trace
=====
SilverStripe\GraphQL\Schema\Storage\AbstractTypeRegistry::fromCache
AbstractTypeRegistry.php:24

SilverStripe\GraphQL\Schema\Storage\AbstractTypeRegistry::get
__graphql-schema.php:59

SSGraphQLSchema_21232f297a57a5a743894a0e4a801fc3\Types::Query

call_user_func
CodeGenerationStore.php:324

SilverStripe\GraphQL\Schema\Storage\CodeGenerationStore->getSchema
SchemaBuilder.php:70

SilverStripe\GraphQL\Schema\SchemaBuilder->getSchema
SchemaTranscribeHandler.php:20

[...]

PR #444, opened 2022-02-14
BUG Always nuke cache when manually rebuilding schema

maxime-rainville:
There's no use case for not clearing the cache when rebuilding the schema

# Parent issue 
- [ ] https://github.com/silverstripe/silverstripe-graphql/issues/443


unclecheese:
In addition to many other bizarre errors. The cache is 95% effective, but the remaining 5% is pure pain. Need to spend some time down the track figuring out all the conditions that defeat it.

PR #447, opened 2022-02-24
Fix server error if not authenticated

lekoala:
For context : https://github.com/silverstripe/silverstripe-framework/issues/9978

Here, my proposal is to return a consistent json response

### DIFF
--- a/src/Extensions/IntrospectionProvider.php
+++ b/src/Extensions/IntrospectionProvider.php
@@ -26,8 +26,9 @@ class IntrospectionProvider extends Extension
         try {
             $manager = $this->owner->getManager();
         } catch (Exception $ex) {
-            if ($ex->getMessage() == "Authentication required") {
-                return (new HTTPResponse(json_encode(["error" => "Authentication required"]), 401))
+            $message = $ex->getMessage();
+            if ($message == "Authentication required" || $message == "Not authorised") {
+                return (new HTTPResponse(json_encode(["error" => $message]), 401))
                     ->addHeader('Content-Type', 'application/json');
             }
             throw $ex;
--- a/src/Extensions/IntrospectionProvider.php
+++ b/src/Extensions/IntrospectionProvider.php
@@ -27,8 +27,12 @@ class IntrospectionProvider extends Extension
             $manager = $this->owner->getManager();
         } catch (Exception $ex) {
             $message = $ex->getMessage();
-            if ($message == "Authentication required" || $message == "Not authorised") {
+            if ($message == "Authentication required") {
                 return (new HTTPResponse(json_encode(["error" => $message]), 401))
+                    ->addHeader('Content-Type', 'application/json');
+            }
+            if ($message == "Not authorised") {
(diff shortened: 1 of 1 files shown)


GuySartorelli:
I'd hesitate to just catch all `Exception`s - if something does actually fail, this will now mean there's no log of the failure other than a context-less 500 response in access logs.

lekoala:
I can improve the pr to catch the specific message if needed

src/Extensions/IntrospectionProvider.php
```diff
@@ -22,7 +23,16 @@ class IntrospectionProvider extends Extension
      */
     public function types(HTTPRequest $request)
     {
-        $manager = $this->owner->getManager();
+        try {
+            $manager = $this->owner->getManager();
+        } catch (Exception $ex) {
+            if ($ex->getMessage() == "Authentication required") {
+                return (new HTTPResponse(json_encode(["error" => "Authentication required"]), 500))
```
GuySartorelli, line 30: 500 isn't the right response code here.
```suggestion
                return (new HTTPResponse(json_encode(["error" => "Authentication required"]), 401))
```

GuySartorelli (changes requested):
The exception being caught is from `SilverStripe\GraphQL\Controller::getRequestUser()`:
https://github.com/silverstripe/silverstripe-graphql/blob/b51ce72bfc3508cda7fbbfb3a6587b50f09263c6/src/Controller.php#L439-L447
The exceptions were originally added to `index()` in #57.

We should probably also treat the "Not authorised" exception message in the same way.

In general I'm okay with this. If you're happy to make a change to check for the "Not authorised" message, and change the response code (401 for "Authentication required" and 403 for "Not authorised"), I'll be happy to approve it. I'll probably ask someone more familiar with GraphQL to review it before I'd be comfortable merging.

Note that while @unclecheese did say it shouldn't require authentication at all in the issue, I don't have enough background on why the authentication checks were added in the first place, so I'm more comfortable with this approach.

lekoala:
> We should probably also treat the "Not authorised" exception message in the same way.

I don't think that's strictly necessary. I don't mind having logged in users get that kind of error page. My issue is that public endpoints can be visited by anyone (bots for example) and trigger errors that get logged for instance.

I think the current design works ok as long as the public use case is covered.

I've changed the error code as you requested.

GuySartorelli:
Sorry it's taken me so long to get back to this.
Please treat the "Not authorised" exception message the same way for consistency.
Please also target this to `3.8` so it can be released as a patch.

Once that's done I'll be happy to merge.

Alternatively if you don't plan on continuing with this PR please close it.

lekoala:
@GuySartorelli done

GuySartorelli:
Thanks for that. Unfortunately as mentioned in https://github.com/silverstripe/silverstripe-graphql/pull/447#pullrequestreview-935897300 the two scenarios news to return differen HTTP code (specifically 401 for "Authentication required" and 403 for "Not authorised")

Can you please make that change?

lekoala:
sorry i missed that one
done

Issue #446, opened 2022-02-16
Logging warning in "ondemand" GraphQL v4 schema build


maxime-rainville:
If you're `.graphql-generated` folder does not exist when you try calling `admin/graphql`, Silverstripe will attempt to regenerate it "on demand". That works pretty well, except for the fact that the output that's meant to go to the console ends up in the HTTP Response.

```
[INFO]: Creating new schema<br>[INFO]: Migrating new schema<br>[INFO]: Deleting temp schema<br>[INFO]: Total types: 59<br>[INFO]: Types built: 59<br>[INFO]: *CopyToStageInputType<br>[INFO]: *VersionedInputType<br>[INFO]: *FolderInput<br>[INFO]: *FileFilterInput<br>[INFO]: *FileUsage<br>[INFO]: *DescendantFileCount<br>[INFO]: *PublicationNotice<br>[INFO]: *Folder<br>[INFO]: *File<br>[INFO]: *FileInput<br>[INFO]: (... and 49 more)<br>[INFO]: Types deleted: 0<br>
```

We probably need need to redirect those logs to to STDERR or something for "on demand" request.

# Pull request
- [ ] https://github.com/silverstripe/silverstripe-graphql/pull/448

maxime-rainville:
Part of the problem is that we're using a new [`SilverStripe\GraphQL\Schema\Logger`](https://github.com/silverstripe/silverstripe-graphql/blob/master/src/Schema/Logger.php)

I get the impression this should only be used when we are on the `dev/graphql/schema` route. The rest of the time, we should probably used one of the option defined in [`silverstripe/framwork:_config/logging.yml`](https://github.com/silverstripe/silverstripe-framework/blob/4/_config/logging.yml)?

Maybe, we shouldn't even be using `SilverStripe\GraphQL\Schema\Logger` at all.

I'm thinking we define a new `Psr\Log\LoggerInterface.graphql` injector key and assign it something different base on context. Then we replace all instances of `SilverStripe\GraphQL\Schema\Logger::singleton()` with `Injector::get()->instance(LoggerInterface::class . '.graphql')`

emteknetnz:
Seems fine, why would we not just use the regular logger on the `dev/graphql/schema` route as well? i.e. what is `SilverStripe\GraphQL\Schema\Logger` even required for?

michalkleiner:
[...]

PR #448, opened 2022-03-09
Only use the custom Logger when explicitely running a build task

maxime-rainville:
The logger is echoing build info statement when building the schema on-demand 

# Parent issues
- https://github.com/silverstripeltd/product-issues/issues/519
- #446


_config/dev.yml
```diff
@@ -13,3 +13,7 @@ SilverStripe\GraphQL\Dev\DevelopmentAdmin:
       controller: SilverStripe\GraphQL\Dev\Build
       links:
         build: Build the GraphQL schema
+    clear:
```
maxime-rainville, line 16: I've added a "clear" task that nukes the `.graphql-generated` folder. Maybe we want to keep this around in the long run ... maybe not.

_config/logging.yml
```diff
@@ -0,0 +1,8 @@
+---
+Name: graphql-logging
+after: '#logging'
+---
+
+SilverStripe\Core\Injector\Injector:
+  # This will be overridden by SilverStripe\GraphQL\Schema\Logger in SilverStripe\GraphQL\Dev\DevAdmin
+  Psr\Log\LoggerInterface.graphql-build: '%$Psr\Log\LoggerInterface.errorhandler'
```
maxime-rainville, line 8: By default, build statement will now go to the standard error handler. Build task will need to override this excitedly to send their build statement to the Logger.

src/Dev/Build.php
```diff
@@ -42,15 +44,8 @@ public function build(HTTPRequest $request)
             echo "<div class=\"build\">";
         }
         $clear = true;
-        $verbosity = strtoupper($request->getVar('verbosity') ?? 'INFO');
```
maxime-rainville, line 45: Had to remove the verbosity because the default logger doesn't support this. Since the scope where Logger is used has been substantially curtail by this PR, this seems reasonable.

src/Dev/Clear.php
```diff
@@ -0,0 +1,40 @@
...
+
+use Psr\Log\LoggerInterface;
+use SilverStripe\Control\Controller;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Core\Injector\Injector;
+
+use SilverStripe\GraphQL\Schema\Storage\CodeGenerationStore;
+use Symfony\Component\Filesystem\Filesystem;
+
+class Clear extends Controller
```
maxime-rainville, line 12: This is just for testing ... but maybe we want to keep it around. Not sure.

src/Dev/DevelopmentAdmin.php
```diff
@@ -42,6 +45,11 @@ protected function init()
             Security::permissionFailure($this);
             return;
         }
+
+        // Define custom logger
```
maxime-rainville, line 49: This bit of logic applies to the `dev/graphql/build` and `dev/graphql/clear` tasks.

src/Extensions/DevBuildExtension.php
```diff
@@ -31,6 +34,11 @@ public function onAfterBuild()
             return;
         }
         if (!self::$done) {
+            // Define custom logger
+            $logger = Logger::singleton();
+            $logger->setVerbosity(Logger::INFO);
+            Injector::inst()->registerService($logger, LoggerInterface::class . '.graphql-build');
+
```
maxime-rainville, line 41: This bit changes the logger when the schema is being built via `dev/build`.

Maybe, I should wrap this in a try-catch and restore it once I'm done. Although, not sure why any other task would want to use the graphql logger.

src/Schema/BulkLoader/BulkLoaderSet.php
```diff
@@ -80,7 +82,7 @@ public function addLoader(AbstractBulkLoader $loader): self
      */
     public function process(): Collection
     {
-        $logger = Logger::singleton();
+        $logger = Injector::inst()->get(LoggerInterface::class . '.graphql-build');
```
maxime-rainville, line 85: Actually not user how one would get to this bit of code.

src/Schema/Schema.php
```diff
@@ -148,7 +149,7 @@ public function __construct(string $schemaKey, SchemaConfig $schemaConfig = null
 
         $this->schemaConfig = $schemaConfig ?: SchemaConfig::create();
         $this->state = Configuration::create();
-        $this->logger = Logger::singleton();
+        $this->logger = Injector::inst()->get(LoggerInterface::class . '.graphql-build');
```
maxime-rainville, line 152: Actually not user how one would get to this bit of code.

src/Schema/Storage/CodeGenerationStore.php
```diff
@@ -123,7 +124,7 @@ public function __construct(string $name, CacheInterface $cache)
      */
     public function persistSchema(StorableSchema $schema): void
     {
-        $logger = Logger::singleton();
+        $logger = Injector::inst()->get(LoggerInterface::class . '.graphql-build');
```
maxime-rainville, line 127: Actually not user how one would get to this bit of code.

src/Extensions/DevBuildExtension.php
emteknetnz: Do you mean try-catch-finally?  I think that would make sense, if nothing else it communicates intent

src/Dev/Clear.php
```diff
@@ -0,0 +1,40 @@
...
+        '' => 'clear',
+    ];
+
+    private static $allowed_actions = [
+        'clear',
+    ];
+
+    public function clear(HTTPRequest $request): void
+    {
+
```
emteknetnz, line 24:
```suggestion
```

src/Dev/Clear.php
```diff
@@ -0,0 +1,40 @@
...
+        $logger->info('Clearing GraphQL code generation directory');
+        if ($fs->exists($expectedPath)) {
+            $logger->info('Directory has been found');
+            $fs->remove($expectedPath);
+            $logger->info('Directory has been removed');
+        } else {
+            $logger->info('Directory was not found. There is nothing to clear');
+        }
+
```
emteknetnz, line 38:
```suggestion
```

emteknetnz (commented):
Looks good

src/Dev/Build.php
GuySartorelli: Verbosity is currently [explicitly mentioned in the docs](https://docs.silverstripe.org/en/4/developer_guides/graphql/getting_started/building_the_schema/#controlling-verbosity) as something that can be controlled, so I think that section in the docs needs to be removed.

GuySartorelli:
Lots of deprecation warnings in the PHP8.0 travis run

_config/logging.yml
```diff
@@ -0,0 +1,8 @@
+---
+Name: graphql-logging
+after: '#logging'
+---
+
+SilverStripe\Core\Injector\Injector:
+  # This will be overridden by SilverStripe\GraphQL\Schema\Logger in SilverStripe\GraphQL\Dev\DevAdmin
```
GuySartorelli, line 7:
```suggestion
  # This will be overridden with SilverStripe\GraphQL\Schema\Logger in SilverStripe\GraphQL\Dev\DevelopmentAdmin
```

GuySartorelli: Probably worth mentioning here that it's a temporary override under certain conditions - otherwise it sounds like this config just doesn't do anything.

src/Dev/DevelopmentAdmin.php
```diff
@@ -42,6 +45,11 @@ protected function init()
...
+        $logger = Logger::singleton();
+        $logger->setVerbosity(Logger::INFO);
+        Injector::inst()->registerService($logger, LoggerInterface::class . '.graphql-build');
```
GuySartorelli, line 52: Why is this done here instead of in `SilverStripe\GraphQL\Dev\Build`? It feels like the intention is to wrap the actions taken by the `Build` controller in this logger, rather than all potential future graphql dev controllers?
Adding this to `Build` would also mean we don't need to override the logger both here and in `DevBuildExtension`.

GuySartorelli (commented):
I think I probably need a bit more context around what this is trying to solve? It's not super clear to me what the problem is and therefore I can't properly judge if this approach a) is appropriate and b) actually does what it needs to do.

src/Dev/DevelopmentAdmin.php
maxime-rainville: `SilverStripe\GraphQL\Dev\Build` is the generic logic that is executed whenever the schema is built. If you try to run a GraphQL query before the schema is built, the request will try to build the schema "on-demand".

If you swap the logger there, it will output the logs in that request and break your GraphQL output.

maxime-rainville:
Answered everyone's feedback.

GuySartorelli (approved):
Tested locally, looks good to me.
Logging during dev/build and dev/graphql/build works the same as it did before the PR.
Logging during a schema build on first request used to output the logs into the HTTP response which caused errors on the front-end - now logs using the `Psr\Log\LoggerInterface.errorhandler` logger defined in Framework instead.

maxime-rainville:
I think it's a bit too meta to test. To test that properly I would need something that nukes the GraphQL folder and rerun a request ... the logging is just as likely to break because it gets captured by PHPUnit instead of being captured in the response.

PR #450, opened 2022-04-04
ENH PHP 8.1 compatibility

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10250

This PR is for Graphql 3

### DIFF
--- a/src/Controller.php
+++ b/src/Controller.php
@@ -298,7 +298,7 @@ class Controller extends BaseController implements Flushable
             if ($allowedOrigin === '*') {
                 return true;
             }
-            if (strcasecmp($allowedOrigin, $origin) === 0) {
+            if (strcasecmp($allowedOrigin ?? '', $origin ?? '') === 0) {
                 return true;
             }
         }
@@ -347,7 +347,7 @@ class Controller extends BaseController implements Flushable
         $referer = $request->getHeader('Referer');
         if ($referer) {
             // Extract protocol, hostname, and port
-            $refererParts = parse_url($referer);
+            $refererParts = parse_url($referer ?? '');
             if (!$refererParts) {
                 return null;
             }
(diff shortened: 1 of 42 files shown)


src/Controller.php
```diff
@@ -515,14 +514,14 @@ public static function flush()
...
-            if (stristr($routeClass, Controller::class) !== false) {
+            if (stristr($routeClass ?? '', Controller::class) !== false) {
                 try {
                     $inst = Injector::inst()->convertServiceProperty($routeClass);
                     if ($inst instanceof Controller) {
                         /* @var Controller $inst */
                         $inst->processTypeCaching();
                     }
-                } catch (DatabaseException $e) {
+                } catch (Exception $e) {
```
GuySartorelli, line 524: Which exceptions are coming through that aren't `DatabaseException`? The comment in this catch says it's specifically there to catch failures where the table doesn't exist etc so it's specifically for database exceptions - I'm a bit worried about catching legitimate exceptions here.

GuySartorelli (commented):
The outstanding question on other PRs about `file_get_contents()` will also affect this PR.

src/Controller.php
emteknetnz: Oh this is wrong.  It was for the `mysql_sqi_exception` that was being triggered by php 8.1.  I must have not reverted, it should be DatabaseException.

GuySartorelli (approved):
For posterity: See discussion in silverstripe/silverstripe-admin#1294 about the general approach. tl;dr:
> the purposes of this PR is to avoid passing null (specifically null) from built-in functions that as of php 8.1 will throw deprecation warnings if null is passed in. The approach is necessarily heavy-handed, and while there are many situations where it isn't needed, it would be prohibitively laborious to find all of those situations by hand.

Looks good to me

PR #454, opened 2022-04-12
DEP Upgrade webonyx/graphql-php to v14

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10250

### DIFF
--- a/tests/Fake/FakeResolveInfo.php
+++ b/tests/Fake/FakeResolveInfo.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace SilverStripe\GraphQL\Tests\Fake;
+
+use GraphQL\Type\Definition\FieldDefinition;
+use GraphQL\Type\Definition\ResolveInfo;
+use GraphQL\Type\Definition\ObjectType;
+use GraphQL\Type\Definition\Type;
+use GraphQL\Type\Schema;
+use SilverStripe\Dev\TestOnly;
+
+class FakeResolveInfo extends ResolveInfo implements TestOnly
+{
+    public function __construct(array $options = [])
+    {
+        // This is a minimal implementation that's just good enough
+        // to get unit tests to pass
+        $name = 'fake';
+        $type = Type::string();
+        foreach ($options as $key => $value) {
+            if ($key === 'name') {
+                $name = $value;
+            }
+            if ($key === 'type') {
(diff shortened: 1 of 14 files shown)


tests/Fake/FakeResolveInfo.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace SilverStripe\GraphQL\Tests\Fake;
+
+use GraphQL\Type\Definition\FieldDefinition;
+use GraphQL\Type\Definition\ResolveInfo;
+use GraphQL\Type\Definition\ObjectType;
+use GraphQL\Type\Definition\Type;
+use GraphQL\Type\Schema;
+
+class FakeResolveInfo extends ResolveInfo
```
GuySartorelli, line 11:
```suggestion
class FakeResolveInfo extends ResolveInfo implements TestOnly
```
Also add `use` statement.

GuySartorelli (changes requested):
Remove the `use GraphQL\Type\Definition\ResolveInfo;` statements from files where you've replaced that with the new `FakeResolveInfo` class.

Why do we need a `FakeResolveInfo` class? I just want to be sure that it's okay to do this for tests without getting false positive results.

In general looks good (haven't tested it locally) - couple of small changes and a bit of context to be given there.

tests/Fake/FakeResolveInfo.php
emteknetnz: > Why do we need a FakeResolveInfo class? I just want to be sure that it's okay to do this for tests without getting false positive results.

Between v0.12 and v14 of webonyx/graphql-php the constructor for ResolveInfo changed from an array, to 10 params that are mostly non-nullable.  For the majority of the tests we don't even need the ResolveInfo object for anything, it's just a non-optional argument.  So the FakeResolveInfo is just a nice way to fill in the 10 constructor arguments really.

Asset-admin, versioned and elemental already have existing FakeResolveInfo classes so I just copied what was done there.

GuySartorelli (changes requested):
Missed a `use GraphQL\Type\Definition\ResolveInfo;` in `tests/Scaffolding/Scaffolders/CRUD/ReadTest.php`

GuySartorelli (approved):
I'm happy with this.
MOG.

PR #457, opened 2022-04-21
FIX Change Logger default verbosity to info

maxime-rainville:
You can't really override verbosity right now because it's explicitly set in DevelopmentAdmin

# Parent issue 
- Fixes https://github.com/silverstripeltd/product-issues/issues/519


GuySartorelli:
I think I need a reminder about this - why can't verbosity be set? There is definitely a `setVerbosity` method and the `level` property is used in several places.

emteknetnz (approved):

Issue #465, opened 2022-05-19
SPIKE Investigate making default location of .graphql-generated folder inside /tmp


emteknetnz:
Timebox: 2 days

Defaulting the .graphql-generated folder in the project root could be problem because this folder needs to be writable.  Existing webservers may suddently break on upgrade if they are configured in a way that does not allow this folder to be created or be writable.

A safer/better alternative for the majority of use cases may be to default the .graphql-generated folder somewhere in /tmp (`TEMP_FOLDER`), possibly within the regular `/tmp/silverstripe-cache-*` where other files such as template files converted to php files are stored

There may be some reasons for wanting to have the .graphql-generated folder in the project root, for instance if the schema is generated in CI/CD as part of a packaged deployment.  So we would need the ability to configure if the folder is stored in /tmp or in the project root.  There is already configuration available through `CodeGenerationStore.$dirName` to configure the name of the folder.

Likely that there would only need to be minimal code changes in `CodeGenerationStore.php` made

### ACs
- [x] Investigate if having the `.graphql-generated` located the `/tmp` folder is viable.
- [ ] Schemas for multiple independent websites on the same server must not collide
- [ ] Test on on prod-like webservers, including multi-server environments where the temp folder may not be shared by the instances behind a load balancer
- [x] If this seems doable, create a PR to `4.0` (which is still pre-stable)

### Questions
- Would we want to define a folder in the web root for new projects (i.e. via silverstripe/installer)?
    - This would make the schema more reliably available for these projects since it won't be cleared out whenever the temp dir is cleared
    - This would mean new projects need different permissions by default than existing projects at launch

### PRs
- https://github.com/silverstripe/silverstripe-graphql/pull/467

michalkleiner:
> There may be some reasons for wanting to have the .graphql-generated folder in the project root, for instance if the schema is generated in CI/CD as part of a packaged deployment. So we would need the ability to configure if the folder is stored in /tmp of in the project root. There is already configuration available through CodeGenerationStore.$dirName to configure the name of the folder.

That was our use case, generating on CI server and packaging it all up with composer dependencies installed and FE files compiled in a deployable package, so that the webserver can serve it right away.

I would also add to the ACs to test it in multi-server environments where the temp folder may not be shared by the instances behind a load balancer whereas assets folder would be.

emteknetnz:
Have added to ACs

GuySartorelli:
[...]

PR #467, opened 2022-05-23
ENH Store generated graphql schema in a temp directory by default [POC]

GuySartorelli:
This is a proof of concept for storing the generated graphql schema in the `silverstripe-cache` directory.

This makes the process for migrating from graphql 3 to 4 more streamline by respecting the filesystem permissions that we have previously indicated (i.e. silverstripe only needs write access to the assets dir - _not_ the project root). For any project without its own custom schema, there should be no need to suddenly start worrying about "what is this new folder I need to add permissions for?"

Because this uses the `silverstripe-cache` directory, the schema will also be built for each system user that runs a command by default - so if the user running `sake dev/graphql/build` is different from the user running the webserver, the webserver will still end up creating a _new schema_ on demand, ignoring the one generated via sake. We could avoid this pitfall by finding a temporary directory in another way instead of relying on `TEMP_FOLDER` (e.g. checking `SS_TEMP_PATH` and falling back to `sys_get_temp_dir()`). We would need to make sure this is fine with the platform team, as they may want to ensure that directory gets blown away on deployments (like the cache directory is).

The directory is still configurable, and the wording I've chosen indicates that it's recommended to defined it. Defining the directory will mean the schema is more readily available, since it won't be blown away any time the temporary directory is cleared. We will probably want to define this by default in silverstripe/installer for new projects.

Prior to this PR, setting the directory config to a blank string would result in errors.

## Parent Issue
- https://github.com/silverstripe/silverstripe-graphql/issues/465

### DIFF
--- a/src/Dev/Build.php
+++ b/src/Dev/Build.php
@@ -67,11 +67,16 @@ class Build extends Controller
 
         // Check for old code dir
         if (is_dir(BASE_PATH . '/.graphql')) {
+            $dirName = CodeGenerationStore::config()->get('dirName');
+            $newDirNameMsg = $dirName
+                ? 'The new directory is named "' . $dirName . '".'
+                : 'The schema will be built inside your silverstripe-cache directory by default,
+                but we recommend setting a specific directory in '
+                . CodeGenerationStore::class . '.dirName';
             $logger->warning(
                 'You have a .graphql/ directory in your project root. This is no longer the default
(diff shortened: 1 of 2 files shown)


GuySartorelli:
The decision was made to leave this as-is in order to retain our release schedule. We will need to ensure our documentation clearly indicates that a change may be required for _all_ projects which don't currently have a writable project root.

PR #470, opened 2022-05-31
ENH Add .htaccess file to schema directories

### LINKED ISSUE #452: Add a `.htaccess` file to `.graphql-generated` to dissallow any direct access to schema file
AS a general pre-caution, it probably would be advisable to automatically create a `.htaccess` file that in the `.graphql-generated` folder that disallow any direct access to the files in there.

## ACs
- [x] If the graphql schema is inside the public directory, it cannot be accessed via a web request.

## PRs
- https://github.com/silverstripe/silverstripe-graphql/pull/470

GuySartorelli:
Silverstripe Cloud will put the schema inside the public directory if it is declared as a shared folder in platform.yml
Declaring it as a shared folder is going to be our recommended best practice.

## Parent issue
- https://github.com/silverstripe/silverstripe-graphql/issues/452


emteknetnz (approved):
Have tested locally with a CodeGenerationStore.php `private static string $dirName = 'public/graphql-generated';` so that it's in the public folder and confirmed that .htaccess prevents directly access to files

PR #475, opened 2022-06-17
typecast vars to ensure the value is an array

jordyrosenbrand:
The method `callMiddleware` expects the fourth parameter to be an array while in the method `queryAndReturnResult` it can be either an array or null.

### DIFF
--- a/src/QueryHandler/QueryHandler.php
+++ b/src/QueryHandler/QueryHandler.php
@@ -102,7 +102,7 @@ class QueryHandler implements
             return GraphQL::executeQuery($schema, $query, null, $context, $vars);
         };
 
-        return $this->callMiddleware($schema, $query, $context, (array) $vars, $last);
+        return $this->callMiddleware($schema, $query, $context, $vars ?? [], $last);
     }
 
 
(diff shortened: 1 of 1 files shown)


src/QueryHandler/QueryHandler.php
```diff
@@ -102,7 +102,7 @@ public function queryAndReturnResult(GraphQLSchema $schema, $query, ?array $vars
             return GraphQL::executeQuery($schema, $query, null, $context, $vars);
         };
 
-        return $this->callMiddleware($schema, $query, $context, $vars, $last);
+        return $this->callMiddleware($schema, $query, $context, (array) $vars, $last);
```
ChloeSartorelli:
```suggestion
        return $this->callMiddleware($schema, $query, $context, $vars ?? [], $last);
```
This is the syntax we've used elsewhere in core code for this purpose - we should continue following this pattern.

GuySartorelli (changes requested):
Thanks! Good catch. Just a small change to make.

Can you please also retarget on the `4.0` branch?

GuySartorelli:
The travis failure is also reflected on the `4` branch itself so it not related to this PR: https://app.travis-ci.com/github/silverstripe/silverstripe-graphql/jobs/573407968#L1976

GuySartorelli (approved):

PR #477, opened 2022-06-21
[v4] Relax possible filter value types

### LINKED ISSUE #476: [v4] InFilter not working as expected
### Problem
The InFilter does not work as expected, or it is not clear how it should be used.

### Steps to reproduce
**_graphql/models.yml**
```yaml
Page:
  fields:
    id: true
    title: true
  operations:
    read: true
```
Query
```graphql
query getPages {
    readPages(filter: {
        title: {in: ["Foo", "Bar"]}
    }) {
        nodes {
            id
            title
        }
    }
}
```
As `InFilter::apply()` expects a string as value instead of an array I also tried this. But I'm not sure how this is expected to work:
```graphql
query getPages {
    readPages(filter: {
        title: {in: "Foo, Bar"}
    }) {
        nodes {
            id
            title
        }
    }
}
```
### Expected behaviour
I expect to see a result with pages that have "Foo" or "Bar" as title.
### Actual behaviour
This error is thrown:
`Argument 3 passed to SilverStripe\\GraphQL\\Schema\\DataObject\\Plugin\\QueryFilter\\Filters\\InFilter::apply() must be of the type string, array given`

bummzack:
Fixes #476

### DIFF
--- a/tests/Schema/DataObject/Plugin/FiltersTest.php
+++ b/tests/Schema/DataObject/Plugin/FiltersTest.php
@@ -1,0 +1,112 @@
+<?php
+
+namespace SilverStripe\GraphQL\Tests\Schema\DataObject\Plugin;
+
+use SilverStripe\Dev\SapphireTest;
+use SilverStripe\GraphQL\Schema\DataObject\Plugin\QueryFilter\FieldFilterInterface;
+use SilverStripe\GraphQL\Schema\DataObject\Plugin\QueryFilter\Filters\ContainsFilter;
+use SilverStripe\GraphQL\Schema\DataObject\Plugin\QueryFilter\Filters\EndsWithFilter;
+use SilverStripe\GraphQL\Schema\DataObject\Plugin\QueryFilter\Filters\EqualToFilter;
+use SilverStripe\GraphQL\Schema\DataObject\Plugin\QueryFilter\Filters\GreaterThanFilter;
+use SilverStripe\GraphQL\Schema\DataObject\Plugin\QueryFilter\Filters\GreaterThanOrEqualFilter;
+use SilverStripe\GraphQL\Schema\DataObject\Plugin\QueryFilter\Filters\InFilter;
+use SilverStripe\GraphQL\Schema\DataObject\Plugin\QueryFilter\Filters\LessThanFilter;
+use SilverStripe\GraphQL\Schema\DataObject\Plugin\QueryFilter\Filters\LessThanOrEqualFilter;
+use SilverStripe\GraphQL\Schema\DataObject\Plugin\QueryFilter\Filters\NotEqualFilter;
+use SilverStripe\GraphQL\Schema\DataObject\Plugin\QueryFilter\Filters\StartsWithFilter;
+use SilverStripe\ORM\ArrayList;
+
+class FiltersTest extends SapphireTest
+{
+    private array $values = [
+        'string' => 'test',
+        'array' => ['a', 'b'],
+        'number' => 42,
+        'null' => null
+    ];
+
+    public function testFilterArguments()
+    {
+        // Contains
+        $this->testFilter(
+            new ContainsFilter(),
+            'contains',
+            array_slice($this->values, 0, 2, true)
+        );
+
+        // Endswith
+        $this->testFilter(
+            new EndsWithFilter(),
+            'endswith',
+            array_slice($this->values, 0, 2, true)
+        );
+
+        // EqualToFilter
+        $this->testFilter(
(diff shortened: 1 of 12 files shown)


tests/Schema/DataObject/Plugin/FiltersTest.php
```diff
@@ -0,0 +1,112 @@
...
+class FiltersTest extends SapphireTest
+{
+    private array $values = [
+        'string' => 'test',
+        'array' => ['a', 'b'],
+        'number' => 42,
+        'null' => null
+    ];
+
+    public function testFilterArguments()
```
ChloeSartorelli: instead of calling `testFilter()`, this should return an array and should be named `filterProvider()`. Then you can put `@dataProvider filterProvider` in the PHDoc for `testFilter()` and phpunit will handle passing in the arguments in a way that tests each case even if one fails, and reports on failures in a useful way.

The array would look something like this:
```php
return [
    [
        new ContainsFilter(),
        'contains',
        array_slice($this->values, 0, 2, true)
    ],
    [
        new EndsWithFilter(),
        'endswith',
        array_slice($this->values, 0, 2, true)
    ],
    //....
];
```

bummzack: @GuySartorelli Oh, that's something I didn't know. Neat! I've updated the test. Thanks for the heads up.

tests/Schema/DataObject/Plugin/FiltersTest.php
```diff
@@ -109,4 +43,70 @@ private function testFilter(FieldFilterInterface $filter, string $identifier, ar
             );
         }
     }
+
+    public function filterProvider(): array
+    {
+        return [
+            // Contains
```
ChloeSartorelli: Please remove these comments - it's clear from the class name of the filter which filter is being tested.

tests/Schema/DataObject/Plugin/FiltersTest.php
```diff
@@ -109,4 +43,70 @@ private function testFilter(FieldFilterInterface $filter, string $identifier, ar
             );
         }
     }
+
+    public function filterProvider(): array
```
ChloeSartorelli:
```suggestion
    public function filterArgumentsProvider(): array
```
The convention tends to be that it matches the name of the test method that's using it like this.

tests/Schema/DataObject/Plugin/FiltersTest.php
```diff
@@ -25,80 +25,14 @@ class FiltersTest extends SapphireTest
...
-            array_slice($this->values, 0, 2, true)
-        );
-    }
-
-    private function testFilter(FieldFilterInterface $filter, string $identifier, array $params)
+    /**
+     * @param FieldFilterInterface $filter
+     * @param string $identifier
+     * @param array $params
+     * @dataProvider filterProvider
```
ChloeSartorelli:
```suggestion
     * @dataProvider filterArgumentsProvider
```

tests/Schema/DataObject/Plugin/FiltersTest.php
```diff
@@ -25,80 +25,14 @@ class FiltersTest extends SapphireTest
...
+     * @return void
```
ChloeSartorelli:
```suggestion
     * @dataProvider filterProvider
```
No need to have PHPDocs that duplicate the typings - these have a tendency to get out of date and since we have typehints they don't provide any extra information.

tests/Schema/DataObject/Plugin/FiltersTest.php
```diff
@@ -25,80 +25,14 @@ class FiltersTest extends SapphireTest
...
+     */
+    public function testFilterArguments(FieldFilterInterface $filter, string $identifier, array $params)
```
ChloeSartorelli:
```suggestion
    public function testFilterArguments(FieldFilterInterface $filter, string $identifier, array $params): void
```

GuySartorelli (changes requested):
Just a few more small changes to make and then we're good to go.

tests/Schema/DataObject/Plugin/FiltersTest.php
bummzack: I have addressed your code-quality improvements in the latest commit.

GuySartorelli (approved):
The only Travis failure is identical to the one happening on the 4.0 branch already https://app.travis-ci.com/github/silverstripe/silverstripe-graphql/builds/252256400 which is not related to this PR.

Thank you for fixing this, it's very much appreciated.

PR #481, opened 2022-07-05
MNT Use GitHub Actions CI

emteknetnz:
Issue https://github.com/silverstripe/gha-ci/issues/11<br /><br />Workflow run https://github.com/creative-commoners/silverstripe-graphql/actions?query=branch%3Apulls%2F3.8%2Fmodule-standards

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,22 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+  # Every Wednesday at 11:00am UTC
+  schedule:
+    - cron: '0 11 * * 3'
+
+jobs:
+  ci:
+    name: CI
+    # Only run cron on the silverstripe account
+    if: (github.event_name == 'schedule' && startsWith(github.repository, 'silverstripe/')) || (github.event_name != 'schedule')
+    uses: silverstripe/gha-ci/.github/workflows/ci.yml@v1
+    with:
+      extra_jobs: |
+        - php: 7.4
+          endtoend: true
+          endtoend_suite: asset-admin
+          endtoend_config: vendor/silverstripe/asset-admin/behat.yml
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,30 @@
...
+          endtoend_suite: asset-admin
+          endtoend_config: vendor/silverstripe/asset-admin/behat.yml
+        - php: 7.4
+          endtoend: true
+          endtoend_suite: versioned-admin
+          endtoend_config: vendor/silverstripe/versioned-admin/behat.yml
+        - php: 7.4
+          endtoend: true
+          endtoend_suite: silverstripe-elemental
+          endtoend_config: vendor/dnadesign/silverstripe-elemental/behat.yml
```
ChloeSartorelli: Why versioned and elemental? Those aren't mentioned in the travis config on this branch.
Also if we _should_ be including versioned-admin, should we do that for the 4.0 branch as well?

composer.json
```diff
@@ -18,7 +18,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^9.5",
-        "squizlabs/php_codesniffer": "^3.0"
+        "squizlabs/php_codesniffer": "^3.0",
+        "silverstripe/frameworktest": "^0.4.3"
```
ChloeSartorelli, line 23: Do we need to add versioned-admin and elemental as well?

GuySartorelli (changes requested):

.github/workflows/ci.yml
emteknetnz: I'm just copying what was previously in the travis file

emteknetnz: oh, copied it from the wrong branch :-) 3.8 only has asset-admin. 4.0 has asset-admin, versioned-admin and elemental

composer.json
emteknetnz: no, only supposed to have asset-admin in this one, have added to require-dev

GuySartorelli (approved):

PR #498, opened 2022-09-12
FIX Use symfony6 cache classes

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/590


GuySartorelli (commented):
This is only in a test - is there any corresponding change required in `src/`?

emteknetnz:
Nope, graphql is just consuming the cache that's defined in symfony/cache that's since been updated to v6

GuySartorelli (approved):

GuySartorelli:
Doesn't resolve the existing CI issues but it is a good change.

PR #502, opened 2022-10-26
DEP Upgrade installer dependencies

sabina-talipova:
### Parent issue
- https://github.com/silverstripe/silverstripe-installer/issues/335

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,10 +7,10 @@
         "php": "^8.1",
         "silverstripe/framework": "^5",
         "silverstripe/vendor-plugin": "^2",
-        "webonyx/graphql-php": "^14.11.6",
+        "webonyx/graphql-php": "^14.11.8",
         "silverstripe/event-dispatcher": "^1",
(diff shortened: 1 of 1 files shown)


sabina-talipova:
Failed on unrelated Behat test, related to TinyMCE.
Probably we should update the following [behat tests](https://github.com/search?q=org%3Asilverstripe+mce-menu&type=code)

GuySartorelli:
The behat test is failing because the `5.0.0-alpha1` version of elemental is being installed, when it should be the `5.x-dev` version.
I've created https://github.com/silverstripe/silverstripe-graphql/pull/504 to resolve this.
Note that the `.mce-menu` stuff you're seeing in the search is showing up because `4` is still the default branch for those modules, which is using the old tinymce 4 and therefore the old classes.

GuySartorelli (approved):
I don't think these updates were necessary but they don't hurt anything - happy to merge.
Failed behat test is unrelated to this PR.

Issue #511, opened 2022-12-14
CMS 5 - Upgrade to webonyx/graphql-php v15


emteknetnz:
We currently use v14 on webonyx/graphql-php - but it is not PHP 8.2 compatible.  This was uncovered as part of the [PHP 8.2 support card](https://github.com/silverstripe/silverstripe-framework/issues/10403)

v14 will emit multiple deprecation warnings when you dev/build flush=1 using PHP 8.2
```
vendor/webonyx/graphql-php/src/Type/Definition/FieldDefinition.php:263
vendor/webonyx/graphql-php/src/Type/Definition/InputObjectField.php:106
vendor/webonyx/graphql-php/src/Utils/BreakChangesFinder.php:106 (and many others in this file)
vendor/webonyx/graphql-php/src/Utils/BuildClientSchema.php:190
vendor/webonyx/graphql-php/src/Validators/Rules/ValuesOfCorrectType.php:182
```

I submitted a [PR add PHP 8.2 support for v14](https://github.com/webonyx/graphql-php/pull/1256) but it was rejected due to v14 now being an unsupported versioned

We're forced to upgrade to v15 for PHP 8.2 support

I tried upgrading to v15 - `"webonyx/graphql-php": "dev-master as 14.99.99"` but it immediately failed unit tests on my local

v15 (dev-master) is currently in beta.  On the linked PR above I asked for a timeline for when v15 goes to stable.  Ideally we should be using a stable version, though we may be forced to use a beta.

If v15 is not PHP 8.2 compatible, then submit a PR to add PHP 8.2 support

### Acceptance Criteria
- [ ] silverstripe/graphql uses v15 of webonyx/graphql-php
- [ ] webonyx/graphql-php does not generate deprecation warnings in PHP 8.2 when you run dev/build flush=1

### PRs
- https://github.com/silverstripe/silverstripe-graphql/pull/512

maxime-rainville:
PR is merged.

PR #512, opened 2023-01-13
DEP Use webonyx/graphq-php 15

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-graphql/issues/511

I've added an extra job to PHP 8.2 in CI

Behat failure for elemental job I believe is because elemental is not yet CMS 5 compatible?  I get the same error locally running both PHP 8.1 and PHP 8.2 with the the old v14 of webonyx/graphql.

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,3 +28,6 @@ jobs:
           endtoend: true
           endtoend_suite: silverstripe-elemental
           endtoend_config: vendor/dnadesign/silverstripe-elemental/behat.yml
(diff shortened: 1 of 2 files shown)


GuySartorelli:
Elemental was made cms5 compatible in the first run - BUT it's JavaScript hasn't been updated yet which is causing the behat failures. 
That's being done in https://github.com/silverstripe/silverstripe-elemental/pull/1032

maxime-rainville (approved):
Such a massive refactor!

PR #515, opened 2023-02-09
Support maximumLimit in pagination

ec8or:
At the moment, you can only change the maximum limit in the overall settings. This enables support for the maximumLimit setting per model.

## Issues
- https://github.com/silverstripe/silverstripe-graphql/issues/543

### DIFF
--- a/src/Schema/Plugin/PaginationPlugin.php
+++ b/src/Schema/Plugin/PaginationPlugin.php
@@ -88,7 +88,7 @@ class PaginationPlugin implements FieldPlugin, SchemaUpdater
 
         $defaultLimit = $config['defaultLimit'] ?? $this->config()->get('default_limit');
         $connectionName = $config['connection'] ?? $plainType;
-        $max = $this->config()->get('max_limit');
+        $max = $config['maximumLimit'] ?? $this->config()->get('max_limit');
         $limit = min($defaultLimit, $max);
         $field->addArg('limit', "Int = $limit")
             ->addArg('offset', "Int = 0")


GuySartorelli (changes requested):
Can you please add a test for this scenario?

ec8or:
I could try and have a look, but doesn't seem like there's any tests for setting it via the main config at the moment (or for defaultLimit)?

flamerohr:
@GuySartorelli looking into this for @ec8or, how would I test `PaginationPlugin::apply`?

I had found `testBasicPaginator` but that seems to use its own resolver with `IntegrationTestResolver::testPaginate`

GuySartorelli:
Probably the simplest way, without diving in too deep, would be to set up a new test schema folder, set up whatever schema you need in order to test a) without setting this and b) with setting this, and then write a test that executes a query on each of those scenarios.
I can't go into more detail than that without diving into the code myself though, and unfortunately I don't have time for that right now. Does this give you enough to work from?

GuySartorelli:
Ultimately though, if you can't get a test working with the time you have to look into this, then if you can give me a simple example of how to test it locally that would be okay for now.

flamerohr:
Thanks! I'm not sure if I'll have the time to build something new like that.

I can include test schema data that I had in mind for this.

The query I'd use for both scenarios:
```gql
query {
  readMyTypes(limit: 5) {
    nodes {
        field1
    }
    pageInfo {
      totalCount
    }
  }
}
```

For with the config schema:
```yml
SilverStripe\GraphQL\Tests\Fake\DataObjectFake:
  fields:
    myField: true
  operations:
    read:
      plugins:
        paginateList:
          maximumLimit: 3
```
and the query from above, which checks the size of `nodes` to be 3, and `totalCount` to be something greater than 5.

For without the config schema:
```yml
SilverStripe\GraphQL\Tests\Fake\DataObjectFake:
  fields:
    myField: true
  operations:
    read:
      plugins:
        paginateList: true
```
and the query from above, which checks the size of `nodes` to be 5, and `totalCount` to be something greater than 5.

GuySartorelli:
I've changed the target branch to `5` since we won't be releasing any new minor versions for the `4.x` release line, and I've rebased the PR so it can merge into the `5` branch without issues. The actual code change in this PR hasn't changed.

GuySartorelli (approved):
Thanks for this. Works exactly as advertised. Will merge when CI goes green.

Issue #500, opened 2022-10-19
GraphQL returns an error, causing elemental to fail to initialise in live mode (sometimes)


edwilde:
When the user opens a page within the CMS the elemental area gets stuck indefinitely on the loading screen, causing the spinning Silverstripe logo of death.

![Screen Shot 2022-08-18 at 1 48 50 PM](https://user-images.githubusercontent.com/415374/196819520-28ca819c-5580-44b9-8f13-52d2c5d80200.jpg)

Response when requesting `/admin/graphql/` is `Unknown graphql type: Query` instead of the expected data.

I have seen this in a few different environments when Silverstripe 4.11.2 is deployed with Silverstripe Cloud and php 8.0. 

I have not seen it locally and I do not know if it can be reproduced with a defined set of steps. It is perhaps linked to many changes being made to page content, as it has typically manifested just before site launches. But that is just a hunch.

This can be resolved by running `/dev/graphql/build` to rebuild the schema, however this will not fix it permanently only for an indeterminate amount of time; I think the minimum time was about 1.5 days.

Apparently this can also be resolved by generating and committing the graphql cache files to the repo and deploying those, however I have not confirmed that.

### Acceptance criteria
- [ ] We have identified what triggers this behaviour
- [x] This behaviour never occurs.

### PRs
- https://github.com/silverstripe/silverstripe-graphql/pull/506
- https://github.com/silverstripe/silverstripe-graphql/pull/507
- https://github.com/silverstripe/silverstripe-graphql/pull/521
- https://github.com/silverstripe/silverstripe-graphql/pull/530
- https://github.com/silverstripe/developer-docs/pull/193

michalkleiner:
Might be the difference between dev-built (dev mode) GraphQL schema and non-dev-built (test/live mode) schema where each uses different obfuscator when generating the classes. Once your schema is built in dev mode it doesn't work in test/live mode and the other way round. Also when deployed if the webserver/PHP doesn't have write access to the `.graphql-generated` folder to be created next to the code it would fail.

Best option is to generate the graphql schema files using the `vendor/bin/sake /dev/graphql/build clear=1 flush=1 schema=default` and `vendor/bin/sake /dev/graphql/build clear=1 flush=1 schema=admin` commands and commit the product of that to the repo for a release. That works for us every time. If you delete the `.graphql` and `.graphql-generated` folders and run `vendor/bin/sake dev/build flush=1` command that should also regenerate the schema files.

[...]

PR #521, opened 2023-03-07
FIX Attempt to rebuild the schema if a schema file is missing

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-graphql/issues/500

Schema building code is copied from https://github.com/silverstripe/silverstripe-graphql/blob/4/src/Dev/Build.php#L81 though I've removed all the logging.

I did have some concerns that this may lead to make it easy to allow bad users to trigger schema rebuilds by sending graphql requests the deliberately contain things that aren't in the schema, though there appears to be validation done before it reaches that point.  For instances this is the standard graphql request send when doing a request for elemental, :

`{"operationName":"ReadBlocksForArea","variables":{"id":1},"query":"query ReadBlocksForArea($id: ID!) {\n  readOneElementalArea(filter: {id: {eq: $id}}, versioning: {mode: DRAFT}) {\n    elements {\n      id\n      title\n      blockSchema\n      obsoleteClassName\n      isLiveVersion\n      isPublished\n      version\n      canCreate\n      canPublish\n      canUnpublish\n      canDelete\n      __typename\n    }\n    __typename\n  }\n}\n"}`

I've tried adding "X"'s to various bits of this request, and tested using Firefox's "Edit and resend" in the network panel, though none of them trigger a debug break point I placed in the new code that rebuilds the schema `AbstractTypeRegistry::get()`.

These are the responses I got from the server.
- `"operationName":"XReadBlocksForAreaX"` -- standard response i.e. "ReadBlocksForArea" key isn't important
- `"XqueryX":"query ReadBlocksForArea(` -- `{ message: 'This endpoint requires a "query" parameter', code: 400, file: "/var/www/vendor/silverstripe/framework/src/Control/RequestHandler.php", … }`
- `"query":"XqueryX ReadBlocksForArea(` -- `{ message: 'Syntax Error: Unexpected Name "XqueryX"', code: 0, file: "/var/www/vendor/webonyx/graphql-php/src/Language/Parser.php", … }`
- `"query":"query XReadBlocksForAreaX(` -- standard response i.e. "ReadBlocksForArea" after query isn't important
- `XreadOneElementalAreaX(filter` -- `{ message: 'Cannot query field "XreadOneElementalAreaX" on type "Query". Did you mean "readOneElementalArea"?', code: 0, file: "/var/www/vendor/webonyx/graphql-php/src/Validator/Rules/FieldsOnCorrectType.php", … }`
- `readOneXElementalAreaX(` -- `{ message: 'Cannot query field "readOneXElementalAreaX" on type "Query". Did you mean "readOneElementalArea"?', code: 0, file: "/var/www/vendor/webonyx/graphql-php/src/Validator/Rules/FieldsOnCorrectType.php", … }`

### DIFF
--- a/src/Schema/Storage/AbstractTypeRegistry.php
+++ b/src/Schema/Storage/AbstractTypeRegistry.php
@@ -9,6 +9,10 @@ use GraphQL\Type\Definition\NonNull;
 use GraphQL\Type\Definition\ListOfType;
 use Exception;
 use SilverStripe\Core\Injector\Injector;
+use SilverStripe\GraphQL\Schema\SchemaBuilder;
+use SilverStripe\GraphQL\Schema\Exception\EmptySchemaException;
+use SilverStripe\Control\Controller;
+use SilverStripe\GraphQL\Controller as GraphQLController;
 
 abstract class AbstractTypeRegistry
 {
@@ -21,7 +25,30 @@ abstract class AbstractTypeRegistry
      */
     public static function get(string $typename)
     {
-        return static::fromCache($typename);
+        try {
+            return static::fromCache($typename);
+        } catch (Exception $e) {
+            if (!Controller::has_curr() ||
+                !(Controller::curr() instanceof GraphQLController) ||
+                !Controller::curr()->autobuildEnabled()
+            ) {
+                throw $e;
+            }
+            // Try to rebuild the whole schema as fallback.
+            // This is to solve mysterious edge cases where schema files do not exist when they should.
+            // These edge cases are more likely on multi-server environments
+            $dirParts = explode(DIRECTORY_SEPARATOR, static::getSourceDirectory());
+            $key = $dirParts[count($dirParts) - 1];
+            $builder = SchemaBuilder::singleton();
+            $schema = $builder->boot($key);
+            try {
+                $builder->build($schema, true);
+            } catch (EmptySchemaException $e) {
+                // noop
+            }
+            // Attempt to return again now the schema has been rebuilt.
+            return static::fromCache($typename);
+        }
     }
 
     abstract protected static function getSourceDirectory(): string;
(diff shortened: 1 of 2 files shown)


src/Schema/Storage/AbstractTypeRegistry.php
```diff
@@ -21,7 +23,24 @@ abstract class AbstractTypeRegistry
      */
     public static function get(string $typename)
     {
-        return static::fromCache($typename);
+        try {
+            return static::fromCache($typename);
+        } catch (Exception $e) {
+            // Try to rebuild the whole schema as fallback.
```
ChloeSartorelli, line 37: We need to make sure we respect `SilverStripe\GraphQL\Controller::autobuildEnabled()`
```suggestion
            if (!Controller::has_curr() || !(Controller::curr() instanceof GraphQLController) || !Controller::curr()->autobuildEnabled()) {
                throw $e;
            }
            // Try to rebuild the whole schema as fallback.
```

The class names here assume the new use statements:
```php
use SilverStripe\Control\Controller;
use SilverStripe\GraphQL\Controller as GraphQLController;
```

Note that the `Controller::has_curr()` here is probably redundant - but please include it. I'd rather be safe than sorry here.

GuySartorelli (changes requested):
Looks fine, works locally. I can't find any way to trip this up or cause it to do anything unexpected. We only ever get to this part of the code if _somewhere in the schema_ it says there's a type, so it only happens in these scenarios:
1. The schema is incorrectly configured (should be caught during dev so no impact on production)
2. The schema is incomplete (e.g. files removed or schema build interrupted somehow - this seems to be the scenario of the original report)

Just one change to make to be sure we're respecting peoples' configuration.

src/Schema/Storage/AbstractTypeRegistry.php
emteknetnz: Updated

GuySartorelli (approved):

GuySartorelli:
Released as [4.1.1](https://github.com/silverstripe/silverstripe-graphql/releases/tag/4.1.1) and [4.2.2](https://github.com/silverstripe/silverstripe-graphql/releases/tag/4.2.2)

Issue #522, opened 2023-03-09
GraphQL errors silently if backtrace includes arguments


GuySartorelli:
When PHP is configured to include arguments in backtrace (`zend.exception_ignore_args` to "Off" in your php.ini) this causes GraphQL to error silently.

## Acceptance criteria
- [x] GraphQL errors are always returned correctly
- [x] GraphQL error backtrace treats arguments the same way as the Backtrace class in framework
- [x] Errors have the same JSON error structure before and after the fix

## PRs
- https://github.com/silverstripe/silverstripe-graphql/pull/523

PR #523, opened 2023-03-09
FIX Don't error silently when arguments are enabled

GuySartorelli:
If PHP is configured to include arguments, the `json_encode($result)` call in graphql's error handling was hitting a recursion error (specifically found while validating https://github.com/silverstripe/silverstripe-graphql/issues/500).

This PR avoids that problem entirely by treating the backtrace the same as we do everywhere else in framework. It also ensures that arguments are filtered appropriately to avoid leaking sensitive information.

## Issue
- https://github.com/silverstripe/silverstripe-graphql/issues/522

### DIFF
--- a/src/Controller.php
+++ b/src/Controller.php
@@ -30,6 +30,7 @@ use SilverStripe\Security\Member;
 use SilverStripe\Security\Permission;
 use SilverStripe\Versioned\Versioned;
 use BadMethodCallException;
+use SilverStripe\Dev\Backtrace;
 
 /**
  * Top level controller for handling graphql requests.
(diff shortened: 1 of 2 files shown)


emteknetnz (changes requested):
Can we add a unit-test that will cause an exception that has a >10000 character argument and assert than it's truncating the argument in the backtrace?

Note if this is unreasonably hard to do in CI e.g. `If PHP is configured to include arguments` then I'm OK with not including the unit test here

GuySartorelli:
Rather than doing this with a real exception, I'm just going to pass in a mocked exception trace into the new private method.
It's a good idea to validate that method in a unit test or 2 anyway to ensure it stays up to date with the `Backtrace` class.

GuySartorelli:
Added backtrace test

GuySartorelli:
CI failures are because there's no version of `silverstripe/installer` with constraints that are happy with the `4.2.x-dev` constraint for this module.

emteknetnz (approved):
Confirmed all the graphql unit-tests are passing locally

PR #527, opened 2023-03-20
FIX: Don't 403 error when origin is on CORS-disallowed (fixes #519)

### LINKED ISSUE #519: POST request from mobile app never allowed
ISSUE:
I'm building a mobile Flutter app that uses Silverstripe GraphQL to pull data via a POST request. Once I enable CORS, the request from the app will always fail because the app doesn't send an Origin header. 

Please note that the app doesn't send a preflight check, as it isn't a browser, just a plain POST request. But the Silverstripe GraphQL Controller seems to handle a POST request in a similar manner to a preflight check, by calling the addCorsHeaders() function for it,  requiring an Origin.

So the only way to to get through is manually adding an Origin Header to the POST, but afaik this is not according to the rules, and could even be considered a spoofing attempt, if a take the app to the app store...

 https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name

As far as I understand it by now, CORS is purely a browser thing, and handled in preflight. So I think restricting the actual POST request might maybe be handled separately, but at least the Origin thing can hopefully be tackled...

Thanks, Martine

## Acceptance criteria
- Can do GraphQL request without Origin header.
- Security concerns are explicitly considered during peer review. 

## PRs
- https://github.com/silverstripe/silverstripe-graphql/pull/527

lozcalver:
This will no longer output a 403 error if origin is invalid - CORS headers are to protect things client side, not server side, so there’s no security benefit in triggering this error.

The `Access-Control-Allow-Origin` is now only output if the origin is valid. For valid origins, that matches the current behaviour. For invalid origins, this is a behavioural change (it used to 403) but this new behaviour will correctly trigger preflight errors:

> Access to XMLHttpRequest at 'http://my-graphql-server.test/' from origin 'http://my-origin.test' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.

One other minor behavioural change here - if developer has opted to allow all origins, this will now explicitly output `Access-Control-Allow-Origin: *` instead of `Access-Control-Allow-Origin: <current origin>`

## Issue
- #519

### DIFF
--- a/src/Controller.php
+++ b/src/Controller.php
@@ -184,17 +184,24 @@ class Controller extends BaseController
             return $response;
         }
 
-        // Calculate origin
+        // Get origin - only one host name is allowed in the Allow-Origin header, so we must return the current origin
         $origin = $this->getRequestOrigin($request);
 
-        // Check if valid
+        // Only output an Allow-Origin header if the current origin is a valid one
         $allowedOrigins = (array)$corsConfig['Allow-Origin'];
-        $originAuthorised = $this->validateOrigin($origin, $allowedOrigins);
-        if (!$originAuthorised) {
-            $this->httpError(403, "Access Forbidden");
+        if ($this->validateOrigin($origin, $allowedOrigins)) {
+            if ($corsConfig['Allow-Origin'] === '*') {
+                // Any origin is allowed
+                $response->addHeader('Access-Control-Allow-Origin', '*');
+            } else {
+                // Specific origins allowed - only one can be output at a time, so we have to output the current one
+                $response->addHeader('Access-Control-Allow-Origin', $origin);
+                // If specific allowed origins are set, the response headers will vary by request origin, so use the
+                // Vary header to tell browsers/CDNs that
+                $response->addHeader('Vary', 'Origin');
+            }
         }
 
-        $response->addHeader('Access-Control-Allow-Origin', $origin);
         $response->addHeader('Access-Control-Allow-Headers', $corsConfig['Allow-Headers']);
         $response->addHeader('Access-Control-Allow-Methods', $corsConfig['Allow-Methods']);
         $response->addHeader('Access-Control-Max-Age', $corsConfig['Max-Age']);


src/Controller.php
```diff
@@ -184,17 +184,24 @@ public function addCorsHeaders(HTTPRequest $request, HTTPResponse $response): HT
...
+        if ($this->validateOrigin($origin, $allowedOrigins)) {
+            if ($corsConfig['Allow-Origin'] === '*') {
+                // Any origin is allowed
+                $response->addHeader('Access-Control-Allow-Origin', '*');
+            } else {
+                // Specific origins allowed - only one can be output at a time, so we have to output the current one
+                $response->addHeader('Access-Control-Allow-Origin', $origin);
+                // If specific allowed origins are set, the response headers will vary by request origin, so use the
+                // Vary header to tell browsers/CDNs that
+                $response->addHeader('Vary', 'Origin');
```
lozcalver, line 201: This is new, based on the advice here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin#cors_and_caching

lozcalver:
bump

GuySartorelli:
The issue attached to this PR is in our "backlog with PRs" column and will be refined in a future backlog refinement session. We're still pretty busy getting the CMS 5 release ready so we haven't had a chance to look at this properly yet.

emteknetnz (changes requested):
Looking at the the original issue this was to allow for use case where there is no Origin header being sent.

It seems like you can **almost** already do that by update the config on the private static `$cors['Allow-Origin']` to include a `*` in there to allow this? `validateOrigin()` will return true if it encounters a `*` -- however `$origin` needs to be non-empty which is what prevents this now

This PR seems like it's completely disabling Origin validation though, based on the discussion in the original issue it looks like it's assumption that doing origin validation at this point is wrong?  I'm not so sure it is?  Seems like we're better of keeping this existing origin validation.

I think all we really need to do in this PR is instead update `validateOrigin()` to return `true` if origin is empty but there's a `*` in the static config

emteknetnz:
@kinglozzer Let me know if you're going to continue with this PR, otherwise I'll close this one for now

lozcalver:
Not likely to have time soon 😔

PR #530, opened 2023-03-22
ENH Rebuild schema if file missing

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-graphql/issues/500

A better version of https://github.com/silverstripe/silverstripe-graphql/pull/521, which was [reverted](https://github.com/silverstripe/silverstripe-graphql/pull/526) due to [CVE-2023-28104](https://www.silverstripe.org/download/security-releases/cve-2023-28104)

This version includes a configurable rate-limit that's defaulted to 15 minutes

### DIFF
--- a/src/Schema/Storage/AbstractTypeRegistry.php
+++ b/src/Schema/Storage/AbstractTypeRegistry.php
@@ -12,10 +12,29 @@ use SilverStripe\Core\Injector\Injector;
 use SilverStripe\GraphQL\Schema\SchemaBuilder;
 use SilverStripe\GraphQL\Schema\Exception\EmptySchemaException;
 use SilverStripe\Control\Controller;
+use SilverStripe\Core\Config\Configurable;
 use SilverStripe\GraphQL\Controller as GraphQLController;
 
 abstract class AbstractTypeRegistry
 {
+    use Configurable;
+
+    /**
+     * Whether to rebuild the schema if there is a mysteriously missing schema file.
+     * This is a fallback in addition to the regular fallback of building the schema
+     * if there is no schema, if autobuildSchema is enabled.
+     * This fallback is for edge case where a __graphql-schema.php file exists, though
+     * the schema files themselves are missing.
+     */
+    private static bool $rebuild_on_missing_schema_file = true;
+    
+    /**
+     * Minimum number of seconds between rebuilding the schema if a schema file is missing.
+     * If set too low this may make the site vulnerable to the schema being repeatedly rebuilt
+     * via specially crafted graphql requests
+     */
+    private static int $rebuild_on_missing_schema_file_minimum_interval = 900;
+
     protected static $types = [];
 
     /**
@@ -28,10 +47,7 @@ abstract class AbstractTypeRegistry
         try {
             return static::fromCache($typename);
         } catch (Exception $e) {
-            if (!Controller::has_curr() ||
-                !(Controller::curr() instanceof GraphQLController) ||
-                !Controller::curr()->autobuildEnabled()
-            ) {
+            if (preg_match('/(Missing|Unknown) graphql/', $e->getMessage()) || !self::canRebuildOnMissing()) {
                 throw $e;
             }
             // Try to rebuild the whole schema as fallback.
@@ -43,12 +59,49 @@ abstract class AbstractTypeRegistry
             $schema = $builder->boot($key);
             try {
                 $builder->build($schema, true);
+                $path = self::getRebuildOnMissingPath();
+                file_put_contents($path, time());
             } catch (EmptySchemaException $e) {
                 // noop
             }
             // Attempt to return again now the schema has been rebuilt.
             return static::fromCache($typename);
         }
+    }
+
+    private static function canRebuildOnMissing(): bool
+    {
+        if (!Controller::has_curr() ||
+            !(Controller::curr() instanceof GraphQLController) ||
+            !Controller::curr()->autobuildEnabled()
+        ) {
+            return false;
+        }
+        if (!self::config()->get('rebuild_on_missing_schema_file')) {
+            return false;
+        }
+        $path = self::getRebuildOnMissingPath();
+        if (!(is_writable(dirname($path)))) {
+            return false;
+        }
+        if (file_exists($path)) {
+            $lastRebuildTimestamp = (int) trim(file_get_contents($path));
+            $minInterval = self::config()->get('rebuild_on_missing_schema_file_minimum_interval');
+            if ((time() - $lastRebuildTimestamp) < $minInterval) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private static function getRebuildOnMissingPath(): string
+    {
+        return static::getSourceDirectory() . DIRECTORY_SEPARATOR . self::getRebuildOnMissingFilename();
+    }
+
+    private static function getRebuildOnMissingFilename(): string
+    {
+        return '__rebuild_on_missing';
     }
 
     abstract protected static function getSourceDirectory(): string;
--- a/tests/Schema/AbstractTypeRegistryTest.php
+++ b/tests/Schema/AbstractTypeRegistryTest.php
@@ -1,0 +1,169 @@
+<?php
+
+namespace SilverStripe\GraphQL\Tests\Schema;
+
+use GraphQL\Type\Definition\AbstractType;
+use SilverStripe\Control\Controller;
+use SilverStripe\Dev\SapphireTest;
+use SilverStripe\GraphQL\Schema\Storage\AbstractTypeRegistry;
+use SilverStripe\GraphQL\Controller as GraphQLController;
+use Symfony\Component\Filesystem\Filesystem;
+use ReflectionObject;
+use stdClass;
+use SilverStripe\Control\Session;
+
+class AbstractTypeRegistryTest extends SapphireTest
+{
+    /**
+     * Putting a .graphql-generated folder in a unit-test folder is similar to what IntegrationTest.php does
+     * It's important that we don't overwrite the real application /.graphql-generated folder
+     */
+    public const SOURCE_DIRECTORY = __DIR__ . '/.graphql-generated';
+
+    protected function setUp(): void
+    {
+        $dir = self::SOURCE_DIRECTORY;
+        if (!file_exists($dir)) {
+            mkdir($dir);
+        }
+    }
+
+    protected function tearDown(): void
+    {
+        $dir = self::SOURCE_DIRECTORY;
+        if (file_exists($dir)) {
+            $fs = new Filesystem();
+            $fs->remove($dir);
+        }
+        // ensure that any GraphqlController added to controller_stack is removed
+        if (Controller::has_curr() && (Controller::curr() instanceof GraphQLController)) {
+            Controller::curr()->popCurrent();
+        }
+    }
+
+    public function testRebuildOnMissingPath()
+    {
+        list($registry, $_, $getRebuildOnMissingPathMethod, $getRebuildOnMissingFilename) = $this->getInstance();
+        $this->assertSame(
+            self::SOURCE_DIRECTORY . '/' . $getRebuildOnMissingFilename->invoke($registry),
+            $getRebuildOnMissingPathMethod->invoke($registry)
+        );
+    }
+
+    /**
+     * @dataProvider provideRebuildOnMissing
+     */
+    public function testRebuildOnMissing(
+        bool $controller,
+        bool $autobuild,
+        bool $config,
+        bool $interval,
+        bool $expected
+    ): void {
+        list($registry, $canRebuildOnMissingMethod, $_, $getRebuildOnMissingFilename) = $this->getInstance();
+        $graphqlController = new GraphQLController('test');
+
+        // autobuild
+        $graphqlController->setAutobuildSchema($autobuild);
+
+        // controller
+        if ($controller) {
+            // Make it so that Controller::curr() returns a GraphQLController
+            $fakeSession = new Session([]);
+            $request = Controller::curr()->getRequest();
+            $request->setSession($fakeSession);
+            $graphqlController->setRequest($request);
+            $graphqlController->pushCurrent();
+        }
+
+        // config
+        AbstractTypeRegistry::config()->set('rebuild_on_missing_schema_file', $config);
+
+        // interval
+        if ($interval) {
+            // put in a timesteamp well in the past so that it's above the interval threshold
+            $time = 5000;
+        } else {
+            // put in a timestamp that is in the future, so it can never be above the interval threshold
+            $time = time() + 100;
+        }
+        file_put_contents(self::SOURCE_DIRECTORY . '/' . $getRebuildOnMissingFilename->invoke($registry), $time);
+
+        // assert
+        $this->assertSame($expected, $canRebuildOnMissingMethod->invoke($registry));
+    }
+
+    public function provideRebuildOnMissing(): array
+    {
+        // controller = current controller is a GraphQLController
+        // autobuild = if autobuild is enabled
+        // config = rebuild_on_missing_schema_file config is set to true
+        // interval = whether it will pass the interval in seconds test
+        return [
+            [
+                'controller' => false,
+                'autobuild' => true,
+                'config' => true,
+                'interval' => true,
+                'expected' => false,
+            ],
+            [
+                'controller' => true,
+                'autobuild' => false,
+                'config' => true,
+                'interval' => true,
+                'expected' => false,
+            ],
+            [
+                'controller' => true,
+                'autobuild' => true,
+                'config' => false,
+                'interval' => true,
+                'expected' => false,
+            ],
+            [
+                'controller' => true,
+                'autobuild' => true,
+                'config' => true,
+                'interval' => false,
+                'expected' => false,
+            ],
+            [
+                'controller' => true,
+                'autobuild' => true,
+                'config' => true,
+                'interval' => true,
+                'expected' => true,
+            ],
+        ];
+    }
+
+    private function getInstance()
+    {
+        $registry = new Class extends AbstractTypeRegistry
+        {
+            protected static function getSourceDirectory(): string
+            {
+                return AbstractTypeRegistryTest::SOURCE_DIRECTORY;
+            }
+
+            protected static function getSourceNamespace(): string
+            {
+                return '';
+            }
+        };
+        $reflector = new ReflectionObject($registry);
+        $canRebuildOnMissingMethod = $reflector->getMethod('canRebuildOnMissing');
+        $canRebuildOnMissingMethod->setAccessible(true);
+        $getRebuildOnMissingPathMethod = $reflector->getMethod('getRebuildOnMissingPath');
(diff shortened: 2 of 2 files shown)


src/Schema/Storage/AbstractTypeRegistry.php
```diff
@@ -12,10 +12,33 @@
...
+use SilverStripe\Core\Config\Configurable;
 use SilverStripe\GraphQL\Controller as GraphQLController;
 
+use function PHPUnit\Framework\isWritable;
+
 abstract class AbstractTypeRegistry
 {
+    use Configurable;
+
+    public const REBUILD_ON_MISSING_FILENAME = '__rebuild_on_missing';
```
michalkleiner: Does the constant need to be public?

src/Schema/Storage/AbstractTypeRegistry.php
```diff
@@ -12,10 +12,33 @@
...
+
+    /**
+     * Whether to rebuild the schema if there is a mysteriously missing schema file
+     * This is a fallback in addition the to regular fallback of building the schema
```
michalkleiner:
```suggestion
     * This is a fallback in addition to the regular fallback of building the schema
```

src/Schema/Storage/AbstractTypeRegistry.php
michalkleiner: Ah, for the test. I would probably swap the visibility around and have the constant private or actually just put the value directly in `getRebuildOnMissingPath` and have that method public.

michalkleiner: That would probably have other complications in the test.. all good. Just rubber ducking here.

src/Schema/Storage/AbstractTypeRegistry.php
emteknetnz: Updated

src/Schema/Storage/AbstractTypeRegistry.php
```diff
@@ -12,10 +12,31 @@
...
 use SilverStripe\GraphQL\Controller as GraphQLController;
 
 abstract class AbstractTypeRegistry
 {
+    use Configurable;
+
+    public const REBUILD_ON_MISSING_FILENAME = '__rebuild_on_missing';
+
+    /**
+     * Whether to rebuild the schema if there is a mysteriously missing schema file
```
ChloeSartorelli:
```suggestion
     * Whether to rebuild the schema if there is a mysteriously missing schema file.
```

src/Schema/Storage/AbstractTypeRegistry.php
```diff
@@ -12,10 +12,31 @@
...
+     * This is a fallback in addition to the regular fallback of building the schema
+     * if there is no schema, if autobuildSchema is enabled.
+     * This fallback is for edge case where a __graphql-schema.php file exists, though
+     * the schema files themselves are missing.
+     */
+    private static bool $rebuild_on_missing_schema_file = true;
+    
+    /**
+     * Minimum number of seconds between rebuilding the schema if a schema file is missing
```
ChloeSartorelli:
```suggestion
     * Minimum number of seconds between rebuilding the schema if a schema file is missing.
```

src/Schema/Storage/AbstractTypeRegistry.php
```diff
@@ -43,6 +61,8 @@ public static function get(string $typename)
             $schema = $builder->boot($key);
             try {
                 $builder->build($schema, true);
+                $path = self::getRebuildOnMissingPath();
+                file_put_contents($path, time());
```
ChloeSartorelli, line 63: The whole point of this is that we can't rely on files in this folder existing when they should - which means we can't rely on _this_ file existing there when it should, either.

But at the same time, we can't store it in the silverstripe-cache dir (either indirectly via the cache API or directly with file_put_contents), because that might be shared between servers and we want this to be a "per graphql cache dir" operation.... so I think this is the best we can do.

Not merging just yet so you have a chance to respond if you have anything to add about this approach.

src/Schema/Storage/AbstractTypeRegistry.php
ChloeSartorelli: Won't hold off merging for this, though, if it's not done.

src/Schema/Storage/AbstractTypeRegistry.php
ChloeSartorelli: Won't hold off merging for this, though, if it's not done.

tests/Schema/AbstractTypeRegistryTest.php
```diff
@@ -0,0 +1,178 @@
...
+                if (((static::$controller_stack[0] ?? new stdClass()) instanceof GraphQLController)) {
+                    array_shift(static::$controller_stack);
+                }
+            }
+        })->removeGraphqlController();
+    }
+
+    public function testRebuildOnMissingPath()
+    {
+        list($registry, , $method) = $this->getInstance();
```
ChloeSartorelli: Is that extra comma there to avoid some second value? If so, put a variable name there (`_` is often used) to indicate that the comma is intentional. It looks like a typo right now.

tests/Schema/AbstractTypeRegistryTest.php
```diff
@@ -0,0 +1,178 @@
...
+        // ensure that any GraphqlController added to controller_stack is removed
+        (new Class extends Controller
+        {
+            public function removeGraphqlController()
+            {
+                if (((static::$controller_stack[0] ?? new stdClass()) instanceof GraphQLController)) {
+                    array_shift(static::$controller_stack);
+                }
+            }
+        })->removeGraphqlController();
```
ChloeSartorelli: `(new ClassName)->` does work, but it's pretty esoteric and makes it look like there's some extra magic going on. I'd prefer the object be captured in a variable, and then call the method on the variable.
That said, I won't hold back merging if you don't change this.

Also same thing in testRebuildOnMissing() below.

Note: This only applies if the controller API change I suggested doesn't work.

tests/Schema/AbstractTypeRegistryTest.php
```diff
@@ -0,0 +1,178 @@
...
+            $fs->remove($dir);
+        }
+        // ensure that any GraphqlController added to controller_stack is removed
+        (new Class extends Controller
+        {
+            public function removeGraphqlController()
+            {
+                if (((static::$controller_stack[0] ?? new stdClass()) instanceof GraphQLController)) {
+                    array_shift(static::$controller_stack);
+                }
```
ChloeSartorelli: Can't you just do:
```php
if (Controller::has_curr() && (Controller::curr() instanceof GraphQLController) {
    Controller::curr()->popCurrent();
}
```
If this works, please do it instead - prefer using existing API where it exists, and avoids having to have this anonymous class altogether.

src/Schema/Storage/AbstractTypeRegistry.php
```diff
@@ -12,10 +12,31 @@
...
 use SilverStripe\GraphQL\Schema\Exception\EmptySchemaException;
 use SilverStripe\Control\Controller;
+use SilverStripe\Core\Config\Configurable;
 use SilverStripe\GraphQL\Controller as GraphQLController;
 
 abstract class AbstractTypeRegistry
 {
+    use Configurable;
+
+    public const REBUILD_ON_MISSING_FILENAME = '__rebuild_on_missing';
```
ChloeSartorelli: This doesn't need to be public - your test that requires it instantiates an anonymous subclass of `AbstractTypeRegistry` - just add a new public method to that anonymous class that returns the value of this constant.

tests/Schema/AbstractTypeRegistryTest.php
ChloeSartorelli:
```suggestion
        list($registry, , $getRebuildOnMissingPath) = $this->getInstance();
```
Make it really clear what method you're calling, without having to go look at the implementation of `getInstance()`

Feel free to also add "method" or "reflection" in the variable name if you want to, to distinguish this reflection method from the boolean result of calling the method.

tests/Schema/AbstractTypeRegistryTest.php
```diff
@@ -0,0 +1,178 @@
...
+     */
+    public function testRebuildOnMissing(
+        string $message,
+        bool $controller,
+        bool $autobuild,
+        bool $config,
+        bool $interval,
+        bool $expected
+    ): void {
+        list($registry, $method) = $this->getInstance();
```
ChloeSartorelli:
```suggestion
        list($registry, $canRebuildOnMissing) = $this->getInstance();
```
Make it clear what the method is

tests/Schema/AbstractTypeRegistryTest.php
```diff
@@ -0,0 +1,178 @@
...
+
+        // controller
+        if ($controller) {
+            // Make it so that Controller::curr() returns a GraphQLController
+            (new Class extends Controller
+            {
+                public function setCurrentController(Controller $controller)
+                {
+                    array_unshift(static::$controller_stack, $controller);
+                }
```
ChloeSartorelli: Cant you just call `$graphqlConroller->pushCurrent()`? If the problem is there's no session, just instantiate a dummy `Session` object and add it to the controller.

tests/Schema/AbstractTypeRegistryTest.php
```diff
@@ -0,0 +1,178 @@
...
+    public function testRebuildOnMissing(
+        string $message,
+        bool $controller,
+        bool $autobuild,
+        bool $config,
+        bool $interval,
+        bool $expected
+    ): void {
+        list($registry, $method) = $this->getInstance();
+        $graphqlConroller = new GraphQLController('test');
```
ChloeSartorelli:
```suggestion
        $graphqlController = new GraphQLController('test');
```

tests/Schema/AbstractTypeRegistryTest.php
```diff
@@ -0,0 +1,178 @@
...
+
+    public function provideRebuildOnMissing(): array
+    {
+        // controller = current controller is a GraphQLController
+        // autobuild = if autobuild is enabled
+        // config = rebuild_on_missing_schema_file config is set to true
+        // interval = whether it will pass the interval in seconds test
+        return [
+            [
+                'message' => 'Controller false',
```
ChloeSartorelli: This (and the other messages below) is not a clear message to output with a PHPunit failure. What is actually being tested here, so that someone debugging a failure knows what needs to be fixed? What does "controller false" imply?

I'm guessing this should be something like "If the current controller is not a GraphQLController the schema should not be rebuilt"

tests/Schema/AbstractTypeRegistryTest.php
```diff
@@ -0,0 +1,178 @@
...
+                'interval' => true,
+                'expected' => false,
+            ],
+            [
+                'message' => 'Autobuild false',
+                'controller' => true,
+                'autobuild' => false,
+                'config' => false,
+                'interval' => true,
+                'expected' => false,
```
ChloeSartorelli: Should config be true here? Otherwise the result could be false just because config is false, right?

GuySartorelli (changes requested):
Works well locally - I wasn't able to trigger [CVE-2023-28104](https://www.silverstripe.org/download/security-releases/cve-2023-28104) because the interval stopped the cache from being rebuilt. After enough time passed that the interval was exceeded, it rebuilt schema again.

Some changes to make

src/Schema/Storage/AbstractTypeRegistry.php
michalkleiner: So when the schema doesn't exist and the write permissions to the `.graphql-generated` folder are lost, this is still subject to DDoS, right? Even if say a schema was committed to the repo and deployed but used a different than expected obfuscator.

tests/Schema/AbstractTypeRegistryTest.php
michalkleiner: Can also be done as `[$registry, $_, $method] = $this->getInstance();` using the array syntax.

michalkleiner: Based on other uses of the `$this->getInstance()`, it seems like the comma is extra by mistake. The array syntax suggestion is still applicable though.

tests/Schema/AbstractTypeRegistryTest.php
emteknetnz: I dunno, it's messy either way.  
I think it would something like

```php
$fakeSession = new Session();
$graphqlController = new GraphQLController();
$request = Controller::curr()->getRequest();
$request->setSession($fakeSession);
$graphqlController->setRequest($request);
$graphqlController->pushCurrent();
```
That's way worse IMO, basically can't tell what's going on there.

tests/Schema/AbstractTypeRegistryTest.php
emteknetnz: $myInst = (new Class ...
{
  //...
});
$myInst->removeGraphqlController();

Surely is really no improvement

In my head makes more sense to NOT store anonymous classes (and functions) in variables

src/Schema/Storage/AbstractTypeRegistry.php
ChloeSartorelli: If the entire schema doesn't exist, this doesn't apply - that's a (slightly) different mechanism and will only trigger a rebuild once. This mechanism is for if _that_ mechanism doesn't get triggered because it's just the one specific shvema file that's missing.

If write permissions are lost then the will not rebuild (see is writable check in `canRebuildOnMissing()`)

If a different obfuscator was used this should trigger once - and only rebuild if all the checks in the above mentioned method pass.

tests/Schema/AbstractTypeRegistryTest.php
ChloeSartorelli: Imo that's much better, because we're using established API. It's clear to me what's happening there but making it into a well named private method with a clear phpdoc can help further clarify it.

src/Schema/Storage/AbstractTypeRegistry.php
michalkleiner: Right, I didn't put the write check in the `canRebuildOnMissing` method together with this, all good.

tests/Schema/AbstractTypeRegistryTest.php
michalkleiner: `src/Dev/TestSession.php` has nearly identical bit, so I'd support sticking with that...

```
$this->session = Injector::inst()->create(Session::class, []);
$this->cookies = Injector::inst()->create(Cookie_Backend::class);
$request = new HTTPRequest('GET', '/');
$request->setSession($this->session());
$this->controller = new Controller();
$this->controller->setRequest($request);
$this->controller->pushCurrent();
$this->controller->doInit();
```

michalkleiner: Creating a dummy request with a dummy session is quite common in tests.

src/Schema/Storage/AbstractTypeRegistry.php
emteknetnz: Updated

src/Schema/Storage/AbstractTypeRegistry.php
emteknetnz: Updated

tests/Schema/AbstractTypeRegistryTest.php
emteknetnz: Extra comma was intentional, have added $_

tests/Schema/AbstractTypeRegistryTest.php
emteknetnz: It won't, because Controller::curr() is a Controller

This is only a unit test, we don't need to overthink this

ChloeSartorelli: But you added a `GraphQLController` to the stack.
While this is a unit test, it should both be readable and as standard as we can reasonably make it.
Having esoteric or unusual syntax makes it look like something special is happening, and that the normal API won't work.

Not only does this add confusion, it also means that if we change the implementation details we might at best have to change this test and at worst have a test that's passing when the actual behaviour doesn't work.

src/Schema/Storage/AbstractTypeRegistry.php
emteknetnz: Needs to be at least protected for that to work.   Also can't use reflection as https://www.php.net/manual/en/reflectionclass.getconstant.php simply returns the value, rather that property/method which lets you to call setAccessible()

Not much difference between protected vs public as they're both API, so may as well just leave public

tests/Schema/AbstractTypeRegistryTest.php
emteknetnz: Updated

tests/Schema/AbstractTypeRegistryTest.php
emteknetnz: Updated

tests/Schema/AbstractTypeRegistryTest.php
emteknetnz: Updated

tests/Schema/AbstractTypeRegistryTest.php
emteknetnz: I just deleted the message, it was basically just for debugging

tests/Schema/AbstractTypeRegistryTest.php
emteknetnz: Updated

tests/Schema/AbstractTypeRegistryTest.php
emteknetnz: Updated

tests/Schema/AbstractTypeRegistryTest.php
emteknetnz: Updated

tests/Schema/AbstractTypeRegistryTest.php
emteknetnz: Has been refactored away

src/Schema/Storage/AbstractTypeRegistry.php
emteknetnz: Changed to private method and set accessible via reflection

Using a method as private properties and constants can't seem to be set accessible via reflection, though methods can

GuySartorelli (approved):
Looks good to me.
I've rerun failed jobs now that the issue with admin is resolved. Merge on green.

GuySartorelli:
behat tests are failing repeatedly

emteknetnz:
It's an existing failure - https://github.com/silverstripe/silverstripe-graphql/actions/runs/4527398776/jobs/7973214759

Note: After approval I added a little bit to filter out Exception so that they match a particular message.  This isn't as good as using a new Exception type, but I didn't want to change the API via changing the Exception thrown, and the message is very unlikely to ever change - https://github.com/silverstripe/silverstripe-graphql/compare/f796b4fd49fbbe6d1b2062d239842a9eea668e6d..0b876c8e2c50f7d69c1eb4805edb5e1de49e5888

GuySartorelli (approved):
Approved a second time.
I'd have preferred a new exception - maybe for CMS 6 we can look at adding explicit exceptions in appropriate places for this module. For now the message check is sufficient.

PR #531, opened 2023-04-04
FIX Reverse conditional

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/693

Behat test failures are existing


GuySartorelli (approved):
Good catch

Issue #529, opened 2023-03-22
Store GraphQL v5 schema in silverstripe-cache folder


GuySartorelli:
This has come up time and time again - most recently in https://github.com/silverstripe/silverstripe-graphql/issues/465

The main argument against it at the time was that we were very close to releasing the stable release of GraphQL v4 and we didn't want to poke the bear.

We now find ourselves in a situation where we're creating a new major release of this module, and we have the opportunity to reconsider what the default directory is for the graphql generated schema.

We've agreed internally that for CMS 5 the graphql schema should be included in the silverstripe-cache directory by default.

## Acceptance criteria
- [x] By default, the graphql schema is stored where it currently is
- [x] If you set it to a blank string, it will go to the silverstripe-cache folder
- [x] Developers can override the location of the generated graphql schema by setting the `SilverStripe\GraphQL\Schema\Storage\CodeGenerationStore.dirname` configuration property (i.e. this config property still works as expected)
- [x] Schemas for multiple independent websites on the same server must not collide
- [ ] Test on on prod-like webservers, including multi-server environments where the temp folder may not be shared by the instances behind a load balancer
- [x] The CMS 5.1 changelog calls out that the location of this directory can be configured
- [x] The [Silverstripe Cloud graphql documentation](https://servicedesk.silverstripe.cloud/support/solutions/articles/75000063363-graphql) has accurate information for both CMS4 and CMS 5
- [x] This new set up is tested on SC and SCPS

## Notes
- A [proof of concept PR](https://github.com/silverstripe/silverstripe-graphql/pull/467) was made for this last year. It might be useful as a basis for this work - but don't add the warning messaging that PR included.
- This will be targeted at CMS 5.1
- Create a follow up card to explore possibility of updating the installer to use the new behaviour out of the box

## PRs
- https://github.com/silverstripe/silverstripe-graphql/pull/534
- https://github.com/silverstripe/developer-docs/pull/257

michalkleiner:
We are in feature freeze for CMS 5.0, right? Would this go against it if it was done for 5.0.0?

michalkleiner:
If the overridden config still supports relative paths, then that will suffice for use cases where developers want to commit the generated files, but if it doesn't, we should still support the use case when the schema is generated e.g. by CI pipeline and deployed as part of the application.

GuySartorelli:
> We are in feature freeze for CMS 5.0, right? Would this go against it if it was done for 5.0.0?

[...]

PR #534, opened 2023-05-03
NEW Enable storing graphql schema in `silverstripe-cache/`

GuySartorelli:
## A note about testing
The acceptance criteria for testing this was originally written with the idea that this would be the _default_ behaviour, which is no longer the plan.

Note that while I _have_ tested this in both SC and SCPS, all I can verify is that there weren't any obvious problems from a user perspective. I have no way to validate what was happening from a filesystem POV in that environment.

## Issue
- https://github.com/silverstripe/silverstripe-graphql/issues/529


src/Schema/Storage/CodeGenerationStore.php
```diff
@@ -395,22 +374,31 @@ public static function getTemplateDir(): string
...
             $this->name
         );
     }
 
     private function getTempDirectory(): string
     {
         return Path::join(
-            TEMP_FOLDER,
-            $this->config()->get('dirName'),
+            TEMP_PATH,
```
ChloeSartorelli, line 397: Note that `TEMP_FOLDER` and `TEMP_PATH` evaluate to the same value - but `TEMP_FOLDER` is deprecated.

src/Schema/Storage/CodeGenerationStore.php
```diff
@@ -103,38 +100,19 @@ public function persistSchema(StorableSchema $schema): void
...
+        if ($this->dirNameIsDefined()) {
+            $dest = $this->getDirectory();
+            if ($fs->exists($dest)) {
+                $logger->info('Moving current schema to temp folder');
+                $fs->mirror($dest, $temp);
+            } else {
+                $logger->info('Creating new schema');
+                $this->createNewTempSchema($fs, $temp);
             }
+        } else {
```
emteknetnz, line 118: Seems a bit weird that we're splitting the code path here?  Seems that `$dest->getDirectory` would return something like `/tmp/silverstripe-.../.graphql-generated`, so why aren't we aren't we just treating the that dir the same way we currently treat `BASE_PATH/.graphql-generated`.  In particular there's the line `$fs->mirror($dest, $temp);` which is only present in the `dirNameIsDefined()` code path, I'm not sure what this line is supposed to do, though presumably _something_, though it seems concerning to me that it's not in the `!dirNameIsDefined()` code path

src/Schema/Storage/CodeGenerationStore.php
```diff
@@ -246,16 +224,17 @@ public function persistSchema(StorableSchema $schema): void
         }
 
         // Move the new schema into the proper destination
-        if ($fs->exists($dest)) {
-            $logger->info('Deleting current schema');
-            $fs->remove($dest);
+        if ($this->dirNameIsDefined()) {
```
emteknetnz, line 233: As noted above, seems weird that we're have a different code paths depending on which directory the schema is stored in, ideally we'd treat everything the same

src/Schema/Storage/CodeGenerationStore.php
```diff
@@ -458,4 +446,34 @@ private function toCode(string $rawCode): string
         $code = preg_replace("/(^[\r\n]*|[\r\n]+)[\s\t]*[\r\n]+/", "\n", $rawCode ?? '');
         return "<?php\n\n /** GENERATED CODE -- DO NOT MODIFY **/\n\n{$code}";
     }
+
+    private function createNewTempSchema(FileSystem $fs, string $tempDir)
```
emteknetnz:
```suggestion
    private function createNewTempSchema(FileSystem $fs, string $tempDir): void
```

src/Schema/Storage/CodeGenerationStore.php
```diff
@@ -13,19 +13,13 @@
 use SilverStripe\Core\Path;
 use SilverStripe\GraphQL\Schema\Exception\EmptySchemaException;
 use SilverStripe\GraphQL\Schema\Exception\SchemaNotFoundException;
-use SilverStripe\GraphQL\Schema\Logger;
```
emteknetnz: Prefer not to remove all of these in this PR since it's not related to the current issue.  Presumably they're not actually used and this is obvious in your IDE, though it's not obvious when looking at this pull-request in github

If we are to remove unused imports, I'd much prefer to bulk to this in a dedicated issue

emteknetnz (changes requested):

src/Schema/Storage/CodeGenerationStore.php
ChloeSartorelli: Fair enough. Easy enough to view the file as a whole and ctrl+f to check against them but I'm fine with removing these changes from this PR.

ChloeSartorelli: Done

src/Schema/Storage/CodeGenerationStore.php
ChloeSartorelli: Done

src/Schema/Storage/CodeGenerationStore.php
ChloeSartorelli: > Seems a bit weird that we're splitting the code path here? ... why aren't we aren't we just treating the that dir the same way we currently treat `BASE_PATH/.graphql-generated`

The schema is _always_ initially built in the temporary directory. `getTempDirectory()` gets a directory path inside the silverstripe-cache directory.
Since the temp directory is already in `silverstripe-cache`, I opted to just use that as the directory for when `dirName` is set to an empty string/null. And since the temp directory then _is_ the final destination for the schema, there's no sense in copying the files from itself to itself. This is also why the second branch logs 'Using temporary directory for schema' and 'Creating new schema which will override any old schema'.

> In particular there's the line `$fs->mirror($dest, $temp);` which is only present in the `dirNameIsDefined()` code path, I'm not sure what this line is supposed to do

`$fs->mirror($dest, $temp);` copies all of the files from the _non temporary_ directory (normally `.graphql-generated/`) into the temp directory (hence why it logs 'Moving current schema to temp folder')

ChloeSartorelli: Is the above reasonable to you? Or would you prefer we still have a temporary directory _separate_ to the destination directory even when the destination directory is inside `silverstripe-cache/`? Happy to do either.

emteknetnz: OK makes sense, leave it as is

emteknetnz (approved):

emteknetnz:
Tested locally, works good

emteknetnz:
Opps accidentally closed it - I somehow mistook the close button for the merge button :-)

Issue #535, opened 2023-05-25
NestedInputBuilder (sort and filter plugins) doesn't work with nested '*'


MasonD:
If a sort plugin configuration is, say
```yml
plugins:
  sort:
    fields:
      id: true
      subQuery:
        '*': true
```
This does not work, because the relevant code in NestedInputBuilder is 
```php
        foreach ($fields as $fieldName => $data) {
            if ($fieldName === Schema::ALL) {
                $this->buildAllFieldsConfig($type);
            }
            if ($data === false) {
                continue;
            }
```
which computes and then immediately discards `buildAllFieldsConfig` for the nested type, doing absolutely nothing.

PR should be inc. soon

### PRs
- https://github.com/silverstripe/silverstripe-graphql/pull/536
- https://github.com/silverstripe/silverstripe-graphql/pull/538

GuySartorelli:
I'm not able to reproduce this issue.

Using the following graphql models schema:

```yml
SilverStripe\Security\Group:
  fields:
    '*': true

SilverStripe\Security\Permission:
  fields:
    id: true
    group: true
  operations:
    readOne: true
    read:
      plugins:
        sort:
          fields:
            id: true
            group:
              '*': true
```

and the following query:

```graphql
query getPerms {
  readPermissions(sort: {group: {title: DESC}}) {
    nodes {
      id
      group {
        id
        title
      }
    }
  }
}
```

I see the results being sorted appropriately. I have tested sorting by other group fields and changing the sort direction as well, and I get the expected resiults.

@MasonD When you say "This does not work", what specifically doesn't work?
- Are you shown an error?
  - if so, what are the steps to reproduce that, and what is the error?
- Or does something just not work exactly how you expect it?
  - if so, what are the steps to reproduce that, what is the expected behaviour, and what are you seeing instead?

MasonD:
Oh, sorry, my bad. That does work. The issue only happens when nesting happens one level further, and including `readOne` **masks** the error because it's implicitly `sort: fields: '*'` which read then just extends.

For example with
```yaml
SilverStripe\Security\Group:
  fields:
    description: true
    sort: true
    members: true

SilverStripe\Security\Member:
  fields:
    id: true
    firstName: true
    surname: true

SilverStripe\Security\Permission:
  fields:
    id: true
    group: true
  operations:
    read:
      plugins:
        sort:
          fields:
            '*': false
            id: true
            group:
              id: true
              members:
                '*': true
```

[...]

PR #536, opened 2023-05-25
FIX: NestedInputBuilder doesn't work with nested '*'

MasonD:
Addresses #535

### DIFF
--- a/src/Schema/Services/NestedInputBuilder.php
+++ b/src/Schema/Services/NestedInputBuilder.php
@@ -97,9 +97,6 @@ class NestedInputBuilder
 
         if ($this->fields === Schema::ALL) {
             $this->fields = $this->buildAllFieldsConfig($type);
-        } elseif (isset($this->fields[Schema::ALL]) && $this->fields[Schema::ALL]) {
-            unset($this->fields[Schema::ALL]);
-            $this->fields = array_merge($this->fields, $this->buildAllFieldsConfig($type));
         }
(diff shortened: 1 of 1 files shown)


src/Schema/Services/NestedInputBuilder.php
```diff
@@ -167,6 +164,10 @@ protected function addInputTypesToSchema(
         if (!$inputType) {
             $inputType = InputType::create($inputTypeName);
         }
+        if (isset($fields[Schema::ALL]) && $fields[Schema::ALL]) {
+            unset($fields[Schema::ALL]);
+            $fields = array_merge($fields, $this->buildAllFieldsConfig($type));
+        }
         foreach ($fields as $fieldName => $data) {
             if ($fieldName === Schema::ALL) {
```
ChloeSartorelli, line 172: This if statement (`if ($fieldName === Schema::ALL) {` inside the foreach loop) is dead code since you've added `unset($fields[Schema::ALL]);` above. Please remove it.

GuySartorelli (changes requested):
Thank you for this PR, this is an excellent contribution!

This bug fix should be applied to the CMS 4 compatible version of the module as well. Please retarget this PR to the `4.3` branch. You'll need to reset your PR branch so that it doesn't carry over commits from `5`.

GuySartorelli:
I don't have permissions to directly modify this PR so it has now been superceded by https://github.com/silverstripe/silverstripe-graphql/pull/538 which includes the requested changes

PR #538, opened 2023-06-14
FIX: NestedInputBuilder doesn't work with nested '*'

GuySartorelli:
Supercedes https://github.com/silverstripe/silverstripe-graphql/pull/536 and includes requested changes mentioned in that PR

## Issue
- https://github.com/silverstripe/silverstripe-graphql/issues/535

### DIFF
--- a/src/Schema/Services/NestedInputBuilder.php
+++ b/src/Schema/Services/NestedInputBuilder.php
@@ -97,9 +97,6 @@ class NestedInputBuilder
 
         if ($this->fields === Schema::ALL) {
             $this->fields = $this->buildAllFieldsConfig($type);
-        } elseif (isset($this->fields[Schema::ALL]) && $this->fields[Schema::ALL]) {
-            unset($this->fields[Schema::ALL]);
-            $this->fields = array_merge($this->fields, $this->buildAllFieldsConfig($type));
         }
         $this->addInputTypesToSchema($type, $this->fields, null, null, $prefix);
         $rootTypeName = $prefix . $this->getTypeName($type);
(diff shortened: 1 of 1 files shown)


sabina-talipova (commented):
Is it possible to add an additional test to check the case that was stated in the initial issue.

GuySartorelli:
I thought the answer was going to be "yes, easy peasy" but it turns out that there's something weird about the way test schemas are built which differs in some way from normal schemas.... 
I've created [a new issue](https://github.com/silverstripe/silverstripe-graphql/issues/539) to cover that. Until that issue is resolved, I don't think there's any way to add a test for this - but I don't think this bug fix should be held back by that, since it is likely that will be a complex issue to resolve.

sabina-talipova (approved):
LGTM

sabina-talipova:
@GuySartorelli, I've already approved. I couldn't reproduce issue by using provided code example in main ticket. But I have a question regarding [new issue that you opened](https://github.com/silverstripe/silverstripe-graphql/issues/539). 
Does not seem to you that you get the same Exception that was mentioned in the main ticket? Did you apply your test to this solution? 
Thank you.

GuySartorelli:
> I couldn't reproduce issue by using provided code example in main ticket

Did you use [this example](https://github.com/silverstripe/silverstripe-graphql/issues/535#issuecomment-1586558963) and did you both [activate the default schema](https://docs.silverstripe.org/en/5/developer_guides/graphql/getting_started/activating_the_server/) and [tell graphql where to look for your models file](https://docs.silverstripe.org/en/5/developer_guides/graphql/getting_started/configuring_your_schema/#avoid-config-flushes)?

> I have a question regarding https://github.com/silverstripe/silverstripe-graphql/issues/539 ... Does not seem to you that you get the same Exception that was mentioned in the main ticket? Did you apply your test to this solution?

That's actually exactly the problem. In that issue, the schema example doesn't cause _any_ exception when used outside of a test context. But when used in a test, it gets the exception mentioned in that issue - which is indeed a _different exception_ than the one this PR is fixing. So there's something about the way tests set up or build schemas that's different from the normal execution, which is causing that exception when no exception should be thrown.

sabina-talipova:
> Did you use https://github.com/silverstripe/silverstripe-graphql/issues/535#issuecomment-1586558963 and did you both [activate the default schema](https://docs.silverstripe.org/en/5/developer_guides/graphql/getting_started/activating_the_server/) and [tell graphql where to look for your models file](https://docs.silverstripe.org/en/5/developer_guides/graphql/getting_started/configuring_your_schema/#avoid-config-flushes)?
Yes, I did.

PR #541, opened 2023-07-04
fix(SortPlugin): amend sorting to all fields

### LINKED ISSUE #542: `SortPlugin` only sorts based on the last field in the sort array, rather than cascading sort based on all sort fields.
GraphQL's SortPlugin only uses the last field in the sort array, rather than sorting based on all sort fields.

This is because in framework DataList each call to ->sort() actually resets previous sorts made, and the result is that only the end sort happens.

## PR
- https://github.com/silverstripe/silverstripe-graphql/pull/541

flamerohr:
Fix sorting to all fields in the args.

This is because in framework `DataList` each call to `->sort()` actually resets previous sorts made, and the result is that only the end sort happens.

## Issue
- fixes #542

### DIFF
--- a/src/Schema/Plugin/SortPlugin.php
+++ b/src/Schema/Plugin/SortPlugin.php
@@ -91,9 +91,7 @@ class SortPlugin implements FieldPlugin, SchemaUpdater
                 return null;
             }
             $sortArgs = $args[$fieldName] ?? [];
-            foreach ($sortArgs as $field => $dir) {
-                $list = $list->sort($field, $dir);
-            }
+            $list = $list->sort($sortArgs);
 
             return $list;
         };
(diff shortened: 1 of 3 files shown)


GuySartorelli:
Thank you for this. This feels like something which ought to be tested. Could you please add a unit test to validate the sort args are being correctly applied?

flamerohr:
is an existing unit test for this plugin I can start writing the tests on? I'm not familiar enough with the test setup to jump right in

GuySartorelli:
Looks like `IntegrationTest` class in `tests/Schema/IntegrationTest.php` has a `testFilterAndSort()` method which is probably an appropriate place for this. There are also some pre-existing test schemas with sorts applied:
https://github.com/search?q=repo%3Asilverstripe%2Fsilverstripe-graphql+sort+path%3A%2F%5Etests%5C%2F%2F&type=code

flamerohr:
@GuySartorelli test added, the diff looks a bit weird because I needed to tweak the other tests to accomodate to a third row added to the db

GuySartorelli (approved):

GuySartorelli:
Looks good, thank you!

PR #544, opened 2023-07-24
Run dev/build add-on only once

NightJar:
Instead of once for every DataObject that exists in the project. Although there's an early exist after the first execution runs, there are N-1 more runs checking the same thing. This isn't really a change, nor impactful... it is just a touch of code hygiene :)

Because, well, lol
image


GuySartorelli (approved):
Thank you, that's eminently sensible. I thought this was how it was already to be honest 😅
An additional benefit to this is we're not polluting the `DataObject` config space with an overly generic `enabled` property.

GuySartorelli:
Failing behat test is unrelated

Issue #524, opened 2023-03-14
[v4] Infinite loop with Injector replaced classes (subclass replacing parent class)


AljosaB:
## Affected Version
SS 4.11+
GrapQl v4

## Steps to Reproduce
**/app/src/SomePage.php**
```php
<?php

namespace App\Pages;

use Page;

class SomePage extends Page
{
    
    private static $table_name = 'SomePage';

    private static $singular_name = 'SomePage';
    private static $description = 'SomePage';

    private static $db = [
        'SomeField' => 'Int'
    ];

}
```

**/app/src/PageExtendingSomePage.php**
```php
<?php

namespace App\Pages;

class PageExtendingSomePage extends SomePage
{
    public function NewFunction()
    {
        // do something
    }
}
```

**/app/_config/injectors.yml**
```yml
SilverStripe\Core\Injector\Injector:
  App\Pages\SomePage:
    class: App\Pages\PageExtendingSomePage
```

**/app/_config/graphql.yml**
```yml
SilverStripe\Control\Director:
  rules:
    'graphql': '%$SilverStripe\GraphQL\Controller.default'

SilverStripe\GraphQL\Schema\Schema:
  schemas:
    default:
      src:
        - app/_graphql
```

**/app/_graphql/models.yml**
```yml
App\Pages\SomePage:
  fields: '*'
  operations:
    read: true
```

**Run**
```bash
/vendor/silverstripe/framework$ php cli-script.php dev/graphql/build schema=default
```

**Error**
```
[INFO]: --- Building schema "default" ---
ERROR [Emergency]: Uncaught Error: Xdebug has detected a possible infinite loop, and aborted your script with a stack depth of '256' frames at /vendor/silverstripe/framework/src/Core/Injector/SilverStripeServiceConfigurationLocator.php line 50
```

## Additional info
- it starts working the moment you comment out injectors.yml lines
- related ticket silverstripe/silverstripe-elemental#1014

## Acceptance criteria
- GraphQL schema models honour Injector overrides even for subclasses
- Schema builds complete successfully

## PRs
- https://github.com/silverstripe/silverstripe-graphql/pull/549

sukhwinder-somar:
Hi, any update on this. I am also facing same issue when upgrading from v3 to v4

maxime-rainville:
Replicated the issue. One important point here is that the thing would work fine if `PageExtendingSomePage` didn't extend `SomePage`. 

[...]

PR #549, opened 2023-08-15
FIX Handle using injected subclasses

emteknetnz:
Issue - https://github.com/silverstripe/silverstripe-graphql/issues/524

Original fix - https://github.com/wilr/silverstripe-graphql/commit/623bc9833039996c0c52f1a756b753730923e2b4

Without code fix, new unit test will either hang indefinately or go to 256 levels of recursion if xdebug is enabled

### DIFF
--- a/src/Schema/DataObject/InheritanceBuilder.php
+++ b/src/Schema/DataObject/InheritanceBuilder.php
@@ -49,15 +49,21 @@ class InheritanceBuilder
         $chain = InheritanceChain::create($modelType->getModel()->getSourceClass())
             ->hideAncestors($this->hideAncestors);
         $ancestors = $chain->getAncestralModels();
-        if (empty($ancestors)) {
+        $ancestorModel = null;
+        foreach ($ancestors as $ancestor) {
+            $ancestorModel = $this->getSchema()->findOrMakeModel($ancestor);
+            if ($ancestorModel && $ancestorModel->getName() !== $modelType->getName()) {
+                break;
+            }
+            $ancestorModel = null;
+        }
+        if (!$ancestorModel) {
             return;
         }
-        $parent = $ancestors[0];
-        $parentModel = $this->getSchema()->findOrMakeModel($parent);
         // Merge descendant fields up into the ancestor
         foreach ($modelType->getFields() as $fieldObj) {
             // If the field already exists on the ancestor with the same config, skip it
-            if ($existing = $parentModel->getFieldByName($fieldObj->getName())) {
+            if ($existing = $ancestorModel->getFieldByName($fieldObj->getName())) {
                 if ($existing->getSignature() === $fieldObj->getSignature()) {
                     continue;
                 }
@@ -66,12 +72,12 @@ class InheritanceBuilder
                 ? $fieldObj->getPropertyName()
                 : $fieldObj->getName();
             // If the field is unique to the descendant, skip it.
-            if ($parentModel->getModel()->hasField($fieldName)) {
+            if ($ancestorModel->getModel()->hasField($fieldName)) {
                 $clone = clone $fieldObj;
-                $parentModel->addField($fieldObj->getName(), $clone);
+                $ancestorModel->addField($fieldObj->getName(), $clone);
             }
         }
-        $this->fillAncestry($parentModel);
+        $this->fillAncestry($ancestorModel);
     }
 
     /**
--- a/tests/Fake/Inheritance/MyOrig.php
+++ b/tests/Fake/Inheritance/MyOrig.php
@@ -1,0 +1,17 @@
+<?php
+
+
+namespace SilverStripe\GraphQL\Tests\Fake\Inheritance;
+
+use SilverStripe\Dev\TestOnly;
+use SilverStripe\ORM\DataList;
+use SilverStripe\ORM\DataObject;
+
+class MyOrig extends DataObject implements TestOnly
+{
+    private static $db = [
+        'MyField' => 'Varchar',
+    ];
+
+    private static $table_name = 'MyOrig_test';
+}
--- a/tests/Fake/Inheritance/MySubclass.php
+++ b/tests/Fake/Inheritance/MySubclass.php
@@ -1,0 +1,16 @@
+<?php
+
+
+namespace SilverStripe\GraphQL\Tests\Fake\Inheritance;
+
+use SilverStripe\Dev\TestOnly;
+use SilverStripe\ORM\DataList;
+
+class MySubclass extends MyOrig implements TestOnly
+{
+    private static $db = [
+        'MySubclassField' => 'Varchar',
+    ];
+
+    private static $table_name = 'MySubclass_test';
+}
--- a/tests/Schema/DataObject/InheritanceBuilderTest.php
+++ b/tests/Schema/DataObject/InheritanceBuilderTest.php
@@ -21,6 +21,10 @@ use SilverStripe\GraphQL\Tests\Fake\Inheritance\C;
 use SilverStripe\GraphQL\Tests\Fake\Inheritance\C1;
 use SilverStripe\GraphQL\Tests\Fake\Inheritance\C2;
 use SilverStripe\GraphQL\Tests\Fake\Inheritance\C2a;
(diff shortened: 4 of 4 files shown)


src/Schema/DataObject/InheritanceBuilder.php
```diff
@@ -52,12 +52,17 @@ public function fillAncestry(ModelType $modelType): void
         if (empty($ancestors)) {
             return;
         }
-        $parent = $ancestors[0];
-        $parentModel = $this->getSchema()->findOrMakeModel($parent);
+        foreach ($ancestors as $ancestor) {
+            $ancestorModel = $this->getSchema()->findOrMakeModel($ancestor);
+            if (!$ancestorModel || $ancestorModel->getName() === $modelType->getName()) {
+                continue;
+            }
```
ChloeSartorelli: I'm struggling to understand this condition... how would the ancestor ever be the same as the current model? I can't do a full review on this without understanding the conditions that happens under.

emteknetnz: Condition is when Injector overrides a class with a subclass.  

Using the replication code in the original description - https://github.com/silverstripe/silverstripe-graphql/issues/524

At line 58 - you'll have the following:

```
$ancestors = [
  'App\Pages\SomePage',
  'Page',
  'Silverstripe\CMS\Model\SiteTree'
];

$ancestor = 'App\Pages\SomePage';

$modeType->getName() => 'PageExtendingSomePage"
$ancestorModel->getName() => 'PageExtendingSomePage"
```

Basically because of the injection `$ancestorModel->getName()` returns the name of the subclass, rather than the name of the baseclass

src/Schema/DataObject/InheritanceBuilder.php
```diff
@@ -52,12 +52,17 @@ public function fillAncestry(ModelType $modelType): void
...
+            break;
+        }
```
ChloeSartorelli:
```suggestion
        $ancestorModel = null;
        foreach ($ancestors as $ancestor) {
            $ancestorModel = $this->getSchema()->findOrMakeModel($ancestor);
            if ($ancestorModel && $ancestorModel->getName() !== $modelType->getName()) {
                break;
            }
            $ancestorModel = null;
        }
        if (!$ancestorModel) {
            return;
        }
```
Even if there's an ancestor in the `$ancestors` array, if we don't find an ancestor that _isn't_ identical (due to injection) to the current `$modelType`, we don't want do any of the rest of this stuff.

Also swapped the condition around, so instead of "break if we don't continue", we only break once we've found an ancestor. This is a more intuitive flow.

src/Schema/DataObject/InheritanceBuilder.php
```diff
@@ -52,12 +52,17 @@ public function fillAncestry(ModelType $modelType): void
         if (empty($ancestors)) {
             return;
         }
```
ChloeSartorelli:
```suggestion
```
No longer needed - if it's empty, we simply won't do anything in the loop, and `$ancestorModel` will remain null, so we'll be returning at that point.

src/Schema/DataObject/InheritanceBuilder.php
```diff
@@ -66,12 +71,14 @@ public function fillAncestry(ModelType $modelType): void
...
+            if ($ancestorModel->getModel()->hasField($fieldName)) {
                 $clone = clone $fieldObj;
-                $parentModel->addField($fieldObj->getName(), $clone);
+                $ancestorModel->addField($fieldObj->getName(), $clone);
             }
         }
-        $this->fillAncestry($parentModel);
+        if ($modelType->getName() !== $ancestorModel->getName()) {
+            $this->fillAncestry($ancestorModel);
+        }
```
ChloeSartorelli:
```suggestion
        $this->fillAncestry($ancestorModel);
```
The changes above mean we won't ever get in this scenario by the time we get down here, which makes the flow a little cleaner.

GuySartorelli (changes requested):

src/Schema/DataObject/InheritanceBuilder.php
emteknetnz: Ah, I think while this seems seems like simple inverted logic to make it read better, I think this may introduce an error in some scenarios.  I'm not entirely sure so I'm inclined to not change it right now.

The 
```
if (!$ancestorModel) {
    return;
}
```
Should probably actually throw an exception as previously it would fail when it would got to `$ancestorModel->getFieldByName`.  However you could also been in a weird situation `$ancestorModel` is null, though `$modelTypes->getFields()` is an empty array so you never actually get to that point.

I don't know if any of this stuff would happen, I'm just speculating for stuff that might happen

I think I'd prefer to just leave as and not risk causing additional regressions here.

src/Schema/DataObject/InheritanceBuilder.php
emteknetnz: Not changing logic above, so not changing here either

src/Schema/DataObject/InheritanceBuilder.php
GuySartorelli: Lets talk about this offline - there's a lot going on here and there are definitely some changes required (the `if ($modelType->getName() !== $ancestorModel->getName()) {` condition at the bottom could result in some ancestry further up the chain not being processed, or at least _looks_ like it could which is arguably just as bad) - but we should work out whether the problem you've perceived with my recommendation is real or otherwise avoidable.

ChloeSartorelli: The `findOrMakeModel()` method _cannot_ return null. It's got a concrete return type of `ModelType`. So before if we'd had a `if (!$parentModel)` condition, it simply would never have been true.

The scenario you're describing was impossible before, and will only be possible now because in my suggestion we _explicitly_ set the variable to null in scenarios where the model we have for that variable is not value for the logic that follows.

GuySartorelli (changes requested):

src/Schema/DataObject/InheritanceBuilder.php
emteknetnz: > The scenario you're describing was impossible before, and will only be possible now because in my suggestion we explicitly set the variable to null in scenarios where the model we have for that variable is not value for the logic that follows.

I don't understand this sentence

Feels like we're dragging this out for the sake of making some code that works read a little nicer? This PR has a test showing things work as they should, can't we just merge this as is?

ChloeSartorelli: That sentence basically says: You described a scenario that you think should throw an exception. The scenario where your `if (!$ancestorModel)` condition resolves to `true` wasn't possible before this PR, nor in this PR as it currently stands. The only reason it will potentially resolve to `true` _with_ my suggested change is because we _explicitly_ set it to null. The actual returned value from `findOrMakeModel()` will _never_ be null, so we don't need to worry about that. There was never a scenario where we would have gotten to `$ancestorModel->getFieldByName()` where `$ancestorModel` was null in your current PR, nor prior to this PR if we exchange `$ancestorModel` for `$parentModel`.

It is a response to "I don't know if any of this stuff would happen, I'm just speculating for stuff that might happen" - basically, none of that stuff _would_ nor _could_ happen.

> Feels like we're dragging this out for the sake of making some code that works read a little nicer?

Nope. I'm recommending this change in part because it increases code readability, in part because it reduces _confusion_, and in part because it reduces _complexity_. It is an all-round good suggestion that make the code easier to maintain and I can't see any reason to _not_ apply the change - I've responded to your one concern by pointing out that it is not a problem. Please apply the change.

emteknetnz: done

src/Schema/DataObject/InheritanceBuilder.php
emteknetnz: done

src/Schema/DataObject/InheritanceBuilder.php
emteknetnz: done

GuySartorelli (approved):
Looks good to me

PR #551, opened 2023-08-24
MNT Fix unit test

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/101

This particular unit test failure was only seen on silverstripe/installer recipe-kitchen-sink as it requires the silverstripe/admin test UsedOnTableTest::getUsage() to run first and cache some config before running this test


GuySartorelli (approved):
I was going to suggest using `Injector::inst()->registerService()` instead but given the end result would be the same and there's a big comment explaining what's happening I'm fine with this approach.

Issue #561, opened 2023-11-09
MariaDB - failing integration test


emteknetnz:
Unit test failure only with MariaDB, not MySQL

https://github.com/silverstripe/silverstripe-installer/actions/runs/6787809028/job/18507503396

```
1) SilverStripe\GraphQL\Tests\Schema\IntegrationTest::testFilterAndSort
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'test2'
+'test3'

/home/runner/work/silverstripe-installer/silverstripe-installer/vendor/silverstripe/graphql/tests/Schema/IntegrationTest.php:1579
/home/runner/work/silverstripe-installer/silverstripe-installer/vendor/silverstripe/graphql/tests/Schema/IntegrationTest.php:494
/home/runner/work/silverstripe-installer/silverstripe-installer/vendor/bin/phpunit:122
```

## PR
- https://github.com/silverstripe/silverstripe-graphql/pull/562
- https://github.com/silverstripe/silverstripe-graphql/pull/563

GuySartorelli:
PR merged.

PR #562, opened 2023-11-14
FIX QuerySort::sort method should support sorting for multi arguments

sabina-talipova:
## Description
Even if a method receives several parameters, then sorting occurs by the field of the arguments that is last specified in the schema, and not in the passed parameters. That is, if `SilverStripe\GraphQL\Tests\Fake\DataObjectFake` has the following schema
```yml
fields:
    myField: true
    AuthorID: true
```
and sorting is required `AuthorID => DESC, myField => ASC`, then sorting was done only by `AuthorID`, since in the schema this field is indicated after myField.
Therefore, the call `$list->sort()` must be completed after all the necessary processing has been carried out.

## Parent issue
- https://github.com/silverstripe/silverstripe-graphql/issues/561

### DIFF
--- a/src/Schema/DataObject/Plugin/QuerySort.php
+++ b/src/Schema/DataObject/Plugin/QuerySort.php
@@ -102,6 +102,9 @@ class QuerySort extends AbstractQuerySortPlugin
             }
             $filterArgs = $args[$fieldName] ?? [];
             $paths = NestedInputBuilder::buildPathsFromArgs($filterArgs);
+            if (empty($paths)) {
+                return $list;
+            }
             $schemaContext = SchemaConfigProvider::get($context);
             if (!$schemaContext) {
                 throw new Exception(sprintf(
@@ -111,6 +114,7 @@ class QuerySort extends AbstractQuerySortPlugin
                 ));
             }
 
+            $normalisedPaths = [];
             foreach ($paths as $path => $value) {
                 $normalised = $schemaContext->mapPath($rootType, $path);
                 Schema::invariant(
@@ -120,10 +124,11 @@ class QuerySort extends AbstractQuerySortPlugin
                     $path,
                     $rootType
                 );
-                $list = $list->sort($normalised, $value);
+
+                $normalisedPaths[$normalised] = $value;
             }
 
-            return $list;
+            return $list->sort($normalisedPaths);
         };
     }
 
--- a/src/Schema/DataObject/Plugin/QuerySort.php
+++ b/src/Schema/DataObject/Plugin/QuerySort.php
@@ -18,6 +18,7 @@ use SilverStripe\ORM\DataObject;
 use Closure;
 use SilverStripe\ORM\Sortable;
 use Exception;
+use GraphQL\Type\Definition\ResolveInfo;
 
 /**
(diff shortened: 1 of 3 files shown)


GuySartorelli (changes requested):
This can't be `MNT` since it's changing actual code that projects will be using.

If the changes in the `QuerySort` class are fixing a bug, you should use `FIX` and the commit message should explain what the bug is that's being fixed.
If they're just an enhancement, then use `ENH` and the commit message should explain what the enhancement is.

tests/Schema/_testFilterAndSort/models.yml
```diff
@@ -5,8 +5,8 @@ SilverStripe\GraphQL\Tests\Fake\DataObjectFake:
         filter: true
         sort: true
   fields:
-    myField: true
     AuthorID: true
+    myField: true
```
ChloeSartorelli: Does this change actually affect the test at all?

ChloeSartorelli: Talked offline - this does apparently affect the tests. I'll see if I can find out what's going on there because this shouldn't affect the sort order, especially given we're explicitly sorting in the query.

GuySartorelli:
The reason changing the order of the fields in the schema affected the results is because the underlying GraphQL library we're using ignores the order that fields are entered in the query arguments. It uses the order that the fields were defined in the schema.

I've added a new commit on top of yours to resolve this. It also changes the test to use a query provider so it will run all of the FilterAndSort scenarios even if one fails. I removed the scenarios that had commented out tests because they weren't actually contributing anything to the test coverage.

Normally I would have asked you to make those changes, but I didn't know what changes were necessary or even possible - and to find that out I had to essentially do the full implementation. At that point it was just easier to commit that implementation rather than ask you to redo what I had just done.

## Next steps

1. Have a look over this code and test it locally to make sure you're happy with it. This is still your PR, so make any changes you think are appropriate.
1. Add another test scenario where we sort first by `myField` and then by `AuthorID` to protect us against regressions
2. Add new test coverage to check this works correctly with a normal `read` (not read one - but read many) query
3. See if changes similar to what I've made are necessary in `src/Schema/Plugin/SortPlugin.php`

src/Schema/DataObject/Plugin/QuerySort.php
```diff
@@ -96,12 +97,21 @@ public static function sort(array $context): closure
     {
         $fieldName = $context['fieldName'];
         $rootType = $context['rootType'];
-        return function (?Sortable $list, array $args, array $context) use ($fieldName, $rootType) {
+        return function (?Sortable $list, array $args, array $context, ResolveInfo $info) use ($fieldName, $rootType) {
```
ChloeSartorelli, line 100: Since this is an anonymous function, it's not part of our public API, so we can add arguments without breaking backwards compatibility.

GuySartorelli:
Please also check if this bug affects graphql 4 - if so, we need to retarget to `4.3`

sabina-talipova:
Open new PR for 4.3:
- https://github.com/silverstripe/silverstripe-graphql/pull/563

PR #563, opened 2023-11-22
FIX QuerySort::sort method should support sorting for multi arguments

sabina-talipova:
## Description
Even if a method receives several parameters, then sorting occurs by the field of the arguments that is last specified in the schema, and not in the passed parameters. That is, if `SilverStripe\GraphQL\Tests\Fake\DataObjectFake` has the following schema
```yml
fields:
    myField: true
    AuthorID: true
```
and sorting is required `AuthorID => DESC, myField => ASC`, then sorting was done only by `AuthorID`, since in the schema this field is indicated after myField.
Therefore, the call `$list->sort()` must be completed after all the necessary processing has been carried out.

## Parent issue
- https://github.com/silverstripe/silverstripe-graphql/issues/561

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "php": "^7.4 || ^8.0",
         "silverstripe/framework": "^4.11",
         "silverstripe/vendor-plugin": "^1.0",
-        "webonyx/graphql-php": "^14.0",
+        "webonyx/graphql-php": "^14.11",
         "silverstripe/event-dispatcher": "^0.1.2",
         "guzzlehttp/guzzle": "^7.3",
         "guzzlehttp/psr7": "^2",
--- a/src/Schema/DataObject/Plugin/QuerySort.php
+++ b/src/Schema/DataObject/Plugin/QuerySort.php
@@ -18,6 +18,7 @@ use SilverStripe\ORM\DataObject;
 use Closure;
 use SilverStripe\ORM\Sortable;
 use Exception;
+use GraphQL\Type\Definition\ResolveInfo;
 
 /**
  * Adds a sort parameter to a DataObject query
@@ -87,7 +88,6 @@ class QuerySort extends AbstractQuerySortPlugin
         return $filters;
     }
 
-
     /**
      * @param array $context
      * @return Closure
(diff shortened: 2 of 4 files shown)


tests/Schema/IntegrationTest.php
```diff
@@ -631,6 +668,62 @@ public function testFieldAliases()
         $this->assertResult('readOneDataObjectFake.author', null, $result);
     }
 
+    public function provideFilterAndSortOnlyRead(): array
```
sabina-talipova, line 671: I'll add one more test case for SortPlugin here.

composer.json
```diff
@@ -7,7 +7,7 @@
         "php": "^7.4 || ^8.0",
         "silverstripe/framework": "^4.11",
         "silverstripe/vendor-plugin": "^1.0",
-        "webonyx/graphql-php": "^14.0",
+        "webonyx/graphql-php": "^14.11",
```
sabina-talipova, line 10: We need this to pass `--prefer-lowest` CI. Since in 14.0.0 `FieldDefinition` doesn't have `getName()` method

src/Schema/DataObject/Plugin/QuerySort.php
```diff
@@ -88,6 +89,10 @@ protected function buildAllFieldsConfig(ModelType $modelType, Schema $schema, in
     }
 
 
+    /**
+     * @param array $context
+     * @return Closure
+     */
```
ChloeSartorelli:
```suggestion
```
It already has this below

src/Schema/DataObject/Plugin/QuerySort.php
```diff
@@ -120,13 +135,68 @@ public static function sort(array $context): closure
...
+     * Gets the original order of fields to be sorted based on the query args order.
+     *
+     * This is necessary because the underlying GraphQL implementation we're using ignores the
+     * order of query args, and uses the order that fields are defined in the schema instead.
+     */
+    private static function getSortOrder(ResolveInfo $info, string $fieldName)
+    {
+        if (!method_exists($info->fieldDefinition, 'getName')) {
+            return [];
+        }
```
ChloeSartorelli: Is this still needed with the composer dependency change?

GuySartorelli (changes requested):
Mostly looks good, just a couple small things to tidy up

GuySartorelli (approved):
Looks good, tests are green here and were green when running against CMS 5 CI as well.

PR #564, opened 2023-11-26
FIX Add test cases for SortPlugin

sabina-talipova:
## Parent issue
- https://github.com/silverstripe/silverstripe-graphql/issues/565

### DIFF
--- a/src/Schema/Plugin/SortPlugin.php
+++ b/src/Schema/Plugin/SortPlugin.php
@@ -14,14 +14,17 @@ use SilverStripe\GraphQL\Schema\Interfaces\SchemaUpdater;
 use SilverStripe\GraphQL\Schema\Resolver\ResolverReference;
 use SilverStripe\GraphQL\Schema\Schema;
 use SilverStripe\GraphQL\Schema\Services\NestedInputBuilder;
+use SilverStripe\GraphQL\Schema\Traits\SortTrait;
 use SilverStripe\GraphQL\Schema\Type\InputType;
 use SilverStripe\ORM\Sortable;
 use Closure;
+use GraphQL\Type\Definition\ResolveInfo;
 
 class SortPlugin implements FieldPlugin, SchemaUpdater
 {
     use Configurable;
     use Injectable;
+    use SortTrait;
 
     const IDENTIFIER = 'sorter';
 
@@ -86,11 +89,16 @@ class SortPlugin implements FieldPlugin, SchemaUpdater
     public static function sort(array $context): Closure
     {
         $fieldName = $context['fieldName'];
-        return function (?Sortable $list, array $args) use ($fieldName) {
+        return function (?Sortable $list, array $args, array $context, ResolveInfo $info) use ($fieldName) {
             if ($list === null) {
                 return null;
             }
-            $sortArgs = $args[$fieldName] ?? [];
+
+            if (!isset($args[$fieldName])) {
+                return $list;
+            }
+
+            $sortArgs = self::getSortArgs($info, $args, $fieldName);
             $list = $list->sort($sortArgs);
 
             return $list;
--- a/src/Schema/Traits/SortTrait.php
+++ b/src/Schema/Traits/SortTrait.php
@@ -1,0 +1,58 @@
+<?php
+
+namespace SilverStripe\GraphQL\Schema\Traits;
+
+use GraphQL\Type\Definition\ResolveInfo;
+
+trait SortTrait
+{
+    private static function getSortArgs(ResolveInfo $info, array $args, string $fieldName): array
+    {
+        $sortArgs = [];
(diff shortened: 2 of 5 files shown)


GuySartorelli (changes requested):
You have only one test validating that this sort plugin works when sorting multiple fields at once:
`files(sort: { name: ASC, title: DESC }) {`

Please also add one with those fields reversed, to validate that the order in which the fields are added _in the query_ changes the actual sort order.
`files(sort: { title: DESC, name: ASC }) {`
Make sure this new scenario gives different results than the existing one.

src/Schema/Plugin/SortPlugin.php
```diff
@@ -86,17 +87,67 @@ public function apply(Field $field, Schema $schema, array $config = []): void
     public static function sort(array $context): Closure
     {
         $fieldName = $context['fieldName'];
-        return function (?Sortable $list, array $args) use ($fieldName) {
+        return function (?Sortable $list, array $args, array $context, ResolveInfo $info) use ($fieldName) {
             if ($list === null) {
                 return null;
             }
-            $sortArgs = $args[$fieldName] ?? [];
+            $sortArgs = static::getSortArgs($info, $args, $fieldName);
```
ChloeSartorelli:
```suggestion
            $sortArgs = self::getSortArgs($info, $args, $fieldName);
```
Since it's a private method, `static` is effectively the same as `self` in this scenario - this is a bit of a nitpick but it's nicer to explicitly call private statics on `self` to avoid the appearance that there's some magic allowing a different method to be called on the subclass.

src/Schema/Plugin/SortPlugin.php
```diff
@@ -86,17 +87,67 @@ public function apply(Field $field, Schema $schema, array $config = []): void
     public static function sort(array $context): Closure
     {
         $fieldName = $context['fieldName'];
-        return function (?Sortable $list, array $args) use ($fieldName) {
+        return function (?Sortable $list, array $args, array $context, ResolveInfo $info) use ($fieldName) {
             if ($list === null) {
                 return null;
             }
```
ChloeSartorelli, line 95: Return early if there's no sort argument for this query.
```suggestion
            }

            if (!isset($args[$fieldName])) {
                return $list;
            }
```

src/Schema/Plugin/SortPlugin.php
```diff
@@ -86,17 +87,67 @@ public function apply(Field $field, Schema $schema, array $config = []): void
...
+                $sortOrder = [];
+                foreach ($arg->value->fields as $field) {
+                    $sortOrder[] = $field->name->value;
+                }
+                return $sortOrder;
+            }
+        }
+
+        return [];
+    }
```
ChloeSartorelli: Since this is identical to the code in `src/Schema/DataObject/Plugin/QuerySort.php`, move the code into a new trait and use the trait in both classes. This way if we find some fault in the logic later on we can fix it in both places at the same time.

GuySartorelli (changes requested):
Since this is now a bugfix, please update the commit message to use the `FIX` prefix and to have a message that explains what bug is being fixed.

src/Schema/Plugin/SortPlugin.php
sabina-talipova: Updated

src/Schema/Plugin/SortPlugin.php
sabina-talipova: Updated

src/Schema/Plugin/SortPlugin.php
sabina-talipova: Updated

src/Schema/Traits/SortTrait.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+namespace SilverStripe\GraphQL\Schema\Traits;
```
ChloeSartorelli, line 3:
```suggestion
namespace SilverStripe\GraphQL\Schema\Plugin;
```
Please move this into the `Plugin/` directory - we don't need a whole directory for traits.

GuySartorelli (changes requested):
Just one small change - everything else looks good.

src/Schema/Traits/SortTrait.php
sabina-talipova: Since both sort plugins are located in different directories `SilverStripe\GraphQL\Schema\DataObject\Plugin\QuerySort` and `SilverStripe\GraphQL\Schema\Plugin\SortPlugin`, probably it would be better to leave trait in common directory for now, while we have not decided about future architecture of our plugins.

ChloeSartorelli: Ahh yup good call. Didn't notice there were two plugin directories.

GuySartorelli (approved):
LGTM

Issue #573, opened 2024-02-20
Query sorting approach may be incorrect


lozcalver:
### Module version(s) affected

v4, v5

### Description

Currently sort arguments are passed as an object:

```
{
    "created": "asc",
    "name": "asc"
}
```

According to the [GraphQL spec](https://spec.graphql.org/October2021/#sec-Input-Object-Values.Input-object-fields-are-unordered), there’s no guarantee about the order that these will be interpreted, so the sort order might be interpreted as either `Created ASC, Name ASC` or `Name ASC, Created ASC`. It appears that `webonyx/graphql-php` will interpret them in the order they’re defined in the _schema_, not in the query or inputs.

We’ve managed to work around this by manually parsing the query in https://github.com/silverstripe/silverstripe-graphql/pull/563, but that only works if the sort order is baked into the query itself. When the sort order is passed as an argument to the query, the original order is lost as soon as the query is parsed (more info: https://github.com/silverstripe/silverstripe-graphql/issues/571#issuecomment-1933825823). That aside, manually inspecting the query AST to try to “fix” the sort order is probably an indication we’re approaching this the wrong way.

One alternative approach suggested is to use an array instead (a “list” in GraphQL speak). That might look something like:

```
[
    {"created": "asc"},
    {"name": "asc"}
]
```

_If_ we could use that approach, I think it’d be important to preserve backward compatibility if at all possible. I haven’t looked at how that might be possible yet - perhaps interfaces or union types could help.

### How to reproduce

-

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

lozcalver:
I’ve been trying to think how we might be able to set up types for this while preserving some kind of backward compatibility with the existing sort fields. The only way I can come up with is this:

```graphql
type Query {
  readMembers(
    filter: MemberFilterFields
    sort: MemberSortOptions
  ): [Member!]!
}

# New union type to support both old input (for bc) and new input
union MemberSortOptions = MemberSortFields | MemberSortExpressions

# Already exists
enum SortDirection {
  ASC
  DESC
}

# Already exists
input MemberSortFields {
  created: SortDirection
  name: SortDirection
}

# New enum of available fields to sort by
enum MemberSortField {
  name
  created
}

# New type
[...]

PR #577, opened 2024-02-28
FIX: Don't attempt to correct sort when passed as an argument (closes #571)

### LINKED ISSUE #571: GraphQL errors when passing sort parameters as separate variables
### Module version(s) affected

4.3.5+

### Description

I’ve been seeing the following warning from some projects with GraphQL, have finally decided to stop ignoring them and investigate:

> E_WARNING: Undefined property: GraphQL\Language\AST\VariableNode::$fields

from this line specifically:

https://github.com/silverstripe/silverstripe-graphql/blob/4.3.7/src/Schema/Traits/SortTrait.php#L49

I assume the code isn’t expecting a `VariableNode` instance

Appears to be a regression from https://github.com/silverstripe/silverstripe-graphql/pull/563 (`4.3.4` doesn’t produce the warning, `4.3.5` and up does).

I don’t fully understand the code here, so I’m not sure why this is occurring yet. I’ll try to do some more testing tomorrow.

I’ve only confirmed this against SS4 (GraphQL v4) but I suspect the same issue would apply to SS5.

### How to reproduce

The best I can offer at this stage is an example of the schema:

```yml
SilverStripe\Blog\Model\BlogPost:
  fields:
    id: true
    title: true
    urlSegment: true
    link: true
    editorsPick: true
    timeToRead: true
    publishDate: true
    timesViewed: true
    parentId: true
    epochPublishDate:
      type: 'int'
    cardImageSrc:
      type: 'string'
    minutesToRead:
      type: 'string'
    estimatedTimeToConsume:
      type: 'string'
    renderCard:
      type: 'string'
  operations:
    read: true
```

And an example query:

```json
{
"operationName": "ReadBlogPosts", 
"query": 
"query ReadBlogPosts($filter: BlogPostFilterFields, $sort: BlogPostSortFields, $offset: Int, $limit: Int) {
  readBlogPosts(filter: $filter, sort: $sort, offset: $offset, limit: $limit) {
    edges: nodes {
      id
      title
      editorsPick
      publishDate
      cardImageSrc
      estimatedTimeToConsume
      link
      epochPublishDate
      renderCard
      __typename
    }
    pageInfo {
      totalCount
      hasNextPage
      hasPreviousPage
      __typename
    }
    __typename
  }
}"
, 
"variables": {
"filter": {}, 
"limit": 6, 
"offset": 0, 
"sort": []
}
}
```

### Possible Solution

_No response_

### Additional Context

_No response_

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-graphql/pull/577 (avoids the warning but doesn't fix the underlying problem)

lozcalver:
## Description
Proposed fix for #571. Doesn’t actually solve the underlying issue, but prevents the warnings being emitted by `SortTrait`.

## Manual testing steps
N/A

## Issues
- #571

## Pull request checklist

- [ ] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [x] CI is green

### DIFF
--- a/src/Schema/Traits/SortTrait.php
+++ b/src/Schema/Traits/SortTrait.php
@@ -2,7 +2,10 @@
 
 namespace SilverStripe\GraphQL\Schema\Traits;
 
+use GraphQL\Language\AST\ObjectValueNode;
 use GraphQL\Type\Definition\ResolveInfo;
+use Psr\Log\LoggerInterface;
+use SilverStripe\Core\Injector\Injector;
 
 trait SortTrait
 {
@@ -41,6 +44,14 @@ trait SortTrait
             // Find the sort arg
             foreach ($node->arguments as $arg) {
                 if ($arg->name->value !== $fieldName) {
+                    continue;
+                }
+
+                // If the sort has been passed as a variable, we can't attempt to fix it
+                // See https://github.com/silverstripe/silverstripe-graphql/issues/573
+                if (!$arg->value instanceof ObjectValueNode) {
+                    $logger = Injector::inst()->get(LoggerInterface::class . '.graphql-quiet');
+                    $logger->warning('Unable to adjust sort order, sort fields may be applied in incorrect order.');
                     continue;
                 }
 
--- a/tests/Schema/IntegrationTest.php
+++ b/tests/Schema/IntegrationTest.php
@@ -847,6 +847,112 @@ GRAPHQL;
         $this->assertResults($expected, $records);
     }
 
+    public function provideFilterAndSortWithArgsOnlyRead(): array
+    {
+        return [
+            'read with sort - with sort arg' => [
+                'fixture' => '_QuerySort',
+                'query' => <<<GRAPHQL
+query (\$sort: DataObjectFakeSortFields) {
+  readDataObjectFakes(sort: \$sort) {
+    nodes {
+      myField
+      author {
+        firstName
+      }
+    }
+  }
+}
+GRAPHQL,
+                'args' => [
+                    'sort' => ['author' => ['id' => 'DESC'], 'myField' => 'ASC']
+                ],
+                'expected' => [
+                    ["myField" => "test2", "author" => ["firstName" => "tester2"]],
+                    ["myField" => "test3", "author" => ["firstName" => "tester2"]],
+                    ["myField" => "test1", "author" => ["firstName" => "tester1"]],
+                ],
+            ],
+            'read with sorter files ParentID DESC, name ASC - with sort args' => [
+                'fixture' => '_SortPlugin',
+                'query' => <<<GRAPHQL
+query (\$sort: DataObjectFakeSortFields, \$fileSort: FilesSimpleSortFields) {
+  readDataObjectFakes(sort: \$sort) {
+    nodes {
+      myField
+      files(sort: \$fileSort) {
+        title
+      }
+    }
+  }
+}
+GRAPHQL,
+                'args' => [
+                    'sort' => ['myField' => 'ASC'],
+                    'fileSort' => ['ParentID' => 'DESC', 'name' => 'ASC'],
+                ],
+                'expected' => [
+                    ["myField" => "test1", "files" => [["title" => "file2"], ["title" => "file4"], ["title" => "file1"], ["title" => "file3"]]],
+                    ["myField" => "test2", "files" => []],
+                    ["myField" => "test3", "files" => []],
+                ],
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider provideFilterAndSortWithArgsOnlyRead
+     */
+    public function testFilterAndSortWithArgsOnlyRead(string $fixture, string $query, array $args, array $expected)
+    {
+        $author = Member::create(['FirstName' => 'tester1']);
+        $author->write();
+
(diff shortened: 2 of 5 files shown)


tests/Schema/IntegrationTest.php
```diff
@@ -847,6 +827,119 @@ public function testFilterAndSortOnlyRead(string $fixture, string $query, array
...
+
+        $result = $this->querySchema($schema, $query, $args);
+        $this->assertSuccess($result);
+        $records = $result['data']['readDataObjectFakes']['nodes'] ?? [];
+
+        // We intentionally don't check the sort order, as it's expected it may not match:
+        // See https://github.com/silverstripe/silverstripe-graphql/issues/573
+        usort($expected, function($a, $b) {
+            return strcmp($a['myField'], $b['myField']);
+        });
```
lozcalver: This test has been added to ensure that sort args can be used without throwing warnings, so I’m just forcing the same sort order here

lozcalver: Turns out there’s an assertion for this, `assertEqualsCanonicalizing()`, so I’ve switched to that instead

marwan38:
Bump. This issue is completely breaking one of my queries

tests/Schema/IntegrationTest.php
```diff
@@ -674,126 +674,106 @@ public function testFieldAliases()
...
+  readDataObjectFakes(sort: { myField: ASC }) {
+    nodes {
+      myField
+      files(sort: { title: DESC }) {
+        title
+      }
+    }
+  }
+}
+GRAPHQL,
```
ChloeSartorelli: Can you please revert the indentation to be what it was before?
For this and all other HEREDOC you've touched in this PR.

tests/Schema/IntegrationTest.php
ChloeSartorelli, line 1: I haven't done a thorough review of this test - but at a glance it looks like there's one less test scenario than there was before?

src/Schema/Traits/SortTrait.php
```diff
@@ -44,6 +45,12 @@ private static function getSortOrder(ResolveInfo $info, string $fieldName)
                     continue;
                 }
 
+                // If the sort has been passed as a variable, we can't attempt to fix it
+                // See https://github.com/silverstripe/silverstripe-graphql/issues/573
+                if (!$arg->value instanceof ObjectValueNode) {
+                    continue;
+                }
```
ChloeSartorelli: What are the circumstances that result in the value not being an instance of that class?
Depending on what those circumstances are, we might want to throw an exception (or at the very least log a warning) if someone tries to perform a sort that we explicitly don't respect here.

GuySartorelli (changes requested):

src/Schema/Traits/SortTrait.php
lozcalver: `$arg->value` is only ever an instance of `ObjectValueNode` when the sort order is baked into the query, so the `continue` statement is reached when sort is passed as a variable instead.

lozcalver:
@GuySartorelli a little more context: there are two ways to pass a sort order: (1) baked directly in to the query itself, or (2) as an argument.

For (1), #563 was merged to ensure that the order that you provide sort fields is respected. A regression from that PR is that when you use approach (2), a warning is thrown.

For (2), #563 had no effect anyway. This PR just ensures that when using this approach, the warning isn’t thrown and everything behaves as it did before #563 was merged.

Ideally, we’d expand #563 to cover both approaches to sorting. However that isn’t possible, because the underlying GraphQL library discards the order that sort fields are passed in when parsing the query. That isn’t a bug with the library, it’s actually the intended behaviour as it respects the GraphQL spec. There‘s some more info in #573 about that, but that’s a much bigger job and would have to be targeted at a major release.

GuySartorelli:
Thank you for that extra context. I think in that case I'm okay with this - but a warning needs to be logged so developers can get some feedback that their query isn't working as intended.

lozcalver:
> a warning needs to be logged so developers can get some feedback that their query isn't working as intended

I’m not really sure how to tackle that, the warning is the problem... we could possibly use a custom logger and call `$logger->info()` but unless the logger is set up to actually record that somewhere, nobody will ever see it.

GuySartorelli:
We should log a warning using the `Psr\Log\LoggerInterface.errorhandler` singleton documented in https://docs.silverstripe.org/en/5/developer_guides/debugging/error_handling/#logging-and-error-handling

> unless the logger is set up to actually record that somewhere, nobody will ever see it.

Yup, that's true of all logging. Better to give the option than nothing though.

lozcalver:
I’ve finally got around to this. Using `LoggerInterface::class . '.errorhandler'` results in the warning being output in the HTTP response, which breaks everything, so I’ve copied the approach from the old file migration task: https://github.com/silverstripe/silverstripe-assets/blob/1.13/_config/migration.yml. Essentially it’s a logger instance that doesn’t include an output handler so nothing appears in the HTTP output, but the warning is still logged to the error log

GuySartorelli (approved):
LGTM, thanks for doing this.

Issue #578, opened 2024-03-21
GraphQL Directives


marwan38:
### Description

Directives are a fundamental part of GraphQL and as such I believe should be part of the base providing of any GraphQL implementation. It should be fairly straight forward to add as it's not much different from adding a new type, or interface, etc. Happy to do it with a bit of guidance.

### Additional context or points of discussion

I would not ask or bother with this feature were I able to access the GraphQL schema directly during schema build, but I haven't been able to achieve that as the schema instance is generated right before generating the schema (as far as I understood the source code). If we can expose the schema, I'd prefer doing that as it's much less work Adding a directive is as easy as:

```php
<?php
use GraphQL\Type\Definition\Directive;

$directive = new Directive($config);
$graphqlSchema->directives[] = $directive;
```

That is all.

### Validations

- [X] You intend to implement the feature yourself
- [X] You have read the [contributing guide](https://docs.silverstripe.org/en/contributing/code/)
- [X] You strongly believe this feature should be in core, rather than being its own community module
- [X] You have checked for existing issues or pull requests related to this feature (and didn't find any)

## PRs
- https://github.com/silverstripe/silverstripe-graphql/pull/580

maxime-rainville:
Honestly, it's unlikely we'll be able to prioritise this feature. Would you be keen to implement it yourself?

marwan38:
> Honestly, it's unlikely we'll be able to prioritise this feature. Would you be keen to implement it yourself?

That's my intention. I just hope I can make the time to do it. 

It's taking me quite a long time, migrating our codebase across to graphql 4+. It's been over two months so at this point I'm trying to workaround the little things that have a decent workaround. 

I do think directives are quite valuable (just the basic @deprecatedReason is useful). Client side directives are also used (Apollo graphql) to handle certain use cases. They are also a part of the graphql specification.

No promises, but.. we'll see after I finish my current task.

PR #580, opened 2024-04-13
API Add Directive types

marwan38:
This PR is working as intended for my use case and can be merged in. However, I have a note for the maintainers here to make the decision on. I'm currently using this as a patch in our codebase on version 4.13 of this module. I'm hoping that it can be merged in by the time I'm done with our codebase migration and can then bump up to v5 of this module.

SchemaTransciber [does not introspect the directives](https://github.com/silverstripe/silverstripe-graphql/blob/5/src/Schema/Services/SchemaTranscriber.php#L97). Therefore users relying on the schema transcriber won't be able to get the full schema including the directives. I can't make the decision on behalf of the framework on how much you want added within the introspection query.  I personally don't use it, we generate the schema directly using @graphql-codegen and that introspects what we need. An example of a more comprehensive introspection query can be found [here](https://github.com/graphql/graphql-js/blob/main/src/utilities/getIntrospectionQuery.ts#L41) and what the directives field may contain [here](https://github.com/graphql/graphql-js/blob/main/src/utilities/getIntrospectionQuery.ts#L76). That introspect query also adds the complexity of configuration, so I'm not really sure what you guys would want here. As I said, this feature as-is is enough for me at the moment.

## Description

Directives are part of the GraphQL specification and this PR adds the Directive type for developers to create custom ones. Ideally, this PR would be followed up (or done alongside) with more integration of directives. One use case i would like to see is adding `deprecationReason` to to types.
```yaml
types:
  myType:
    fields: {}
    deprecationReason: Use goodType instead
```
or if there is any reason to add directives to a field (not sure if this is a valid use case)
```yaml
types:
  myType:
    fields: {}
    directives:
      deprecated:
        deprecationReason: Use goodType instead
```

## Manual testing steps

A new type is added `directives` in the graphql schema yaml file.
```yaml
---
directives:
  myDirective:
    description: 'Optional description'
    args:
      if:
        type: Boolean
        defaultValue: true
    locations: [QUERY]
```

Viewing the the graphql schema (e.g. using /dev/graphql/ide) should now contain an entry like so:
```
# Optional description
directive @myDirective(if: Boolean = true) on QUERY
```

The configuration of the `Directive` type can be found here https://webonyx.github.io/graphql-php/type-definitions/directives/#custom-directives.

The directive can then be resolved within the resolver like so:
```php
/** @var ResolveInfo $info */
$allDirectives = $info->operation?->directives?->getIterator() ?? [];
foreach ($allDirectives as $directive) {
  var_dump($directive->name->value);
}
```

## Issues

- #578

## Pull request checklist

- [ ] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [ ] CI is green

### DIFF
--- a/src/Schema/Storage/CodeGenerationStore.php
+++ b/src/Schema/Storage/CodeGenerationStore.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\GraphQL\Schema\Storage;
 
 use Exception;
+use GraphQL\GraphQL;
 use GraphQL\Type\Schema as GraphQLSchema;
 use GraphQL\Type\SchemaConfig as GraphQLSchemaConfig;
 use Psr\Log\LoggerInterface;
@@ -29,6 +30,7 @@ class CodeGenerationStore implements SchemaStorageInterface
     use Configurable;
 
     const TYPE_CLASS_NAME = 'Types';
+    const DIRECTIVE_CLASS_NAME = 'Directives';
 
     /**
      * @config
@@ -137,6 +139,8 @@ class CodeGenerationStore implements SchemaStorageInterface
             'typeClassName' => self::TYPE_CLASS_NAME,
             'namespace' => $this->getNamespace(),
             'obfuscator' => $obfuscator,
+            'directiveClassName' => self::DIRECTIVE_CLASS_NAME,
+            'directives' => $schema->getDirectives()
         ];
 
         $config = $schema->getConfig()->toArray();
@@ -146,10 +150,10 @@ class CodeGenerationStore implements SchemaStorageInterface
             $fs->dumpFile(
                 $configFile,
                 '<?php ' .
-                PHP_EOL .
-                'return ' .
-                var_export($config, true) .
-                ';'
+                    PHP_EOL .
+                    'return ' .
+                    var_export($config, true) .
+                    ';'
             );
         } catch (IOException $e) {
             throw new RuntimeException(sprintf(
@@ -164,7 +168,8 @@ class CodeGenerationStore implements SchemaStorageInterface
             $schema->getEnums(),
             $schema->getInterfaces(),
             $schema->getUnions(),
-            $schema->getScalars()
+            $schema->getScalars(),
+            $schema->getDirectives(),
         );
         $encoder = Encoder::create(Path::join($templateDir, 'registry.inc.php'), $allComponents, $globals);
         $code = $encoder->encode();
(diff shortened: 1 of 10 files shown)


src/Schema/Storage/templates/registry.inc.php
```diff
@@ -26,3 +26,27 @@ public static function <?=$component->getName(); ?>() { return static::get('<?=$
     <?php endforeach; ?>
 
 }
+
```
marwan38, line 29: This probably breaks some PSR specs, I'm not sure if it matters here?

src/Schema/Storage/CodeGenerationStore.php
```diff
@@ -146,10 +150,10 @@ public function persistSchema(StorableSchema $schema): void
...
                 $configFile,
                 '<?php ' .
-                PHP_EOL .
-                'return ' .
-                var_export($config, true) .
-                ';'
+                    PHP_EOL .
+                    'return ' .
+                    var_export($config, true) .
+                    ';'
```
ChloeSartorelli, line 156: Why was this change made? It seems unrelated

src/Schema/Storage/templates/registry.inc.php
ChloeSartorelli: You say "this" on a comment on an empty line... but I assume you are actually referring to having multiple classes defined in the same file?

That does break PSR1 (see [the PSR1 spec](https://www.php-fig.org/psr/psr-1/#3-namespace-and-class-names)).
Ideally each of these templates should be explicitly for a single class.

One approach would be to reuse the same template file (with only one clsas definition) for both of these since they're practically identical, and just add appropriate conditional statements. e.g. type.inc.php uses a lot of conditional logic to build its class.

tests/Schema/IntegrationTest.php
```diff
@@ -1450,6 +1450,26 @@ public function testCustomFilterFields()
...
+query @testQueryDirective {
+  readOneDataObjectFake @testFieldDirective {
+    id
+  }
+}
+GRAPHQL;
+        $result = $this->querySchema($schema, $query);
+        // The GraphQL server throws errors for unknown directives.
+        // A simple assertion on the result proves that directive was processed successfully.
+        $this->assertSuccess($result);
```
ChloeSartorelli, line 1470: Please add an assertion for the result as well - presumably the directive should add something to the result which we can test for.

GuySartorelli (changes requested):
Please add a PR to https://github.com/silverstripe/developer-docs as well, adding documentation for this new feature.

The documentation will be necessary for other developers to take advantage of this - but it's also a way I can make sure I'm testing it correctly. If I follow the documentation and things don't work as expected, it's clear something (either the documentation or the implementation) isn't working as expected.

I'm also still not sure what these directives *do* or what they're *for*... You've mentioned they're "part of the GraphQL specification" and that they're "a fundamental part of GraphQL" but not much more. Can you please briefly explain what this feature actually is, and why it is useful?

GuySartorelli:
@marwan38 This PR hasn't been updated for a while. Are you still keen for this feature to be included?
If so please address the comments above. If not, I will close this PR.

marwan38:
@GuySartorelli Yeah, still want it included. I've just not dedicated any time to it just yet. If you prefer to close this PR and bring up a new one when I'm ready; that's fine. 

There's a few bad commits in here for various different reasons (my own formatters, debug code, etc..).

GuySartorelli:
Sweet. Yes, please feel free to open a new PR when you have a chance to work on this - but I'll close it for now.

PR #596, opened 2024-08-13
API Migrate code and docs from other modules

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1799

Have validated that the front end schema is still usable:

`composer require silverstripe/graphql-devtools`

**app/_config/graphql.yml**
```
---
name: app-graphql
after: "*"
---
SilverStripe\Control\Director:
  rules:
    'graphql': '%$SilverStripe\GraphQL\Controller.default'

  schemas:
    default:
      src:
       - "app/_graphql"
```

**app/_graphql/models.yml**
```
Page:
  fields: '*'
  operations: '*'
```

`dev/build flush=1`

### Test using graphql devtool

`https://mysite.test/dev/graphql/ide`

**Test a query**
```
query {
  readPages {
    nodes {
      title
      content
    }
  }
}
```

**Test migrated versioning plugin**

(Make draft changes on page ID = 2 in the CMS)

```
mutation {
  publishPage(id: 2) {
    id
  }
}
```

Refersh the CMS - page has been published


tests/Middleware/CSRFMiddlewareTest.php
```diff
@@ -8,6 +8,26 @@
 
 class CSRFMiddlewareTest extends BaseMiddlewareProcessTest
 {
+    private bool $securityTokenWasEnabled = false;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        // CSRF token check is normally disabled for unit-tests
```
emteknetnz, line 16: This is needed because after moving the location of this test in recipe-kitchen-sink, some other test was somehow interfering with this.  This just sets up the Security token so that it's in the correct state for the test, and then resets it to however it was before

_config/_versioned/config.yml
```diff
@@ -0,0 +1,25 @@
+modelConfig:
```
ChloeSartorelli: These should be in `_graphql/versioned/` not `_config/_versioned/`
They're not config in the silverstripe/config sense.

_config/asset-admin.yml
```diff
@@ -0,0 +1,10 @@
+---
+Name: graphql-asset-admin-schema
+Only:
+  moduleexists: 'silverstripe/asset-admin'
+---
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\AssetAdmin\Model\ThumbnailGenerator.graphql:
+    class: SilverStripe\AssetAdmin\Model\ThumbnailGenerator
+    properties:
+      Generates: false
```
ChloeSartorelli, line 10: This was explicitly only used inside asset admin, so we don't need to carry this over. Delete this IMO

_config/versioned.yml
```diff
@@ -0,0 +1,21 @@
+---
+Name: graphql-versioned-schema
+Only:
+  moduleexists: 'silverstripe/versioned'
+---
+SilverStripe\GraphQL\Schema\Schema:
+  schemas:
+    '*':
+      src:
+        versionedSrc: 'silverstripe/graphql: _config/_versioned'
```
ChloeSartorelli:
```suggestion
        versionedSrc: 'silverstripe/graphql: _config/versioned'
```
I'm also not sure what that `silverstripe/graphql` bit does - if it still works when you remove that, you should probably remove it.

GuySartorelli (changes requested):
I assume those docs are all just directly copy/pasted, and haven't been altered at all?
They may need retouching when they get added back into the "other features" section but there's a separate card for that so that's okay.

_config/asset-admin.yml
emteknetnz: Needed for AssetAdmin unit tests, and also anyone who wants to get the thumbnail urls via the resolver would presumably benefits from this

docs/en/01_getting_started/01_server_requirements.md
```diff
@@ -0,0 +1,17 @@
...
+
+During runtime, Silverstripe CMS needs read access for the webserver user to your webroot. When the GraphQL module is installed it also needs write access for the webserver user to the following locations:
+
+- `.graphql-generated`: silverstripe/graphql uses this directory. This is where your schema is
+  stored once it [has been built](/developer_guides/graphql/getting_started/building_the_schema). Best practice
+  is to create it ahead of time, but if the directory doesn't exist and your project root is writable, the GraphQL
+  module will create it for you.
+- `public/_graphql`: silverstripe/graphql uses this directory. It's used for
+  [schema introspection](/developer_guides/graphql/tips_and_tricks#schema-introspection). You should treat this folder
+  the same way you treat the `.graphql-generated` folder.
```
ChloeSartorelli:
```suggestion
- `.graphql-generated`: This directory is where your schema is
  stored once it [has been built](/developer_guides/graphql/getting_started/building_the_schema). Best practice
  is to create it ahead of time, but if the directory doesn't exist and your project root is writable, the GraphQL
  module will create it for you.
- `public/_graphql`: This directory is used for
  [schema introspection](/developer_guides/graphql/tips_and_tricks#schema-introspection). You should treat this folder
  the same way you treat the `.graphql-generated` folder.
```
No need to say this module uses it - it's in this module's docs!

docs/en/01_getting_started/01_server_requirements.md
ChloeSartorelli, line 1: Links here are incorrect

GuySartorelli (changes requested):

docs/en/01_getting_started/01_server_requirements.md
emteknetnz: Updated

docs/en/01_getting_started/01_server_requirements.md
emteknetnz: Updated, they *should* work once we add included these docs on docs.silverstripe.org as an optional module similar to elemental and linkfield

GuySartorelli (approved):
LGTM will merge when other PRs are happy

GuySartorelli (approved):
re-approved most recent changes

PR #598, opened 2024-08-19
DEP Limit PHP support for CMS 6

GuySartorelli:
## Issue
- https://github.com/silverstripe/.github/issues/297


tests/Schema/IntegrationTest.php
ChloeSartorelli, line 1: Changes to this test needed to fix https://github.com/silverstripe/silverstripe-graphql/actions/runs/10447767783/job/28980872372?pr=598


> Calling ReflectionProperty::setValue() with a single argument is deprecated

https://www.php.net/manual/en/reflectionproperty.setvalue.php says
> Note: As of PHP 8.3.0, calling this method with a single argument is deprecated, use [ReflectionClass::setStaticPropertyValue()](https://www.php.net/manual/en/reflectionclass.setstaticpropertyvalue.php) instead.

emteknetnz (approved):

PR #605, opened 2024-09-13
API Use Symfony's Path class

GuySartorelli:
Reflects changes in https://github.com/silverstripe/silverstripe-framework/pull/11380

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11340


GuySartorelli:
No need for path traversal protection anywhere here because if an attacker can manipulate these paths they already control either your server or your codebase.

PR #610, opened 2024-09-26
MNT Fix lint failure in test

GuySartorelli:
Fixes https://github.com/silverstripe/silverstripe-graphql/actions/runs/11002947177/job/30551276517

> PHP keywords must be lowercase; expected "class" but found "Class"

## Issue
- https://github.com/silverstripe/.github/issues/313

PR #611, opened 2024-09-26
MNT Fix test that didn't account for i18n spaces

GuySartorelli:
Fixes https://github.com/silverstripe/silverstripe-graphql/actions/runs/11061549418/job/30734504257?pr=610

> 1) SilverStripe\GraphQL\Tests\Schema\IntegrationTest::testDBFieldArgs
> Failed asserting that two strings are equal.
> --- Expected
> +++ Actual
> @@ @@
> -'2/29/20, 5:00 PM'
> +'2/29/20, 5:00 PM'

The date formatter used is `IntlDateFormatter` which sometimes uses narrow non-breaking spaces. These are valid, given the intention of using `IntlDateFormatter` is to give a format that is correct for the locale used on the front-end, and in some locales the narrow non-breaking space is correct.

This test doesn't have to care what specific spaces are used - just that the correct date is there in the correct format.

## Issue
- https://github.com/silverstripe/.github/issues/313

### DIFF
--- a/tests/Schema/IntegrationTest.php
+++ b/tests/Schema/IntegrationTest.php
@@ -1434,8 +1434,8 @@ GRAPHQL;
         $node = $result['data']['readOneDataObjectFake'] ?? null;
         $this->assertEquals('This is a varchar field', $node['myField']);
         $this->assertEquals('Saturday', $node['date1']);
-        $this->assertEquals('2/29/20, 5:00 PM', $node['date2']);
-        $this->assertEquals('5:00:00 PM', $node['date3']);
+        $this->assertMatchesRegularExpression('#2/29/20,\h5:00\hPM#u', $node['date2']);
+        $this->assertMatchesRegularExpression('#5:00:00\hPM#u', $node['date3']);
         $this->assertEquals('2020', $node['date4']);
         $this->assertEquals('This is a really long text field. It has a few sentences.', $node['myText']);
     }


GuySartorelli:
lint failure is unrelated and will be fixed by https://github.com/silverstripe/silverstripe-graphql/pull/610

tests/Schema/IntegrationTest.php
```diff
@@ -1434,8 +1434,8 @@ public function testDBFieldArgs()
         $node = $result['data']['readOneDataObjectFake'] ?? null;
         $this->assertEquals('This is a varchar field', $node['myField']);
         $this->assertEquals('Saturday', $node['date1']);
-        $this->assertEquals('2/29/20, 5:00 PM', $node['date2']);
-        $this->assertEquals('5:00:00 PM', $node['date3']);
+        $this->assertMatchesRegularExpression('#2/29/20,\h5:00\hPM#u', $node['date2']);
```
emteknetnz, line 1437: This seems wrong, SapphireTest should be [setting](https://github.com/silverstripe/silverstripe-framework/blob/5/src/Dev/SapphireTest.php#L285) the i18n default locale to [en_US](https://github.com/silverstripe/silverstripe-framework/blob/5/src/i18n/i18n.php#L80).

emteknetnz (changes requested):

tests/Schema/IntegrationTest.php
emteknetnz: Actually it might be html5 overriding it - https://github.com/silverstripe/silverstripe-framework/blob/5/src/Forms/DateField.php#L450

However there's a [comment](https://github.com/silverstripe/silverstripe-framework/blob/5/src/Forms/DateField.php#L108) that this should force a particular standard of date formats, so even then we still shouldn't get thin spaces?

ChloeSartorelli: This one is draft for a reason.

ChloeSartorelli: Based on https://github.com/php/php-src/issues/11600 it seems that there was an intentional change in "ICU" which seems to be what's used under-the-hood for the i18n date formatting. This change includes the narrow non-breaking spaces which seem like they're here to stay.

We'll need to update a bunch of tests to account for it, including this one.

emteknetnz (approved):

emteknetnz (changes requested):
Linting failure

GuySartorelli:
https://github.com/silverstripe/silverstripe-graphql/pull/611#issuecomment-2378132987
> lint failure is unrelated and will be fixed by https://github.com/silverstripe/silverstripe-graphql/pull/610

emteknetnz (approved):

Issue #630, opened 2025-04-09
License for GraphQL module doesn't match other module


maxime-rainville:
### Module version(s) affected

5.3.0

### Description

The [license for the GraphQL module](https://github.com/silverstripe/silverstripe-graphql/blame/5/LICENSE) is assigned to @chillu rather than Silverstripe.

### How to reproduce

- Read the license.

### Possible Solution

Copy the license form framework. https://github.com/silverstripe/silverstripe-framework/blob/5/LICENSE

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

GuySartorelli:
@chillu feel free to raise a PR to change this if you would like to. I think technically we can't change who owns the license without that person saying "yes"

PR #631, opened 2025-04-10
DOC Update the LICENSE file to be in line with other modules

maxime-rainville:
## Description
Copy license from [silverstripe/cms](https://github.com/silverstripe/silverstripe-cms/blob/5/LICENSE).

## Issues
- #630

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,16 +1,27 @@
-Copyright (c) 2016, Ingo Schommer
+Copyright (c) 2016, Silverstripe Ltd.
 All rights reserved.
 
-Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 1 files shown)


michalkleiner:
Should @chillu approve this? :-D

GuySartorelli:
I think he'd have to from a legal standpoint. He owns the license right now, according to the license file. We can't really just rip that away without his say so :sweat_smile:

maxime-rainville:
Technically, you could argue that he was a Silverstripe employee at the time and the module was created on Silverstripe time, therefore Silverstripe owns the thing. But yes, it's probably best to wait for his OK.

(If you try to claim ownership of any of the OSS module on my private GitHub, I will fight you to death!)

Issue #632, opened 2025-04-19
Auto-generate schema description for DataObject types


maxime-rainville:
### Description

GraphQL schema are meant to be "self-documenting". You can provide description for developers using your schema so they can have a clear description of they are interacting with.

We can use the [`class_description`](https://api.silverstripe.org/5/SilverStripe/ORM/DataObject.html#property_class_description) to auto populate that description.

![Screenshot from 2025-04-19 21-30-17](https://github.com/user-attachments/assets/03a0f4fa-976d-48eb-addf-31fccde3f9e7)

### Validations

- [x] You intend to implement the feature yourself
- [x] You have read the [contributing guide](https://docs.silverstripe.org/en/contributing/code/)
- [x] You strongly believe this feature should be in core, rather than being its own community module
- [x] You have checked for existing issues or pull requests related to this feature (and didn't find any)

### Pull requests
- https://github.com/silverstripe/silverstripe-graphql/pull/633

PR #633, opened 2025-04-19
NEW Generate description from class_description when scaffolding types from DataObject

maxime-rainville:
## Description
Update the schema DatObject scaffolding logic to read the description from DataObjects and use that to populate the GraphQL schema description for the resulting types.

## Manual testing steps

- Add `class_description` to a Test DataObject.
- [Set up your default graphql schema](https://gist.github.com/maxime-rainville/de1c1ef5d716551f0eb93b135c4c7105) for your project and your test DataObject to your schema.
- Generate your schema.
- Use the GraphQL dev tools to access your schema via `/dev/graphql/ide`
- Use the Schema doc view to see the description for your generated type.

Screenshot from 2025-04-19 21-30-17

## Issues
- #632

## Pull request checklist
- [X] The target branch is correct
- [x] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [x] CI is green


src/Schema/Interfaces/ModelTypeDescription.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+
+namespace SilverStripe\GraphQL\Schema\Interfaces;
+
+/**
+ * Implementors of this interface can provide a description for a model type
+ */
+interface ModelTypeDescription
```
maxime-rainville, line 9: This probably should be in `SchemaModelInterface` but that would be a breaking change.

src/Schema/Type/ModelType.php
```diff
@@ -54,6 +55,10 @@ public function __construct(SchemaModelInterface $model, array $config = [])
             array_map('strtolower', $model->getBlacklistedFields()) :
             [];
 
+        if ($model instanceof ModelTypeDescription) {
```
maxime-rainville, line 58: This mimics the approach above that is used with ModelBlacklist.

tests/Schema/DataObject/DataObjectModelTest.php
maxime-rainville, line 1: Honestly, I didn't find any obviously great place to park those test. The existing tests don't seem to be targeted at specific classes.

maxime-rainville (commented):
I considered allowing developers to override the description in their schema definition, but opted against it because it felt like subverting the `class_description` API. It would have look something like that.

```
SilverStripe\FrameworkTest\Model\Company:
  description: "My overridden description"
  fields: '*'
  operations:
    '*': true
```

I also considered provided description for the scaffolded queries and mutations, but opted against it because the description would have been so painfully obvious, it didn't seem worth it. e.g. `readOneCompany: Retrieve a Company`

src/Schema/DataObject/DataObjectModel.php
```diff
@@ -382,6 +384,11 @@ public function getTypeName(): string
         return $this->getModelConfiguration()->getTypeName($class);
     }
 
+    public function getTypeDescription(): ?string
+    {
+        return $this->dataObject->classDescription() ?? null;
```
ChloeSartorelli, line 389:
```suggestion
        return $this->dataObject->i18nClassDescription() ?? null;
```
Unless there's a good reason to not use i18n here?

src/Schema/Type/ModelType.php
```diff
@@ -54,6 +55,10 @@ public function __construct(SchemaModelInterface $model, array $config = [])
...
+            $this->setDescription($model->getTypeDescription());
```
ChloeSartorelli, line 59: Seems fine - if  `$config['description']` has been set this will be overridden in `applyConfig` which gives developers a way to override this if they super duper want to.

GuySartorelli (changes requested):

src/Schema/DataObject/DataObjectModel.php
maxime-rainville: Not overly fuss either way.

I would argue that the description in the GraphQL schema is more like developer doc than CMS user doc. We don't normally localise the dev doc.

We also don't provide a way to localise the description when manually defining types or queries.

```yml
types:
  Monkey:
    description: "This description will only be available in English and can't be translated."
    fields:
      name:
        type: String
      age:
        type: Int
      species:
        type: String
```

It's worth pointing out that the description is set when you build your schema. The description will then be stored in the `.graphql-generated` folder. So if Friedrich has his CMS locale set to German and decides to rebuild the schema, everyone else will get their GraphQL schema description in German whether they like or not.

PR #640, opened 2025-07-30
ENH Optimise SQL queries to improve performance

GuySartorelli:
While not reducing the _number_ of queries, I've updated a bunch of calls to `find()` and `byID()` (which were only being used to check if a record exists) to instead use `filter()->exists()` which is more efficient.

## Issue

- https://github.com/silverstripe/.github/issues/416

### DIFF
--- a/src/Modules/AssetAdmin/FileFilter.php
+++ b/src/Modules/AssetAdmin/FileFilter.php
@@ -97,8 +97,8 @@ class FileFilter
 
         // Filter unknown id by known child if search is not applied
         if (!$search && isset($filter['anyChildId'])) {
-            $child = File::get()->byID($filter['anyChildId']);
-            $id = $child ? ($child->ParentID ?: 0) : 0;
+            $parentIDs = File::get()->filter('ID', $filter['anyChildId'])->limit(1)->column('ParentID');
+            $id = empty($parentIDs) ? 0 : $parentIDs[0];
             if ($id) {
                 $list = $list->filter('ID', $id);
             } else {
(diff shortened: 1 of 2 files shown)


src/Modules/AssetAdmin/FileFilter.php
```diff
@@ -97,8 +97,8 @@ public static function filterList(SS_List $list, $filter)
 
         // Filter unknown id by known child if search is not applied
         if (!$search && isset($filter['anyChildId'])) {
-            $child = File::get()->byID($filter['anyChildId']);
-            $id = $child ? ($child->ParentID ?: 0) : 0;
+            $parentIDs = File::get()->filter('ID', $filter['anyChildId'])->limit(1)->column('ParentID');
+            $id = empty($parentIDs) ? 0 : $parentIDs;
```
GuySartorelli, line 101: Since the `ParentID` is the only thing we actually need here, it's more efficient to just grab that directly from the database instead of hydrating the whole `File` record.

src/Modules/AssetAdmin/FileFilter.php
```diff
@@ -97,8 +97,8 @@ public static function filterList(SS_List $list, $filter)
 
         // Filter unknown id by known child if search is not applied
         if (!$search && isset($filter['anyChildId'])) {
-            $child = File::get()->byID($filter['anyChildId']);
-            $id = $child ? ($child->ParentID ?: 0) : 0;
+            $parentIDs = File::get()->filter('ID', $filter['anyChildId'])->limit(1)->column('ParentID');
+            $id = empty($parentIDs) ? 0 : $parentIDs[0];
```
ChloeSartorelli, line 101: Since the `ParentID` is the only thing we actually need here, it's more efficient to just grab that directly from the database instead of hydrating the whole `File` record.

We unfortunately don't have a `columnForSingleRecord()` method so we have to get `column()` which returns an array (with only one item in it because of the `limit(1)` even if we somehow had multiple records with the same ID which ought to be impossible) and then get the 0th item.

emteknetnz (approved):

PR #647, opened 2026-03-16
FIX AbstractTypeRegistry should be able to handle scalar types (fixes #646)

### LINKED ISSUE #646: "Missing graphql file for Int" with latest webonyx/graphql-php version
### Module version(s) affected

5.3, 6.x

### Description

[`v15.31.0`](https://github.com/webonyx/graphql-php/releases/tag/v15.31.0) of `webonyx/graphql-php` has caused errors in our module, specifically [these changed lines](https://github.com/webonyx/graphql-php/commit/ef79ce18c14dd0d88bfc3a6c02a00f2adcbdf1bd#diff-8aeb029ad22e58e6a5dac793f8367eb5a28a1ba50a135de2904c6978762dfb0aR191-R205) (GitHub doesn’t auto-scroll for me, lines 191-205 of src/Type/Schema.php). The code now calls our type loader to check if we’ve overridden scalar types, it seems our type loader can’t currently handle this.

I don’t think it’s an issue in `webonyx/graphql-php` which is why I’ve raised it here: it seems to me that any type loader implementation should support loading scalar types.

### How to reproduce

`composer update` a Silverstripe 5 site to install `v15.31.0` of `webonyx/graphql-php`, then run a dev/build. You’ll see the error coming from the admin schema build process.

### Possible Solution

PR to come, workaround is to pin to `"webonyx/graphql-php": "15.30.2"`

### Additional Context

_No response_

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

lozcalver:
## Description
Updates `AbstractTypeRegistry` to support scalar types following [a change in `webonyx/graphql-php`](https://github.com/webonyx/graphql-php/commit/ef79ce18c14dd0d88bfc3a6c02a00f2adcbdf1bd)

## Manual testing steps
Run a dev/build with `webonyx/graphql-php: v15.31.0` or higher

## Issues

- #646

## Pull request checklist

- [x] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [x] CI is green

### DIFF
--- a/src/Schema/Storage/AbstractTypeRegistry.php
+++ b/src/Schema/Storage/AbstractTypeRegistry.php
@@ -8,6 +8,7 @@ use GraphQL\Type\Definition\Type;
 use GraphQL\Type\Definition\NonNull;
 use GraphQL\Type\Definition\ListOfType;
 use Exception;
+use ReflectionMethod;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\GraphQL\Schema\SchemaBuilder;
 use SilverStripe\GraphQL\Schema\Exception\EmptySchemaException;
@@ -47,6 +48,13 @@ abstract class AbstractTypeRegistry
         try {
             return static::fromCache($typename);
         } catch (Exception $e) {
+            // Built-in scalars (with no schema-level override) will trigger exceptions as they don't have a
+            // generated file. So we have to catch that here and return the built-in types.
+            $builtIn = static::getBuiltinScalar($typename);
+            if ($builtIn && !static::hasRegistryScalarOverride($typename)) {
+                return $builtIn;
+            }
+
             if (!preg_match('/(Missing|Unknown) graphql/', $e->getMessage()) || !static::canRebuildOnMissing()) {
                 throw $e;
             }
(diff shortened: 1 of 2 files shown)


src/Schema/Storage/AbstractTypeRegistry.php
```diff
@@ -139,6 +147,24 @@ protected static function fromCache(string $typename)
...
+    private static function hasRegistryScalarOverride(string $typename): bool
+    {
+        if (!method_exists(static::class, $typename)) {
+            return false;
+        }
+
+        // Override exists only when concrete registry implementation declares Int()/String() etc itself.
+        // Inherited methods from this abstract class don't count
+        $method = new ReflectionMethod(static::class, $typename);
+        return $method->getDeclaringClass()->getName() === static::class;
```
lozcalver, line 165: This method _seems_ unnecessary, because it looks like we could just do this instead as there are already methods in this class for each of the scalar types:

```php
if (method_exists(self::class, $typename)) {
    return call_user_func([self::class, $typename]);
}
```

That could lead to infinite recursion though. The generated code for the type loader (`.graphql-generated/__graphql-schema.php`) would look like this if the `Int` type was overridden:

```php
public static function Int() { return static::get('Int'); }
```

In normal operation, that would be a cache-hit inside the `get()` method and would return the generated `Int` file. But in the event of an exception, it would result in an infinite loop because `get('Int')` would call `static::Int()`, which would call `get('Int')` etc.

lozcalver:
Hang fire on this pending the discussion in https://github.com/webonyx/graphql-php/issues/1874

emteknetnz (commented):
Seems good, thanks for doing this

Note that the upstream webonyx fix https://github.com/webonyx/graphql-php/issues/1874 - which was released in  [v15.31.1](https://github.com/webonyx/graphql-php/releases/tag/v15.31.1) did not appear to fix issue, so we should merge this.

lozcalver:
I think we could revert this now, [v15.31.2](https://github.com/webonyx/graphql-php/releases/tag/v15.31.2) has rolled back the change that caused problems with type loaders. So the code in this PR is essentially redundant (though won’t cause any harm)

